PR #1, opened 2019-04-24
Feature/summary care record

Philip-Woods:


.gitignore
```diff
@@ -102,3 +102,7 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+
+# PyCharm settings
+.idea
```
alibombaye: I've already add this to master 👍

SCR/TestRunner.py
```diff
@@ -0,0 +1,22 @@
+import unittest
+from spike_partials.test_RecordTarget import RecordTargetTest
+from spike_partials.test_ReplacementOf import ReplacementOfTest
+
+if __name__ == '__main__':
+    #suite = unittest.TestLoader().loadTestsFromTestCase(RecordTargetTest)
+    #unittest.TextTestRunner(verbosity=2).run(suite)
```
alibombaye: Are these comments still required?

.gitignore
Philip-Woods: Just merged master in 👍

SCR/scr/tests/test_gpsummaryupdate.py
```diff
@@ -0,0 +1,110 @@
...
+        render = self.summaryCareRecord.populate_template_with_file(file_path)
+
+        Utilities.assert_xml_equal(expected, render)
+
+    def test_empty_hash(self):
+        """
+        Tests the contents are empty when a completely blank hash is provided
+        """
+        root = etree.parse(str(Path(ROOT_DIR) / 'scr/tests/test_xmls/EmptyHash.xml'))
+        expected = etree.tostring(root)
```
alibombaye: You convert tostring here and then pass this to your utility function `assert_xml_equal` which I think objectifies it and converts it tostring again. Is this step required if the Utility function is doing this step also?

SCR/TestRunner.py
Philip-Woods: I've removed this file as it was mainly concerned with the spike tests I build when investigating mustache

SCR/scr/tests/test_gpsummaryupdate.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class FullTest(unittest.TestCase):
+
+    summaryCareRecord = SummaryCareRecord()
+
+    def test_basic(self):
+        """
+        A basic test using the clean summary update from the spine tests
+        :return:
```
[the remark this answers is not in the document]
Philip-Woods: Removed these return instances.

SCR/TestRunner.py
alibombaye: 👍

SCR/scr/tests/test_gpsummaryupdate.py
Philip-Woods: This is a good point, there isn't really any reason for objectifying it at the beginning of the test other than just to read the data from the file, I could change it to read the string from the file as normal I/O if you think that's better?

alibombaye: Yeah if you don't really need the objectifying of the file then might cut out a few unnecessary steps. Might not need the Utility function if both files are read in as strings that can be compared 🤷‍♂️

Philip-Woods: Yeah maybe, though I think it's the same number of lines to read the file normally. I think I'll still need the utility method as a simple string comparison was having issues with formatting and new lines  when I first tried it

alibombaye: ah yeah that would makes sense.

PR #2, opened 2019-04-25
Red 7 edifact payload

alibombaye:
This PR contains a domain model of the edifact message for a basic patient birth registration also known as Type 1.

The domain model is populated with the registration details required. These details are then substituted into the equivalent segment.

### DIFF
--- a/nhais-adaptor/edifact/models/message.py
+++ b/nhais-adaptor/edifact/models/message.py
@@ -103,10 +103,12 @@ class MessageSegmentTrigger2(SegmentCollection):
         formatted_date = datetime.strptime(date_of_birth, '%Y-%m-%d').strftime('%Y%m%d')
         segments = [
             Segment(key="S02", value="2"),
-            Segment(key="PNA", value=f"PAT+{id_number}:OPI+++SU:{name.family_name}+FO:{name.first_given_forename}+TI:{name.title}+MI:{name.middle_name}+FS:{name.third_given_forename}"),
+            Segment(key="PNA",
+                    value=f"PAT+{id_number}:OPI+++SU:{name.family_name}+FO:{name.first_given_forename}+TI:{name.title}+MI:{name.middle_name}+FS:{name.third_given_forename}"),
             Segment(key="DTM", value=f"329:{formatted_date}:102"),
             Segment(key="PDI", value=f"{gender}"),
-            Segment(key="NAD", value=f"PAT++{address.house_name}:{address.address_line_1}:{address.address_line_2}:{address.town}:{address.county}+++++{address.post_code}")
+            Segment(key="NAD",
+                    value=f"PAT++{address.house_name}:{address.address_line_1}:{address.address_line_2}:{address.town}:{address.county}+++++{address.post_code}")
         ]
         super().__init__(segments=segments)
 
@@ -129,4 +131,3 @@ class Message(SegmentCollection):
         msg_trailer = MessageTrailer(number_of_segments=number_of_segments, sequence_number=sequence_number)
         segments = [msg_header, message_beginning, message_segment_trigger_1, message_segment_trigger_2, msg_trailer]
         super().__init__(segments=segments)
-
--- a/nhais-adaptor/edifact/models/message.py
+++ b/nhais-adaptor/edifact/models/message.py
@@ -93,12 +93,13 @@ class Message(SegmentCollection):
     a collection of Segments
     """
 
-    def __init__(self, header, message_beginning, trailer):
+    def __init__(self, header, message_beginning, message_segment_trigger_1, trailer):
         """
         :param header: the header of the message
         :param message_beginning: the beginning of the message
+        :param message_segment_trigger_1: Segment trigger 1 registration information
         :param trailer: the trailer of the message
         """
-        segments = [header, message_beginning, trailer]
+        segments = [header, message_beginning, message_segment_trigger_1, trailer]
         super().__init__(segments=segments)
 
--- a/nhais-adaptor/edifact/models/message.py
+++ b/nhais-adaptor/edifact/models/message.py
@@ -1,7 +1,6 @@
 from datetime import datetime
 from edifact.models.segment import Segment, SegmentCollection
-from edifact.models.name import Name
-from edifact.models.address import Address
+
 
 class MessageHeader(Segment):
     """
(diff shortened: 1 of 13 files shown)


nhais-adaptor/edifact/models/message.py
```diff
@@ -0,0 +1,161 @@
...
+        :param message_segment_trigger_2: Segment trigger 2 personal information about patient
+        """
+        msg_header = MessageHeader(sequence_number=sequence_number)
+        number_of_segments = message_beginning.size() + message_segment_trigger_1.size() + message_segment_trigger_2.size() + 2
+        msg_trailer = MessageTrailer(number_of_segments=number_of_segments, sequence_number=sequence_number)
+        segments = [msg_header, message_beginning, message_segment_trigger_1, message_segment_trigger_2, msg_trailer]
+        super().__init__(segments=segments)
+
+
+class Messages(object):
```
Philip-Woods: Would it hold any value to have these collection classes inherit from the list class? This would allow you to append messages instead of passing a list in and wouldn't need a custom size method.

nhais-adaptor/edifact/models/segment.py
```diff
@@ -0,0 +1,51 @@
...
+    def to_edifact(self):
+        """
+        generates the edifact message of the segment
+        :return: a string of the formatted edifact message using the key and value
+        """
+        edifact_segment = f"{self.key}+{self.value}{Segment.TERMINATOR}"
+        return edifact_segment
+
+
+class SegmentCollection(object):
```
Philip-Woods: Again is making this a List useful?

alibombaye: 👍 Nice

nhais-adaptor/edifact/models/message.py
alibombaye: 👍

nhais-adaptor/edifact/models/tests/interchange_test.py
```diff
@@ -0,0 +1,83 @@
...
+                            third_given_forename="John")
+        patient_address = Address(house_name="MOORSIDE FARM", address_line_1="OLD LANE",
+                                  address_line_2="ST PAULS CRAY", town="ORPINGTON", county="KENT", post_code="BR6 7EW")
+
+        msg_trg_2 = MessageSegmentTrigger2(id_number="N/10/10", name=patient_name, date_of_birth="2019-04-20",
+                                           gender="1", address=patient_address)
+
+        msg = Message(sequence_number="00001", message_beginning=msg_bgn, message_segment_trigger_1=msg_trg_1,
+                      message_segment_trigger_2=msg_trg_2)
+        msgs = Messages([msg])
```
Philip-Woods: Could these longer initialisations be formatted consistently? And I couldn't find variable naming conventions for python but I know in C# I think you attempt to avoid shortened variable names, does the same apply in Python?

alibombaye: I am not sure on best practice here for python. 
I suppose I want to make it clear what the named arguments are and shorter naming convention might no achieve that but this does make creating an instance of a class ridiculously long.
Do you think something like this is still clear?
`Message(seq_num="00001", msg_bgn=msg_bgn, msg_sgt_tgr_1=msg_trg_1, msg_sgt_tgr_2=msg_trg_2)`

nhais-adaptor/edifact/models/message.py
```diff
@@ -0,0 +1,155 @@
+from datetime import datetime
```
Philip-Woods: Are these message/interchange/segment files actually models? They're performing additional logic compared to the Addess/name classes. Might be worth separating these files into Edifact generation/information models?

nhais-adaptor/edifact/models/message.py
```diff
@@ -0,0 +1,155 @@
...
+            Segment(key="NAD", value=f"GP+{party_id}:900"),
+            Segment(key="HEA", value=f"ACD+{acceptance_code}:ZZZ"),
+            Segment(key="HEA", value=f"ATP+{acceptance_type}:ZZZ"),
+            Segment(key="DTM", value=f"956:{formatted_date_time}:102"),
+            Segment(key="LOC", value=f"950+{location}"),
+        ]
+        super().__init__(segments=segments)
+
+
+class MessageSegmentTrigger2(SegmentCollection):
```
Philip-Woods: Is there any better name we can use for these two classes?

nhais-adaptor/edifact/models/message.py
```diff
@@ -0,0 +1,155 @@
...
+            Segment(key="S02", value="2"),
+            Segment(key="PNA",
+                    value=f"PAT+{id_number}:OPI+++SU:{name.family_name}+FO:{name.first_given_forename}+TI:{name.title}+MI:{name.middle_name}+FS:{name.third_given_forename}"),
+            Segment(key="DTM", value=f"329:{formatted_date}:102"),
+            Segment(key="PDI", value=f"{gender}"),
+            Segment(key="NAD",
+                    value=f"PAT++{address.house_name}:{address.address_line_1}:{address.address_line_2}:{address.town}:{address.county}+++++{address.post_code}")
+        ]
+        super().__init__(segments=segments)
+
```
Philip-Woods: Seems like some of these keys are being used multiple times ("NAD"?), could they be extracted into human readable variable names? Just to avoid having a lot of repeated magic strings

nhais-adaptor/edifact/models/tests/interchange_test.py
Philip-Woods: Just to document this I was referring to the actual variable names (`msg`, `msgs`, `msg_trg_2`) but we decided it's acceptable convention to shorten variable names in Python.

nhais-adaptor/edifact/models/message.py
alibombaye: I was thinking about this. Segment Trigger 2 refers to the patient details Segment. But this is always a Segment Trigger 2 to edifact. I suppose that can detail can be abstracted away here. Will update to more meaningful names 👍

nhais-adaptor/edifact/models/message.py
alibombaye: I see what you mean. They are being used here to represent to model the edifact message that is generated. I think once I do the first part of this POC which is to get the details for a FHIR payload. This may or may not make sense. Is it ok to revisit this later?

nhais-adaptor/edifact/models/message.py
alibombaye: I think so but I'm little inclined to doing this at the moment. When I come to build the parser of the edifact message this will probably change so the parser will need to know what it is working with. So constants and keys will be generated then

nhais-adaptor/edifact/models/message.py
Philip-Woods: Yeah that's fine by me

nhais-adaptor/edifact/models/message.py
Philip-Woods: That's fine, as long as it's something we're aware of.

PR #5, opened 2019-04-30
MHS Reference Implementation - Add Request Wrapper

gareth-allan:
This PR adds the request wrapper component required by the MHS reference implementation. This consists of a mustache template for the EBXML message (`data/templates/ebxml.mustache`) and a specialisation of the common `PystacheMessageBuilder` class, `EbXmlMessageBuilder` that uses this template by default.

I've also updated the common directory to make it a valid package suitable for `pipenv install`ing from the other projects in this repository. This required the addition of a setup.py file. I've added a Pipfile to the common directory as well, to allow us to work on it as a stand-alone project.

Finally, I've made the `SCR` project use Pipenv and import the common module in the same way as the MHS project. This now allows us to use `pipenv shell` from the `SCR` directory, or run the `pipenv run tests` command to run all the tests.


common/builder/tests/testpystachemessagebuilder.py
```diff
@@ -0,0 +1,22 @@
+import os
```
alibombaye: Just one on the file name. I'm not sure on best practice but I've been using underscore to separate words.

gareth-allan: As discussed offline, I think I was misinterpreting [PEP-8's section on package and module names](https://www.python.org/dev/peps/pep-0008/#package-and-module-names). It says pretty clearly that:
> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability.

I think underscores would improve readability here, so I'll make this change (and do the same in any similarly-named files).

common/utilities/tests/test_file_utilities.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_get_file_string(self):
+        current_dir = os.path.dirname(__file__)
+        test_files_dir = os.path.join(current_dir, TEST_FILE_DIR)
+        test_file = os.path.join(test_files_dir, TEST_FILE)
+
+        loaded_string = FileUtilities.get_file_string(test_file)
+
+        self.assertEqual(TEST_STRING, loaded_string, "The string loaded should match the one expected.")
+
+    def test_normalize_line_endings(self):
```
alibombaye: One thing I found quite useful is the `subTest` to group tests together but different variations. Perhaps this could be of use here
`with self.subTest("description of test"):`

gareth-allan: As discussed offline, this does look like it could be useful to produce a parameterised version of this test. I'll look at doing this in a subsequent change.

PR #8, opened 2019-05-02
Feature/red 16 spine client

gareth-allan:
This PR adds the capability to send messages to the MHS reference implementation.

Details:
- Added an `interactions.json` config file and the `InteractionsFile` class to handle loading this message-specific config from the file
- Added an `HttpTransport` class that handles sending a message over HTTP, using details taken from the interactions configuration
- Added a `Sender` class that uses the `EbxmlMessageBuilder`, `InteractionsFile` & `HttpTransport` classes to send a message (wrapping it with an ebXML wrapper if required)
- Added a `main.py` script that uses the above components to send an example GP summary upload message to Opentest and print the response


mhs-reference-implementation/mhs/config/interactions.py
```diff
@@ -0,0 +1,24 @@
+import json
+
+
+class InteractionsFile:
```
alibombaye: I am wondering if naming this class as `InteractionsConfig` would be better here, since this is specifically used to load the config file and get the resulting config details based on the interaction name?

mhs-reference-implementation/mhs/sender/sender.py
```diff
@@ -0,0 +1,43 @@
...
+        :param message_builder: The message builder that can be used to build a message.
+        :param transport: The transport that can be used to send messages.
+        """
+
+        self.interactions_config = interactions_config
+        self.message_builder = message_builder
+        self.transport = transport
+
+    def send_message(self, interaction_name, message_to_send):
+        """Send the specified messaage for the interaction named.
```
alibombaye: small typo here `messaage`

mhs-reference-implementation/mhs/sender/sender.py
```diff
@@ -0,0 +1,43 @@
...
+        :return: The content of the response received when the message was sent.
+        """
+
+        interaction_details = self.interactions_config.get_interaction_details(interaction_name)
+
+        requires_ebxml_wrapper = interaction_details[WRAPPER_REQUIRED]
+        if requires_ebxml_wrapper:
+            context = interaction_details.copy()
+            context[FROM_PARTY_ID] = PARTY_ID
+            context[MESSAGE] = message_to_send
```
alibombaye: I'm a little unclear what's going on here. Does this add some additional properties to the interaction details received from the config? if an ebxml wrapper flag is true?

mhs-reference-implementation/mhs/sender/tests/test_sender.py
```diff
@@ -0,0 +1,38 @@
+from unittest import TestCase
+from unittest.mock import Mock, sentinel
+
+from mhs.sender.sender import Sender
+
+WRAPPER_REQUIRED = "ebxml_wrapper_required"
```
alibombaye: is this constant here required? can we not just import it from the Sender module

mhs-reference-implementation/mhs/transport/http_transport.py
```diff
@@ -0,0 +1,59 @@
...
+
+        request_method = HttpTransport._get_request_method(interaction_details)
+
+        response = request_method(url,
+                                  data=message,
+                                  headers=headers,
+                                  cert=self._cert,
+                                  verify=self._verify)
+
+        response.raise_for_status()
```
alibombaye: Just out of interest what does this do?

mhs-reference-implementation/mhs/config/interactions.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self, interactions_file_name):
+        """Create a new InteractionsFile that loads the interaction details from the specified JSON file.
+
+        :param interactions_file_name: The file to load interaction details from. This file should be in JSON format.
+        """
+        with open(interactions_file_name) as interactions_file:
+            self.interactions = json.loads(interactions_file.read())
```
Philip-Woods: Not really worth changing but here you can use `json.load(interactions_file)` to get rid of the `read()` call. 

What might be worth doing though is moving this to the utility class if we keep having to load json into a dictionary, I know I've written this code at least once before in SCR.

mhs-reference-implementation/mhs/sender/sender.py
```diff
@@ -0,0 +1,43 @@
...
+        :param interaction_name: The name of the interaction the message is related to.
+        :param message_to_send: A string representing the message to be sent.
+        :return: The content of the response received when the message was sent.
+        """
+
+        interaction_details = self.interactions_config.get_interaction_details(interaction_name)
+
+        requires_ebxml_wrapper = interaction_details[WRAPPER_REQUIRED]
+        if requires_ebxml_wrapper:
+            context = interaction_details.copy()
```
Philip-Woods: This might want to use `.deepcopy()`, otherwise problems will arise if the json ever gets changed to contain nested elements.

mhs-reference-implementation/mhs/config/interactions.py
gareth-allan: My goal was to reference the fact that this is a file-based implementation, as I think we'll ultimately replace it with something that goes and looks up interaction details in the SDS. Would something like `InteractionsConfigFile` be a bit clearer?

mhs-reference-implementation/mhs/sender/sender.py
gareth-allan: Yes, if the ebXML wrapper flag is true we need to pass a dictionary of values to the message builder in order to populate the ebXML template, so this takes the interaction details (message-specific config) and adds the party ID (specific to an MHS instance) and actual message received to the dictionary. My thinking was to allow the interaction details config to be passed through and avoid tying this class to knowing about the content of the template being populated.

Worth extracting to a nicely-named/commented method? Or do you think it's worth revising this approach?

mhs-reference-implementation/mhs/sender/tests/test_sender.py
gareth-allan: It was really just because it's used in a couple of locations, so I wanted to make it easier to update in future.

mhs-reference-implementation/mhs/config/interactions.py
gareth-allan: That's a good call (on both points). Since it'll be such a small change, I'll move it into the common module as part of this PR.

mhs-reference-implementation/mhs/transport/http_transport.py
gareth-allan: It checks the status code of the response and raises an `HTTPError` if it's 4xx or 5xx. Worth a comment?

mhs-reference-implementation/mhs/config/interactions.py
alibombaye: Yeah I think that would be a little more clearer to make it obvious that it is only concerned with the Config of the interactions and not the actual interactions

mhs-reference-implementation/mhs/sender/sender.py
alibombaye: I think the approach you have taken is fine and now I understand the syntax better it makes sense too. I think an extracted method that describes this though might read a little better.

mhs-reference-implementation/mhs/sender/tests/test_sender.py
alibombaye: but you have the same constant and value in the `sender` file too? Would it make sense to reference that constant here too.

mhs-reference-implementation/mhs/transport/http_transport.py
alibombaye: I think that is pretty clear now. I don't think it needs a comment.

PR #9, opened 2019-05-02
Feature/red 8 deserialise edifact

alibombaye:
This PR deserialises an edifact payload to domain objects containing only the relevant information needed from the edifact payload.

This edifact domain representation will then be used to populate the fhir response to the GP with the details required.


nhais-adaptor/edifact/incoming/parser/creators.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def create_message_segment_registration(message_registration_dict):
+    """
+    Creates an incoming message registration from the message registration dictionary
+    :param message_registration_dict: The dictionary will contain a list of lines relevant to the
+    message registration section S01
+    :return: MessageSegmentRegistrationDetails: The incoming representation of the edifact message registration details
+    """
+    transaction_segment = [value for key, value in message_registration_dict if key == "RFF"][0]
```
Philip-Woods: Is this not just effectively `message_registration_dict.get('RFF')[0]`? Or am I misunderstanding some logic

alibombaye: Yes unfortunately it is a list of tuples since it is possible for it to contain duplicate keys.

alibombaye: I'm going to extract this out to a little function though. Since it's doing a fair bit in the one line and it's reused in a few places in this file too

Philip-Woods: Yeah that is a little unfortunate, I just had a quick look and python has some support for type 'suggestions', which might be helpful here just to clearly specify that the dict passed in isn't a standard python dictionary

nhais-adaptor/edifact/incoming/models/interchange.py
```diff
@@ -0,0 +1,28 @@
+class InterchangeHeader(object):
```
gareth-allan: Do these classes need to specify `object` as a base class?

nhais-adaptor/edifact/incoming/parser/tests/test_creators.py
```diff
@@ -0,0 +1,56 @@
+import unittest
+from testfixtures import compare
+import edifact.incoming.parser.creators as creators
+from edifact.incoming.models.message import MessageSegmentRegistrationDetails, MessageSegmentBeginningDetails, \
+    MessageSegmentPatientDetails
+from edifact.incoming.models.interchange import InterchangeHeader
+
+
+class TestMessage(unittest.TestCase):
```
gareth-allan: I don't think this class name is right.

nhais-adaptor/edifact/incoming/parser/creators.py
```diff
@@ -0,0 +1,71 @@
...
+
+def create_message_segment_beginning(message_beginning_dict):
+    """
+    Creates an incoming message beginning from the message beginning dictionary
+    :param message_beginning_dict: The dictionary will contain a list of lines relevant to the
+    message beginning section BGM
+    :return: MessageSegmentBeginningDetails: The incoming representation of the edifact message beginning details
+    """
+    reference_segment = get_value_in_dict(dict_to_search=message_beginning_dict, key_to_find="RFF")
+    reference_values = reference_segment.split(":")
```
gareth-allan: Probably worth constants for the separators.

nhais-adaptor/edifact/incoming/parser/tests/test_creators.py
```diff
@@ -0,0 +1,56 @@
...
+    MessageSegmentPatientDetails
+from edifact.incoming.models.interchange import InterchangeHeader
+
+
+class TestMessage(unittest.TestCase):
+
+    def test_create_interchange_header(self):
+        expected = InterchangeHeader("SO01", "ROO5", "190429:1756")
+
+        interchange_header_dict = [
```
gareth-allan: Isn't this a list, rather than a dictionary?

nhais-adaptor/edifact/incoming/parser/creators.py
```diff
@@ -0,0 +1,71 @@
+from edifact.incoming.models.interchange import InterchangeHeader
```
gareth-allan: Given that there is a top-level `adaptor.outgoing` package for the adaptors that create the outgoing messages, would it make sense for this `parser` package to live at the top level too?

nhais-adaptor/edifact/incoming/parser/deserialiser.py
```diff
@@ -0,0 +1,100 @@
...
+MESSAGE_REGISTRATION_KEY = "S01"
+MESSAGE_PATIENT_KEY = "S02"
+MESSAGE_TRAILER_KEY = "UNT"
+INTERCHANGE_TRAILER_KEY = "UNZ"
+
+
+def extract_relevant_lines(original_dict, starting_pos, trigger_key):
+    """
+    From the original dict generate a smaller dict just containing the relevant lines based upon the trigger key
+    will keep looping till the terminating key is found in the terminating config
```
gareth-allan: You might need a full stop between this line and the previous one.

nhais-adaptor/edifact/incoming/parser/deserialiser.py
```diff
@@ -0,0 +1,100 @@
...
+    for (key, value) in original_dict[starting_pos:]:
+        if key not in terminating_config[trigger_key]:
+            new_dict.append((key, value))
+        else:
+            break
+
+    return new_dict
+
+
+def convert_to_dict(lines):
```
gareth-allan: As in other places, this isn't producing a dictionary, is it? It's returning a list of tuples.

nhais-adaptor/edifact/incoming/parser/tests/test_deserialiser.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def test_convert_to_dict(self):
+        expected = [
+            ("UNB", "UNOA:2+SO01+ROO5+190429:1756+00016288++FHSREG+++FHSA EDI TRANSFERS"),
+            ("UNH", "00024986+FHSREG:0:1:FH:FHS001"),
+            ("BGM", "++507")
+        ]
+
+        input_lines = [
+            "UNB+UNOA:2+SO01+ROO5+190429:1756+00016288++FHSREG+++FHSA EDI TRANSFERS",
```
gareth-allan: Maybe worth including duplicate keys in this, to verify that they are handled correctly?

nhais-adaptor/edifact/incoming/parser/deserialiser.py
```diff
@@ -0,0 +1,100 @@
...
+    :param original_dict: The original larger dictionary
+    :param starting_pos: The starting position to start the loop from.
+    This is to prevent starting the loop from the start each time and be slightly more efficient
+    :param trigger_key: The trigger key for this section that will be used to find what the
+    terminating key for the section is
+    :return: A smaller dictionary with just the relevant lines for the section
+    """
+    terminating_config = {
+        INTERCHANGE_HEADER_KEY: [MESSAGE_HEADER_KEY],
+        MESSAGE_BEGINNING_KEY: [MESSAGE_REGISTRATION_KEY],
```
gareth-allan: Isn't this quite tightly tied to the structure of the patient registration message we're parsing? i.e. The message beginning section is presumably common across messages, so it won't always be followed by a message registration key (as that is presumably specific to patient registration messages).

nhais-adaptor/edifact/incoming/parser/creators.py
alibombaye: That's interesting. 
I've added a new Type `EdifactDict = NewType("EdifactDict", List[Tuple[str, str]])`
Which I have then used to specify in the args and return type this is the Type we are expecting.
I didn't know this could be done in Python and I like that you can be explicit about the Typing

nhais-adaptor/edifact/incoming/parser/deserialiser.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if key == INTERCHANGE_HEADER_KEY:
+            interchange_header_line = extract_relevant_lines(original_dict, index, key)
+            interchange_header = creators.create_interchange_header(interchange_header_line)
+
+        elif key == MESSAGE_BEGINNING_KEY:
+            msg_bgn_lines = extract_relevant_lines(original_dict, index, key)
+            msg_bgn_details = creators.create_message_segment_beginning(msg_bgn_lines)
+
+        elif key == MESSAGE_REGISTRATION_KEY:
```
gareth-allan: I don't think this is going to scale well when we need to handle additional messages. I appreciate that we're only handling this patient registration message for now, but have you considered how this could be expanded to handler other messages?

nhais-adaptor/edifact/incoming/models/interchange.py
alibombaye: Probably not. Will remove 👍

nhais-adaptor/edifact/incoming/parser/tests/test_creators.py
alibombaye: ooops 🤦‍♂

nhais-adaptor/edifact/incoming/parser/tests/test_creators.py
alibombaye: Actually this will be the new EdifactDict type that I can now use here too

nhais-adaptor/edifact/incoming/parser/creators.py
alibombaye: 👍

nhais-adaptor/edifact/incoming/parser/creators.py
alibombaye: Do you mean to move this `parser` module to -> `adaptor/incoming`?

nhais-adaptor/edifact/incoming/parser/deserialiser.py
alibombaye: 👍 I'll keep an eye out. I imagine this isn't the only place where I've left a full stop out

nhais-adaptor/edifact/incoming/parser/deserialiser.py
alibombaye: Yes it's a special type of dict which has now a specific `EdifactDict` type. Does calling these a `dict` still apply? It's still a key value pair but with potential duplicate keys

nhais-adaptor/edifact/incoming/parser/deserialiser.py
alibombaye: Yes, that's right. So for example an `SO2` section may not exist in the response. So the terminator for `SO1` can be an `SO2` or a `UNT` (message trailer). This is why I've added the terminating keys as a List so it can be added to easily. I'm hoping working on another example may unearth more. But for now this is the structure I've come across.

nhais-adaptor/edifact/incoming/parser/deserialiser.py
alibombaye: You are probably right. I don't think this does scale well and this `if` `elif` was something I wasn't too happy about here. I'm hoping to tackle this when I take on a more complicated response next 😅

nhais-adaptor/edifact/incoming/parser/tests/test_deserialiser.py
alibombaye: There are some duplicate keys in there. Such as `NAD` is in there twice. However in different sections.

nhais-adaptor/edifact/incoming/parser/deserialiser.py
gareth-allan: I think now you've defined a specific type of edifact dictionary it's probably reasonable to stick with the dict name.

nhais-adaptor/edifact/incoming/parser/creators.py
gareth-allan: Yes, sorry, that's a much clearer way of putting it!

gareth-allan: As discussed offline, the `adaptor` module is focused on FHIR conversion, so it wouldn't make sense to move this EDIFACT parser there. I'm happy to leave it where it is.

nhais-adaptor/edifact/incoming/parser/tests/test_deserialiser.py
alibombaye: Added tests with duplicate keys 👍

PR #10, opened 2019-05-07
Feature/red 9 fhir response

alibombaye:
This PR takes in a deserialised edifact interchange and generates the FHIR response to the GP for the transaction.

It will create multiple FHIR operation definitions for multiple messages within a single interchange.


nhais-adaptor/adaptor/incoming/operation_definition_adaptor.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    format the edifact date time to a fhir format
+    :param edifact_date_time: The incoming date time stamp from the interchange header
+    :return: The formatted date time stamp
+    """
+    current_format = "%y%m%d:%H%M"
+    desired_format = "%Y-%m-%d %H:%M"
+    formatted_date = datetime.strptime(edifact_date_time, current_format).strftime(desired_format)
+    return formatted_date
+
```
Philip-Woods: Could this go with the other date formatters in the helper class?

alibombaye: 👍 yes it does. Will move

nhais-adaptor/incoming.py
```diff
@@ -0,0 +1,34 @@
...
+        'message': 'Select a NHAIS outbox to read from: ',
+        'choices': dirs
+    }
+]
+
+answers = prompt(nhais_outbox_choices)
+nhais_cypher = answers["nhais"]
+
+with open(f"./mailbox/NHAIS/{nhais_cypher}/outbox/edifact.txt", "r") as incoming_edifact_file:
+    incoming_interchange_raw = incoming_edifact_file.read()
```
Philip-Woods: There's a method in the utilities for loading a string from a file, might also be worth using the Path() module to manage your file path strings as I don't think they convert to Unix very nicely otherwise.

nhais-adaptor/incoming.py
```diff
@@ -0,0 +1,34 @@
...
+
+lines = incoming_interchange_raw.split("'\n")
```
Philip-Woods: python's string has a `.splitlines()` method, might be better to use that over a specific line ending character.

nhais-adaptor/outgoing.py
```diff
@@ -0,0 +1,48 @@
...
+        'choices': files
+    }
+]
+
+file_answer = prompt(file_choices)
+file_name = file_answer["file"]
+print(file_name)
+
+with open(f"./mailbox/GP/{gp_cypher}/outbox/{file_name}", "r") as patient_register:
+    patient_register_json = json.load(patient_register)
```
Philip-Woods: Same again here, there's a utils method that loads files into a python dict

nhais-adaptor/adaptor/incoming/operation_definition_adaptor.py
alibombaye: Discussed with @Philip-Woods and agreed I will do this as part of the next ticket RED-35.

nhais-adaptor/adaptor/incoming/tests/test_operation_definition_adaptor.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        formatted_date = incoming_adaptor.format_date_time("190501:0902")
+        self.assertEqual(formatted_date, "2019-05-01 09:02")
+
+    def test_create_operation_definition(self):
+        """
+        Tests the function to create a fhir operation definition from an incoming edifact interchange
+        """
+        with self.subTest("adapt an interchange with one message to a single fhir operation definition"):
+            sender_parameter = creators.create_parameter_with_binding(name="senderCypher", value="XX11",
```
gareth-allan: There's a lot of setting up of the interchange in these test cases. I think it's probably worth extracting this to a test helper method to make the tests read more clearly and reduce duplication.

nhais-adaptor/incoming.py
alibombaye: Yeah that's right. Do you guys think it ok to leave it as it is or worth investigating further?

nhais-adaptor/adaptor/incoming/operation_definition_adaptor.py
```diff
@@ -0,0 +1,58 @@
...
+
+def create_operation_definition(interchange: Interchange):
+    """
+    Create a fhir operation definition from the incoming interchange
+    :param interchange: The incoming interchange
+    :return: OperationDefinition:
```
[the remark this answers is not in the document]
alibombaye: Yeah the docstring needs updating. Umm I'm kinda tempted to make these statically typed. So it's obvious what's getting returned and always up to date. I'm not sure if that is the pythonic way but I think it's valuable in an adaptor when we are taking in one type and transforming it to another type. What do you guys think?

nhais-adaptor/adaptor/incoming/tests/test_operation_definition_adaptor.py
alibombaye: Yeah will do. I'm finding most of the work and time is being spent in setting up the fixture or expected data for these tests😂. I'm still trying to see if there is a more concise way of doing this.

nhais-adaptor/incoming.py
gareth-allan: As discussed offline, it looks like actual Edifact output won't contain newlines anyway, so probably OK to leave this as it is in this PR.

Philip-Woods: That's fine I missed the Edifact line ending character, it's fine to leave it I agree

nhais-adaptor/adaptor/incoming/operation_definition_adaptor.py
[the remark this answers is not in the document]
gareth-allan: Could you use [type hints](https://lwn.net/Articles/640359/) for this? It would help users of the code and IDEs understand what's going to be returned, but still be pythonic.

PR #11, opened 2019-05-07
RED-17 - Aysnc Callback Interface

gareth-allan:
This PR adds an endpoint that listens for asynchronous responses from Spine to the MHS reference implementation.

Details:
- Added a `MessageUtilities` class to the common module, with utilities to generate UUIDs and timestamps for messages
- Made `EbXmlMessageBuilder` add `message_id` (a UUID) and `timestamp` to the dictionary of values passed to it. This means each message gets a unique message ID and timestamp generated for it.
- Added `EbXmlAckMessageBuilder` & `EbXmlRequestMessageBuilder` specialisations of `EbXmlMessageBuilder` that reference the `templates/ebxml_ack.mustache` and `templates/ebxml_request.mustache` templates respectively
- Created a dummy `EbXmlMessageParser` that returns fixed values. The message parsing logic will be implemented in a subsequent task
- Added `AsyncResponseHandler` - a Tornado `RequestHandler` that uses `EbXmlMessageParser` to parse the message received and `EbXmlAckMessageBuilder` to build an acknowledgment message to return. This endpoint is intended to handle the asynchronous responses from Spine and acknowledge them. Note, I have also added `TestAsyncResponseHandler`, which uses [`tornado.testing.AsyncHTTPTestCase`](https://www.tornadoweb.org/en/stable/testing.html#tornado.testing.AsyncHTTPTestCase) to perform a simple integration test of this endpoint
- Updated `main.py` to start a Tornado application that uses the `AsyncResponseHandler` and `ClientRequestHandler` endpoints, so that requests to Spine can be triggered by hitting an endpoint. The full implementation of `ClientRequestHandler` will be handled as part of another subsequent task


mhs-reference-implementation/mhs/builder/tests/test_ebxml_ack_message_builder.py
```diff
@@ -0,0 +1,44 @@
...
+            CPA_ID: "S1001A1630",
+            CONVERSATION_ID: "79F49A34-9798-404C-AEC4-FD38DD81C138",
+            RECEIVED_MESSAGE_TIMESTAMP: "2013-04-16T07:52:09Z",
+            RECEIVED_MESSAGE_ID: "0CDBA95F-74DA-47E9-8383-7B8E9167D146",
+        })
+
+        # Pystache does not convert line endings to LF in the same way as Python does when loading the example from
+        # file, so normalize the line endings of both strings
+        normalized_expected_message = FileUtilities.normalize_line_endings(expected_message)
+        normalized_message = FileUtilities.normalize_line_endings(message)
```
alibombaye: I wonder if something like this could be used instead to compare xmls
https://python-xmlunittest.readthedocs.io/en/latest/xmlunittest.html#xml-documents-comparison-assertion
Seems like this might remove the need to normalise line endings

mhs-reference-implementation/mhs/handler/async_response_handler.py
```diff
@@ -0,0 +1,43 @@
...
+import mhs.builder.ebxml_ack_message_builder as ack_builder
+import mhs.builder.ebxml_message_builder as builder
+import mhs.parser.ebxml_message_parser as parser
+from mhs.sender.sender import PARTY_ID
+
+
+class AsyncResponseHandler(RequestHandler):
+    """A RequestHandler for asynchronous responses from a remote MHS."""
+
+    def initialize(self, ack_builder, message_parser):
```
alibombaye: just wondering why this constructor isn't `__init__`?

mhs-reference-implementation/mhs/handler/tests/test_async_response_handler.py
```diff
@@ -0,0 +1,45 @@
...
+    @patch.object(MessageUtilities, "get_uuid")
+    def test_post(self, mock_get_uuid, mock_get_timestamp):
+        mock_get_uuid.return_value = "5BB171D4-53B2-4986-90CF-428BE6D157F5"
+        mock_get_timestamp.return_value = "2012-03-15T06:51:08Z"
+        expected_response = FileUtilities.get_file_string(str(self.expected_message_dir / EXPECTED_RESPONSE_FILE))
+
+        # Once message parsing is implemented, this will need to contain values matching the expected response.
+        request_body = "test"
+
+        response = self.fetch("/", method="POST", body=request_body)
```
alibombaye: So is `get_app` creating an app server here and then this is then getting call with this fetch call?

gareth-allan: Yep, I'm basically just following the [Tornado doc on `AsyncHTTPTestCase`](https://www.tornadoweb.org/en/stable/testing.html#tornado.testing.AsyncHTTPTestCase). `get_app` returns the application to be tested and `fetch` is a method provided to make a request to a URL.

mhs-reference-implementation/mhs/handler/async_response_handler.py
gareth-allan: It's because `RequestHandler`s are instantiated by Tornado. You pass the class and a list of arguments to pass to the `initialize` method when setting up your application (see [main.py](https://github.com/nhsconnect/integration-adaptors/blob/dab73f2e0e1167b169ecd8a6617237bd777e3779/mhs-reference-implementation/main.py#L34)).

Details are in the [Tornado doc on the `Application` object](https://www.tornadoweb.org/en/stable/guide/structure.html#the-application-object).

mhs-reference-implementation/mhs/builder/tests/test_ebxml_ack_message_builder.py
gareth-allan: Ah, I hadn't done this (or rather, used @Philip-Woods' [`XmlUtilities.assert_xml_equal`](https://github.com/nhsconnect/integration-adaptors/blob/ffdecbdbbf06e0eb025d905787ad54add1549060/common/utilities/xml_utilities.py#L10) common method) in the corresponding test of the request builder, as it builds a MIME message containing an XML message so splitting that up to allow comparison, etc. seemed worse than this approach.

That doesn't apply here, so I'll update it to use the common utility method.

mhs-reference-implementation/mhs/handler/tests/test_async_response_handler.py
alibombaye: 👍 Nice

mhs-reference-implementation/mhs/handler/async_response_handler.py
alibombaye: 👍

mhs-reference-implementation/mhs/builder/tests/test_ebxml_ack_message_builder.py
alibombaye: Cool that sounds good 👍

mhs-reference-implementation/mhs/builder/tests/test_ebxml_request_message_builder.py
```diff
@@ -0,0 +1,49 @@
...
+            SYNC_REPLY: True,
+            MESSAGE: '<QUPA_IN000006UK02 xmlns="urn:hl7-org:v3"></QUPA_IN000006UK02>'
+        })
+
+        # Pystache does not convert line endings to LF in the same way as Python does when loading the example from
+        # file, so normalize the line endings of both strings
+        normalized_expected_message = FileUtilities.normalize_line_endings(expected_message)
+        normalized_message = FileUtilities.normalize_line_endings(message)
+
+        self.assertEqual(normalized_expected_message, normalized_message)
```
alibombaye: Can we use the XmlUtility here too?

gareth-allan: As discussed offline, this is [the case I mentioned above](https://github.com/nhsconnect/integration-adaptors/pull/11#discussion_r281680525) that involves a MIME message which contains XML. This seemed like the cleaner approach to testing it.

PR #14, opened 2019-05-09
Feature/one one one

Philip-Woods:
- Added a basic web request handler for post requests
- A message handler evaluates particular parts of the incoming message to provide the appropriate response
- Unit tests for the message handler


OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+import xml.etree.ElementTree as ET
+
+from builder.pystache_message_builder import PystacheMessageBuilder
+from utilities.file_utilities import FileUtilities
+from definitions import XML_PATH, TEMPLATE_PATH
+import logging
+
+basic_success_response = FileUtilities.get_file_string(XML_PATH / 'basic_success_response.xml')
+
+base_fault_response = FileUtilities.get_file_string(XML_PATH / 'basic_fault_response.xml')
```
alibombaye: should that be `basic_fault_response`?

Philip-Woods: Yeah that makes sense for them to be the same, updated.

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+
+class MessageHandler:
+    namespaces = {
+        'soap': 'http://schemas.xmlsoap.org/soap/envelope/',
+        'a': 'http://www.etis.fskab.se/v1.0/ETISws',
+        'wsa': 'http://www.w3.org/2005/08/addressing',
+        'itk': 'urn:nhs-itk:ns:201005'
+    }
+
+    def __init__(self, message_string):
```
alibombaye: any reason why this arg needs to be named `message_string` - could it just be named as `message`?

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+        logging.basicConfig(level=logging.DEBUG)
+        self.message = message_string
```
alibombaye: Don't think you need to do this if you don't need to access to it later on. Just pass the arg straight into the line below.

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+        self.message_tree = ET.fromstring(self.message)
+
+        self.check_list = [
+            self.check_action_types,
+            self.check_manifest_and_payload_count,
+            self.check_manifest_count_against_actual,
+            self.check_payload_count_against_actual,
+            self.check_payload_id_matches_manifest_id
```
alibombaye: I don't think all these in the array need to have the `self` prefix. I could be wrong. But looks like you just need access to `check_list` later on.

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+        ]
+
+    def evaluate_message(self):
+        """
+        This iterates over the message check methods searching for errors in the message, if no errors are found
+        a success response is returned
+
+        :return:
```
alibombaye: Missing what is being returned. Although you have explained it in the explanation above 🤷‍♂

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+        return 200, basic_success_response
+
+    def check_action_types(self):
+        """
+        This method checks for equality between the action type in the header, and the service value in the message
+        body as per the 'DE_INVSER' requirement specified in the requirements spreadsheet
+        :return: status code, response content
+        """
+        action = ""
+        for type_tag in self.message_tree.findall("./soap:Header"
```
alibombaye: I think perhaps naming this to `action_tag` and the one below to `service_tag` may make it clear that they are different types of tags you are iterating over.

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+            logging.warning("Action type does not match service type: (Action, Service) (%s, %s)", action,
+                            service)
+            return 500, build_error_message("Manifest action does not match service action")
+
+        return 200, basic_success_response
+
+    def check_manifest_and_payload_count(self):
+        """
+        This verifies the manifest count is equal to the payload count as per 'DE_INVMPC' requirement
+        :return:
```
alibombaye: Missing here and a few other places in this file 😈

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+
+        manifest_count = self.get_manifest_count()
+
+        payload_count = self.get_payload_count()
```
alibombaye: little confused as to where these functions are coming from. Can't see them in the `__init__`

Philip-Woods: Do they need to be in the `init`? They're just helper methods that I pulled into methods because I was using the same code a few times

alibombaye: ah I see they are functions within here. Ignore me 👍

OneOneOne/OneOneOne/tests/test_message_handler.py
```diff
@@ -0,0 +1,154 @@
+import unittest
+from pathlib import Path
+from builder.pystache_message_builder import PystacheMessageBuilder
+from utilities.file_utilities import FileUtilities
+from utilities.xml_utilities import XmlUtilities
+from OneOneOne.OneOneOne.message_handler import MessageHandler
```
alibombaye: Finding this a little odd `OneOneOne.OneOneOne` - Can the second module be named something different?

OneOneOne/OneOneOne/tests/test_message_handler.py
```diff
@@ -0,0 +1,154 @@
...
+                            'service': "urn:nhs-itk:services:201005:SendNHS111Report-Bad_Service-ThisDoesNotMatchAbove"}
+
+            msg = builder.build_message(service_dict)
+            mh = MessageHandler(msg)
+            status_code, response = mh.check_action_types()
+
+            assert (status_code == 500)
+
+            expected = FileUtilities.get_file_string(
+                str(self.expectedXmlFileDir / 'invalid_action_service_values_response.xml'))
```
alibombaye: I personally would move your expectation to the top and keep your asserts together.

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+        a success response is returned
+
+        :return:
+        """
+        for check in self.check_list:
+            status, response = check()
+            if status != 200:
+                return status, response
+
+        return 200, basic_success_response
```
alibombaye: In general I am a little confused as to why these returns are returning HttpStatus codes. Can this return something like a flag called `valid` since it looks like an evaluation is done. I don't think the Http needs to leak into here.

OneOneOne/OneOneOne/message_handler.py
alibombaye: Yeah sorry I saw it later on as I carried on scrolling.

OneOneOne/OneOneOne/message_handler.py
alibombaye: Again ignore me here. I can see these are functions within this class.

OneOneOne/OneOneOne/web_service.py
```diff
@@ -0,0 +1,50 @@
...
+    def initialize(self, servicesdict):
+        logging.basicConfig(level=logging.DEBUG)
+
+    def post(self):
+        logging.debug("Post message received ")
+        mh = MessageHandler(self.request.body)
+        status_code, message_response = mh.evaluate_message()
+
+        self.set_status(status_code)
+        logging.warning("Status code %i", status_code)
```
gareth-allan: This (and the `Sending reply`) message don't seem like they should be logged at the warning level.

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
+from lxml import etree, objectify
```
gareth-allan: Pycharm tells me this import is unused.

OneOneOne/OneOneOne/message_handler.py
gareth-allan: Why not just use a `build_error_message("Invalid message")` call instead of having this one error response defined in its own file?

OneOneOne/OneOneOne/message_handler.py
gareth-allan: Agreed, this class doesn't need to know it's being invoked via an HTTP request.

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+        'wsa': 'http://www.w3.org/2005/08/addressing',
+        'itk': 'urn:nhs-itk:ns:201005'
+    }
+
+    def __init__(self, message_string):
+        logging.basicConfig(level=logging.DEBUG)
+        self.message = message_string
+        self.message_tree = ET.fromstring(self.message)
+
+        self.check_list = [
```
gareth-allan: I like that this is built to make it easier to add new check methods, but it might be more flexible if the list was of objects that all implement a `check` method (for example). That way a new check could be added by implementing a new class, rather than adding a method to this one. If you combined that with passing the list of checks as an argument to this constructor, this class then wouldn't need to know any details of the actual checks being run, that wiring information could live in the main application class.

OneOneOne/OneOneOne/message_handler.py
```diff
@@ -0,0 +1,190 @@
...
+        :return: status code, response content
+        """
+        action = ""
+        for type_tag in self.message_tree.findall("./soap:Header"
+                                                  "/wsa:Action",
+                                                  self.namespaces):
+            action = type_tag.text
+
+        service = ""
+        for type_tag in self.message_tree.findall('./soap:Body'
```
gareth-allan: There are a lot of repeated string values for the SOAP tags. Probably worth extracting them to constants.

OneOneOne/OneOneOne/message_handler.py
gareth-allan: Do you need to use `findall` here? If you're only expecting a single instance wouldn't `find` make more sense?

OneOneOne/OneOneOne/tests/test_message_handler.py
```diff
@@ -0,0 +1,154 @@
...
+from builder.pystache_message_builder import PystacheMessageBuilder
+from utilities.file_utilities import FileUtilities
+from utilities.xml_utilities import XmlUtilities
+from OneOneOne.OneOneOne.message_handler import MessageHandler
+
+from definitions import ROOT_DIR
+
+
+class MessageHandlerTest(unittest.TestCase):
+    expectedXmlFileDir = Path(ROOT_DIR) / 'OneOneOne' / 'tests' / 'expected_output_xmls'
```
gareth-allan: Given your expected output XMLs are so similar, would it not make sense to produce the expected output from a mustache template?

OneOneOne/OneOneOne/tests/test_message_handler.py
```diff
@@ -0,0 +1,154 @@
...
+    inputXmlFileDir = Path(ROOT_DIR) / 'OneOneOne' / 'tests' / 'input_xmls'
```
gareth-allan: Your input templates all seem fairly similar too. Could you have a single template that you customise to produce the failure scenario, rather than multiples?

OneOneOne/OneOneOne/tests/test_message_handler.py
```diff
@@ -0,0 +1,154 @@
...
+
+        with self.subTest("Two differing services result in a 500 error"):
+            service_dict = {'action': "urn:nhs-itk:services:201005:SendNHS111Report-v2-0-ThisDoesNotMatchBelow",
+                            'service': "urn:nhs-itk:services:201005:SendNHS111Report-Bad_Service-ThisDoesNotMatchAbove"}
+
+            msg = builder.build_message(service_dict)
+            mh = MessageHandler(msg)
+            status_code, response = mh.check_action_types()
+
+            assert (status_code == 500)
```
gareth-allan: This should be using `self.assertEqual` (i.e the unittest methods) shouldn't it?

OneOneOne/OneOneOne/tests/test_message_handler.py
gareth-allan: Agreed.

OneOneOne/OneOneOne/tests/test_message_handler.py
```diff
@@ -0,0 +1,154 @@
...
+
+        with self.subTest("Two services which are the same should return 200 code"):
+            service_dict = {'action': "urn:nhs-itk:services:201005:SendNHS111Report",
+                            'service': "urn:nhs-itk:services:201005:SendNHS111Report"}
+
+            msg = builder.build_message(service_dict)
+            mh = MessageHandler(msg)
+            status_code, response = mh.check_action_types()
+
+            assert (status_code == 200)
```
gareth-allan: Shouldn't this verify that the expected success message is returned too?

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: Good point, I've removed the `basic_fault_response`

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: I think it was just to differentiate between a full message like the web_service and the actual 111 update message that is part of the body, I can rename it if you feel it suffices as `message`

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: 👍

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: Updates the docs to have returns here and as mentioned in a later comment

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: Yeah that's not greatly named, I've updated it to be `action_tag_value` to specify it's the actual value of the tag as well and not just the tag object.

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: 👍

OneOneOne/OneOneOne/tests/test_message_handler.py
Philip-Woods: Yep, I've renamed it to `receiver`

OneOneOne/OneOneOne/tests/test_message_handler.py
Philip-Woods: Moved it.

OneOneOne/Pipfile
```diff
@@ -0,0 +1,15 @@
+[[source]]
```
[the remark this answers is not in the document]
Philip-Woods: Yeah good point, I've primarily been running tests in pycharm so I hadnt generated one

OneOneOne/OneOneOne/web_service.py
```diff
@@ -0,0 +1,50 @@
+import logging
+from time import sleep
+
+from tornado.httpserver import HTTPServer
```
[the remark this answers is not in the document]
Philip-Woods: Binned them, was using them for trying to deal with that bug

OneOneOne/OneOneOne/web_service.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def get(self):
+        print("Test get method")
+
+    def on_finish(self) -> None:
+        self.flush()
+
+
+if __name__ == "__main__":
+    servicesDict = {'debug': True }
```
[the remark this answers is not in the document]
Philip-Woods: Good point, think it was originally used for creds, I'll give it a default in the `initialize` method in case I use it for logging later, then bin the `servicesDict` line.

OneOneOne/OneOneOne/web_service.py
```diff
@@ -0,0 +1,50 @@
...
+        mh = MessageHandler(self.request.body)
+        status_code, message_response = mh.evaluate_message()
+
+        self.set_status(status_code)
+        logging.warning("Status code %i", status_code)
+
+        logging.warning("Sending reply")
+        self.write(message_response)
+
+    def get(self):
```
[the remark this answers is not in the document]
Philip-Woods: Removed both

OneOneOne/OneOneOne/web_service.py
```diff
@@ -0,0 +1,50 @@
...
+from definitions import XML_PATH
+from OneOneOne.OneOneOne.message_handler import MessageHandler
+
+response = FileUtilities.get_file_string(XML_PATH / 'basic_success_response.xml')
+
+
+class MessageReceiver(tornado.web.RequestHandler):
+
+    def initialize(self, servicesdict):
+        logging.basicConfig(level=logging.DEBUG)
```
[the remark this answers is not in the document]
Philip-Woods: 👍

OneOneOne/OneOneOne/web_service.py
Philip-Woods: Set both as `debug`

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: binned it

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: I've updated these methods to use a pass/fail flag and kept instances of the error code in the message handler and the actual message content evaluation methods have been moved to their own classes as per Gareth's suggestion.

Philip-Woods: Also would it be worth altering the response string to be just the error message and not the full message, this would simplify the functionality to just producing an error string or is it fine as it is?

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: This has been updated as suggested, I now have an interface, `Check`, with generic methods and a check method, the message handler is passed a list of objects which implement the `Check` class and calls each check method

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: Moved a few of the more common xpaths to member variables if the `Check` class

OneOneOne/OneOneOne/tests/test_message_handler.py
Philip-Woods: Updated the values to use `self.assertEqual/False`

OneOneOne/OneOneOne/tests/test_message_handler.py
Philip-Woods: I can see why this would be better, but also feels a bit strange as I'd effectively be running the same code twice and asserting their outputs would be the same, which just seems like repeating the task rather than asserting it against what we know is correct?

OneOneOne/OneOneOne/tests/test_message_handler.py
Philip-Woods: Cut down the templates to just the one

OneOneOne/OneOneOne/tests/test_message_handler.py
Philip-Woods: Added a check for the success string as suggested

OneOneOne/reciever/message_checks/Check.py
```diff
@@ -0,0 +1,54 @@
+import logging
```
alibombaye: Comment on the name of the file. Think it should be in lowercase

OneOneOne/reciever/message_checks/Check.py
```diff
@@ -0,0 +1,54 @@
...
+        'a': 'http://www.etis.fskab.se/v1.0/ETISws',
+        'wsa': 'http://www.w3.org/2005/08/addressing',
+        'itk': 'urn:nhs-itk:ns:201005'
+    }
+
+    def __init__(self, message):
+        self.message_tree = message
+
+    def check(self):
+        pass
```
alibombaye: Cool I take it this is the base class and they'll be specialisation of this function. You might want to use `ABC` Abstract Base Class and tag this function as an `abstractmethod` will make this clear. I've just recently done something similar. Have a look at `nhais-adaptor/adaptor/outgoing/message_adaptor`

OneOneOne/reciever/message_handler.py
```diff
@@ -0,0 +1,194 @@
...
+
+class MessageHandler:
+    namespaces = {
+        'soap': 'http://schemas.xmlsoap.org/soap/envelope/',
+        'a': 'http://www.etis.fskab.se/v1.0/ETISws',
+        'wsa': 'http://www.w3.org/2005/08/addressing',
+        'itk': 'urn:nhs-itk:ns:201005'
+    }
+
+    def __init__(self, message, check_list: CheckList):
```
gareth-allan: I think it would be clearer to write this as `check_list: List[Check]`, rather than defining `CheckList`.

OneOneOne/OneOneOne/message_handler.py
gareth-allan: I think you still need to extract the 200/500 code from `MessageHandler.evaluate_message()`. So it would also return the pass/fail flag and it would be up to `MessageReceiver` to decide which HTTP status code to assign to a pass or a fail.

Also, I think it makes sense to keep the message building within the MessageHandler/Check classes. It's not something `MessageReceiver` needs to be responsible for.

OneOneOne/OneOneOne/message_handler.py
gareth-allan: I think this still holds true for the new `Check` implementations.

OneOneOne/OneOneOne/tests/test_message_handler.py
gareth-allan: A fair point. I'm happy to leave the expected output XMLs as they are then.

OneOneOne/reciever/tests/test_message_handler.py
```diff
@@ -0,0 +1,169 @@
...
+
+            fail_flag, response = mh.check()
+
+            self.assertTrue(fail_flag)
+            XmlUtilities.assert_xml_equal_utf_8(expected, response)
+
+        with self.subTest("Two services which are the same should return 200 code"):
+            service_dict = {'action': "urn:nhs-itk:services:201005:SendNHS111Report",
+                            'service': "urn:nhs-itk:services:201005:SendNHS111Report"}
+            expected = FileUtilities.get_file_string(
```
gareth-allan: `expected` is never used.

nhais-adaptor/Pipfile.lock
```diff
@@ -1,7 +1,11 @@
 {
     "_meta": {
         "hash": {
+<<<<<<< HEAD
```
gareth-allan: Looks like these are merge conflicts. They'll need fixing (presumably there shouldn't be any changes to the NHAIS adaptor Pipfiles as part of this PR?).

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: Ok I've updated the handler to have no reference to HTTP codes, it just returns an error flag now and the `MessageReceiver` now makes this decision as you suggested. I've extracted the message building from the individual check classes to the `MessageHandler` class as well, both the `MessageHandler` and `Check` classes have separate tests as well now.

OneOneOne/OneOneOne/message_handler.py
Philip-Woods: Yep sorry, missed this one!, They've been updated appropriately now

OneOneOne/reciever/message_checks/Check.py
Philip-Woods: This always gets me, fixed it

OneOneOne/reciever/message_checks/Check.py
Philip-Woods: Updated it to use ABC and I've added some docs to explain it a bit more

OneOneOne/reciever/message_checks/Check.py
```diff
@@ -0,0 +1,54 @@
+import logging
+
+from builder.pystache_message_builder import PystacheMessageBuilder
+from definitions import TEMPLATE_PATH, XML_PATH
+from utilities.file_utilities import FileUtilities
+
+
+class Check:
+
+    distribution_envelope = "./soap:Body/itk:DistributionEnvelope"
```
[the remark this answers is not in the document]
Philip-Woods: Updated

OneOneOne/reciever/message_checks/Check.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def check(self):
+        pass
+
+    def get_manifest_count(self):
+        """
+        Extracts the count on the manifest tag in the message
+        :return: manifest count as a string
+        """
+        manifests = self.message_tree.findall(self.distribution_envelope + "/itk:header/itk:manifest", self.namespaces)
```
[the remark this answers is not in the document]
Philip-Woods: Yep, updated

OneOneOne/reciever/message_handler.py
```diff
@@ -0,0 +1,194 @@
...
+            c = check(self.message_tree)  # instantiate the check object instance
+            fail, response = c.check()  # make the check call
+            if fail:
+                return 500, response
+
+        return 200, basic_success_response
+
+
+
+    def check_action_types(self):
```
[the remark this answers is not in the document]
Philip-Woods: Yep sorry, I removed them in the following commit I was working on

OneOneOne/reciever/web_service.py
```diff
@@ -0,0 +1,40 @@
+from tornado.httpserver import HTTPServer
+import tornado.ioloop
```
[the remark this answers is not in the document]
Philip-Woods: Switched it to use this shorter name

OneOneOne/reciever/message_handler.py
Philip-Woods: Good point, moved it to the parameter list

nhais-adaptor/Pipfile.lock
Philip-Woods: Good spot, think this must have appeared when I was running Fawad's tests, I'll update it to match master

OneOneOne/reciever/message_checks/check.py
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    def __init__(self, message):
+        self.message_tree = message
+
+    @abstractmethod
+    def check(self):
+        """
+        An asbtract method called by the validator to run the check
+        :return: flail flag, error message
```
alibombaye: small typo here

OneOneOne/reciever/tests/test_message_handler.py
Philip-Woods: Should be sorted now

OneOneOne/reciever/message_checks/check.py
Philip-Woods: 👍

OneOneOne/reciever/message_handler.py
```diff
@@ -0,0 +1,71 @@
...
+    def evaluate_message(self):
+        """
+        This iterates over the message check methods searching for errors in the message, if no errors are found
+        a success response is returned
+
+        :return: status code, response content
+        """
+
+        for check in self.check_list:
+            c = check(self.message_tree)  # instantiate the check object instance
```
alibombaye: I think here perhaps renaming `check` to `Check` will make it clear that an instantiation is being done. With it being lower case it looks like a method call. Perhaps then the comment isn't needed

OneOneOne/reciever/tests/test_message_checks.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def test_action_not_matching_service(self):
+
+        with self.subTest("Two differing services result in a 500 error"):
+            service_dict = {'action': "urn:nhs-itk:services:201005:SendNHS111Report-v2-0-ThisDoesNotMatchBelow",
+                            'service': "urn:nhs-itk:services:201005:SendNHS111Report-Bad_Service-ThisDoesNotMatchAbove"}
+
+            msg = self.builder.build_message(service_dict)
+            message_tree = ET.fromstring(msg)
+            mv = CheckActionTypes(message_tree)
```
alibombaye: it looks like lines 19 and 20 are quite common here with the other tests in this file. Are these lines just setting up message_tree the check needs? Perhaps a helper function here like `generate_message_tree` with the `service_dict` passed in. Then this could be inlined as `mv = CheckActionTypes(generate_message_tree(service_dict))` might make these bit easier to read. Also `mv` might better be renamed to something else

OneOneOne/OneOneOne/message_handler.py
alibombaye: 👍 This looks lot better to me now

nhais-adaptor/Pipfile
```diff
@@ -11,6 +11,7 @@ integration-adaptors-common = {editable = true,path = "./../common"}
 fhirclient = "*"
 PyInquirer = "*"
 
+
```
gareth-allan: Not a major issue, but this PR shouldn't really change the NHAIS adaptor's Pipfile at all (even if it's only whitespace).

alibombaye: @Philip-Woods stop messing about with my Pipfile dude 😂

Philip-Woods: God dammit how did I manage that 🤦‍♂

OneOneOne/reciever/message_handler.py
Philip-Woods: I've updated it to use `CheckInstance`  since `Check` is the super class which I feel I should avoid?

OneOneOne/reciever/tests/test_message_checks.py
Philip-Woods: Killed two birds with one stone and inclined the whole thing:
`fail_flag, response = CheckActionTypes(self.generate_message_tree(service_dict)).check()`

Philip-Woods: I've also updated some of the names from the `test_message_handler` as well

ghost (approved):

OneOneOne/reciever/message_handler.py
alibombaye: Cool that makes sense 👍

alibombaye (approved):

PR #16, opened 2019-05-10
RED-18 - Response Parser

gareth-allan:
This PR implements the `EbXmlMessageParser` and `EbXmlRequestMessageParser` classes to parse an ebXML header and the mime message containing it. This is used to parse the asynchronous response received to a message in order to produce a dictionary of values.

Details:
- Expanded the `TestAsyncResponseHandler` integration test to send a realistic asynchronous response body that contains values matching those in the expected acknowledgement
- Implemented `EbXmlMessageParser` - a class that parses the ebXML XML header using the [`xml.etree.ElementTree`](https://docs.python.org/3/library/xml.etree.elementtree.html) parser
- Implemented `EbXmlRequestMessageParser` - A specialisation of `EbXmlMessageParser` that splits up the messsage into MIME parts (using the [`email`](https://docs.python.org/3/library/email.html) package) before passing it on to the superclass for XML parsing. This separation is because the asynchronous response we are expecting consists of multiple MIME parts, but ebXML acknowledgements (for example) are plain XML messages. See [`ebxml_request.msg`](https://github.com/nhsconnect/integration-adaptors/compare/feature/RED-18-response-parser?expand=1#diff-9c5ff9a0aa9f545fcaa72eac7acc6077) for an example of the ebXML request format


mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
```diff
@@ -1,38 +1,137 @@
...
+        self._extract_ebxml_text_value(xml_tree, "PartyId", extracted_values, FROM_PARTY_ID, parent="From")
+        self._extract_ebxml_text_value(xml_tree, "PartyId", extracted_values, TO_PARTY_ID, parent="To")
+        self._extract_ebxml_text_value(xml_tree, "CPAId", extracted_values, CPA_ID)
+        self._extract_ebxml_text_value(xml_tree, "ConversationId", extracted_values, CONVERSATION_ID)
+        self._extract_ebxml_text_value(xml_tree, "Service", extracted_values, SERVICE)
+        self._extract_ebxml_text_value(xml_tree, "Action", extracted_values, ACTION)
+        self._extract_ebxml_text_value(xml_tree, "MessageId", extracted_values, MESSAGE_ID, parent="MessageData")
+        self._extract_ebxml_text_value(xml_tree, "Timestamp", extracted_values, TIMESTAMP, parent="MessageData")
+        self._extract_ebxml_text_value(xml_tree, "RefToMessageId", extracted_values, REF_TO_MESSAGE_ID,
+                                       parent="MessageData")
```
Philip-Woods: Would it work to have the `xml_tree` set as a member variable here, seems like we're passing that tree round a lot of methods and it could save a parameter

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
```diff
@@ -1,38 +1,137 @@
...
+        path = ".//"
+
+        if parent is not None:
+            path += EBXML_NAMESPACE + ":" + parent + "/"
+
+        path += EBXML_NAMESPACE + ":" + name
+
+        return path
+
+    def _extract_ebxml_text_value(self, xml_tree, element_name, values_dict, key, parent=None):
```
alibombaye: It looks like this function is doing 2 things. It is extracting the value from the xml tree and also update the values_dict with that value. Do you think there is any value in separating this out? or perhaps renaming this function.

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
```diff
@@ -1,38 +1,137 @@
...
+        ebxml_part, payload_part = self._extract_message_parts(msg)
+        extracted_values = super().parse_message(headers, ebxml_part)
+
+        if payload_part:
+            extracted_values[MESSAGE] = payload_part
+
+        logging.debug("Extracted values from message: %s", extracted_values)
+        return extracted_values
+
+    def _parse_mime_message(self, headers, message):
```
alibombaye: Perhaps a docstring to explain what this function will be returning.

gareth-allan: Yep, fair point. I'll add one.

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
gareth-allan: I think you're right, I could extract the dictionary update to a separate function to make things clearer. Will update.

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
```diff
@@ -1,38 +1,137 @@
...
+        content_type = headers[CONTENT_TYPE_HEADER_NAME]
+        content_type_header = CONTENT_TYPE_HEADER_NAME + ": " + content_type + "\r\n"
+
+        msg = email.message_from_string(content_type_header + message)
+
+        return msg
+
+    def _extract_message_parts(self, msg):
```
alibombaye: I think here it might be useful to show that a tuple is returned here of the ebxml part of the message and the payload part of the message.

mhs-reference-implementation/mhs/parser/tests/test_ebxml_message_parser.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def setUp(self):
+        self.parser = EbXmlMessageParser()
+
+    def test_parse_message(self):
+        message = FileUtilities.get_file_string(str(message_dir / "ebxml_header.xml"))
+
+        extracted_values = self.parser.parse_message(MULTIPART_MIME_HEADERS, message)
+
+        self.assertEqual(expected_values(), extracted_values)
```
alibombaye: why is the message not passed into `expected_values()`?

mhs-reference-implementation/mhs/parser/tests/test_ebxml_message_parser.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def setUp(self):
+        self.parser = EbXmlRequestMessageParser()
+
+    def test_parse_message(self):
+        message = FileUtilities.get_file_string(str(message_dir / "ebxml_request.msg"))
+
+        extracted_values = self.parser.parse_message(MULTIPART_MIME_HEADERS, message)
+
+        self.assertEqual(expected_values(message=EXPECTED_MESSAGE), extracted_values)
```
alibombaye: I feel like this test could do with a better description to explain what is going on and what to expect since I'm not really that sure. Could be my understanding of this area. But it looks like when the message is an ebxml request then the parse message will also add the message to the generated dictionary?

mhs-reference-implementation/mhs/parser/tests/test_ebxml_message_parser.py
```diff
@@ -0,0 +1,97 @@
...
+    def test_parse_message_with_no_header(self):
+        message = FileUtilities.get_file_string(str(message_dir / "ebxml_request_no_header.msg"))
+
+        with (self.assertRaises(EbXmlParsingError)):
+            self.parser.parse_message(MULTIPART_MIME_HEADERS, message)
+
+    def test_parse_message_with_no_payload(self):
+        message = FileUtilities.get_file_string(str(message_dir / "ebxml_request_no_payload.msg"))
+
+        extracted_values = self.parser.parse_message(MULTIPART_MIME_HEADERS, message)
```
alibombaye: I think in this test class perhaps the `with self.subTest("...")` might be useful to be more descriptive about what is being tested and the expectation. For example here with no payload we expect not to see the `MESSAGE` in the extracted values.

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
gareth-allan: I'll add a docstring.

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
gareth-allan: My intention was that a single instance of `EbXmlMessageParser` could be used to parse multiple messages, so setting this tree as a member variable wouldn't make sense in that respect.

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
gareth-allan: I've refactored this. Hopefully it makes things a little clearer.

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
gareth-allan: Added

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
gareth-allan: Added

mhs-reference-implementation/mhs/parser/tests/test_ebxml_message_parser.py
gareth-allan: I've used sub-tests to attempt to make this test class easier to follow.

mhs-reference-implementation/mhs/parser/tests/test_ebxml_message_parser.py
gareth-allan: I've extracted these `expected_values()` calls into named variables to hopefully make it a bit clearer why they do/don't contain a message.

alibombaye: 👍 yeah I think that makes it bit more clearer what the expectations are here

mhs-reference-implementation/mhs/parser/tests/test_ebxml_message_parser.py
alibombaye: 👍

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
```diff
@@ -1,38 +1,166 @@
...
+        self._add_if_present(extracted_values, MESSAGE_ID,
+                             self._extract_ebxml_text_value(xml_tree, "MessageId", parent="MessageData"))
+        self._add_if_present(extracted_values, TIMESTAMP,
+                             self._extract_ebxml_text_value(xml_tree, "Timestamp", parent="MessageData"))
+        self._add_if_present(extracted_values, REF_TO_MESSAGE_ID,
+                             self._extract_ebxml_text_value(xml_tree, "RefToMessageId", parent="MessageData"))
+        self._add_flag_if_present(extracted_values, DUPLICATE_ELIMINATION,
+                                  self._extract_ebxml_value(xml_tree, "DuplicateElimination"))
+        self._add_flag_if_present(extracted_values, SYNC_REPLY, self._extract_ebxml_value(xml_tree, "SyncReply"))
+        self._add_flag_if_present(extracted_values, ACK_REQUESTED, self._extract_ebxml_value(xml_tree, "AckRequested"))
```
alibombaye: I am just wondering if some of this can be extracted into some kind of parsing config. With some details about where to extract the relevant bits of data from. Then perhaps just iterating through this config to generate this extracted_value dict. Then it should be a case of updating this config if anything else is needed in the future.

gareth-allan: As discussed offline, I've created a tech-debt ticket to do this, as I think it's probably a little beyond the scope of this task.

Philip-Woods: I was wondering something similar, seems like there are quite a few similar operations happening here which makes me feel like this should be looped somehow
Also quite a few of the constants used here are the only occurrence of these constants, seeing a there are already some fixed strings here with only 1 use do we need to keep all the constants at the top?

Philip-Woods: I didn't refresh the page till I'd posted this so didn't see the previous comment but that's reasonable

gareth-allan: Regarding the constants, the idea was that they should be available for import from other modules, so that you can use them to index the dictionary returned from `parse_message`.

mhs-reference-implementation/mhs/parser/ebxml_message_parser.py
Philip-Woods: Makes sense, though if the case is this class doesn't retain state it makes me wonder if it should be static but I don't think this would add any value

PR #18, opened 2019-05-14
Nhais/red 40 move date formatter

alibombaye:
This PR moves the responsibility of date formatting for outgoing messages at the adaptor level rather than at the model level where it was previously.

The outgoing edifact model now assumes that dates are formatted correctly when they are populated.


nhais-adaptor/adaptor/incoming/common/date_formatter.py
```diff
@@ -0,0 +1,13 @@
...
+def format_date_time(edifact_date_time):
+    """
+    format the edifact date time to a fhir format
+    :param edifact_date_time: The incoming date time stamp from the interchange header
+    :return: The formatted date time stamp
+    """
+    current_format = "%y%m%d:%H%M"
+    desired_format = "%Y-%m-%d %H:%M"
+    formatted_date = datetime.strptime(edifact_date_time, current_format).strftime(desired_format)
+    return formatted_date
```
Philip-Woods: Is it worth naming this method in a way that reflects that it's turning edifact => fhir, I guess you can probably imply this from the directories but I think there's a few date formatters?

alibombaye: That's right there is a date formatter in the outgoing too module too. I wanted to avoid having incoming and outgoing everywhere and hoping the package and module structure would make it clear. I'll have a think about renaming to make it bit more obvious though 👍

PR #19, opened 2019-05-14
RED-19 - Sync-Async Wrapper

gareth-allan:
This PR implements a sync-async wrapper for the MHS reference implementation.

If the interaction requested is configured as an asynchronous one, the `ClientRequestHandler` does this by registering a callback (one of its own methods) in a shared dictionary and then 'pausing' the inbound request by waiting on an [`Event` from the tornado.locks package](https://www.tornadoweb.org/en/stable/locks.html#tornado.locks.Event). Then, when the `AsyncResponseHandler` receives the asynchronous response, it retrieves the callback from the dictionary and calls it, passing the received message payload (which is returned on the original request). This callback notifies the original thread to stop waiting and the response received is returned in response to the original request. If no asynchronous response is received after a certain timeout, the original request returns an error.

I've verified that this works as expected when running against Opentest, and the "GP Summary upload successful" message is returned for GP summary upload requests and the NHS number is returned for a (synchronous) NHS number query.

Other points of note:
- `TestClientRequestHandler` is missing a method that tests the asynchronous messaging pattern. This is because the addition of the sync-async wrapper means that it can no longer be tested with a single HTTP request. I plan to investigate the implementation of a mock spine server that will send the required asynchronous responses, but do not want to delay this PR while I do this
- I have updated `EbXmlMessageBuilder` so that it returns the message ID of the message built, as this is used as the key to tie asynchronous responses to the original requests
- I have extracted the party key of the MHS node from a constant in `Sender` to `main.py` and added details to the README of setting it to your specific value

### DIFF
--- a/mhs-reference-implementation/main.py
+++ b/mhs-reference-implementation/main.py
@@ -34,13 +34,15 @@ message_parser = EbXmlRequestMessageParser()
 # Configure logging
 logging.basicConfig(format="%(asctime)s - %(levelname)s: %(message)s", level=logging.DEBUG)
 
-# Run the Tornado server
+# Run the Tornado servers
+callbacks = {}
+
 supplier_application = Application([(r"/.*", ClientRequestHandler, dict(sender=sender))])
 supplier_server = HTTPServer(supplier_application)
 supplier_server.listen(80)
 
 mhs_application = Application([
-    (r".*", AsyncResponseHandler, dict(ack_builder=ack_builder, message_parser=message_parser))
+    (r".*", AsyncResponseHandler, dict(ack_builder=ack_builder, message_parser=message_parser, callbacks=callbacks))
 ])
 mhs_server = HTTPServer(mhs_application,
                         ssl_options=dict(certfile=certs_file, keyfile=key_file, cert_reqs=ssl.CERT_REQUIRED,
--- a/mhs-reference-implementation/main.py
+++ b/mhs-reference-implementation/main.py
@@ -17,6 +17,8 @@ from mhs.sender.sender import Sender
 from mhs.transport.http_transport import HttpTransport
 
 ASYNC_TIMEOUT = 30
+# TODO: Replace this with your party ID.
+PARTY_ID = "PARTY-ID"
 
 data_dir = Path(ROOT_DIR) / "data"
 certs_dir = data_dir / "certs"
@@ -28,7 +30,7 @@ interactions_config_file = str(data_dir / "interactions" / "interactions.json")
 interactions_config = InteractionsConfigFile(interactions_config_file)
 message_builder = EbXmlRequestMessageBuilder()
 transport = HttpTransport(str(certs_dir))
-sender = Sender(interactions_config, message_builder, transport)
+sender = Sender(interactions_config, message_builder, transport, PARTY_ID)
 
 ack_builder = EbXmlAckMessageBuilder()
 message_parser = EbXmlRequestMessageParser()
@@ -45,7 +47,8 @@ supplier_server = HTTPServer(supplier_application)
 supplier_server.listen(80)
 
 mhs_application = Application([
-    (r".*", AsyncResponseHandler, dict(ack_builder=ack_builder, message_parser=message_parser, callbacks=callbacks))])
+    (r".*", AsyncResponseHandler,
+     dict(ack_builder=ack_builder, message_parser=message_parser, callbacks=callbacks, party_id=PARTY_ID))])
 mhs_server = HTTPServer(mhs_application,
                         ssl_options=dict(certfile=certs_file, keyfile=key_file, cert_reqs=ssl.CERT_REQUIRED,
                                          ca_certs=certs_file))
--- a/mhs-reference-implementation/main.py
+++ b/mhs-reference-implementation/main.py
@@ -16,6 +16,8 @@ from mhs.parser.ebxml_message_parser import EbXmlRequestMessageParser
 from mhs.sender.sender import Sender
 from mhs.transport.http_transport import HttpTransport
 
+ASYNC_TIMEOUT = 30
+
 data_dir = Path(ROOT_DIR) / "data"
 certs_dir = data_dir / "certs"
 certs_file = str(certs_dir / "client.pem")
@@ -37,13 +39,13 @@ logging.basicConfig(format="%(asctime)s - %(levelname)s: %(message)s", level=log
 # Run the Tornado servers
 callbacks = {}
 
-supplier_application = Application([(r"/.*", ClientRequestHandler, dict(sender=sender))])
+supplier_application = Application(
+    [(r"/.*", ClientRequestHandler, dict(sender=sender, callbacks=callbacks, async_timeout=ASYNC_TIMEOUT))])
 supplier_server = HTTPServer(supplier_application)
 supplier_server.listen(80)
 
 mhs_application = Application([
-    (r".*", AsyncResponseHandler, dict(ack_builder=ack_builder, message_parser=message_parser, callbacks=callbacks))
-])
+    (r".*", AsyncResponseHandler, dict(ack_builder=ack_builder, message_parser=message_parser, callbacks=callbacks))])
 mhs_server = HTTPServer(mhs_application,
                         ssl_options=dict(certfile=certs_file, keyfile=key_file, cert_reqs=ssl.CERT_REQUIRED,
                                          ca_certs=certs_file))
--- a/mhs-reference-implementation/mhs/handler/async_response_handler.py
+++ b/mhs-reference-implementation/mhs/handler/async_response_handler.py
@@ -1,6 +1,7 @@
 import logging
+from typing import Callable, Dict
 
-from tornado.web import RequestHandler
+from tornado.web import RequestHandler, HTTPError
 
 import mhs.builder.ebxml_ack_message_builder as ack_builder
 import mhs.builder.ebxml_message_builder as builder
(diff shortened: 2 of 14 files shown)


mhs-reference-implementation/main.py
```diff
@@ -26,22 +30,25 @@
...
+
+supplier_application = Application(
+    [(r"/.*", ClientRequestHandler, dict(sender=sender, callbacks=callbacks, async_timeout=ASYNC_TIMEOUT))])
 supplier_server = HTTPServer(supplier_application)
 supplier_server.listen(80)
 
 mhs_application = Application([
-    (r".*", AsyncResponseHandler, dict(ack_builder=ack_builder, message_parser=message_parser))
-])
+    (r".*", AsyncResponseHandler,
```
alibombaye: I've noticed the `supplier_application` has path set as `r"/.*"` but this is just `r".*` - just wondering why the difference here?

mhs-reference-implementation/mhs/handler/async_response_handler.py
```diff
@@ -1,46 +1,59 @@
...
         :param ack_builder: The message builder to use when building ebXML acknowledgement messages.
         :param message_parser: The message parser to use to parse ebXML asynchronous responses.
-        :return:
+        :param callbacks: The dictionary of callbacks to use when a message is received.
+        :param party_id: The party ID of this MHS. Sent in ebXML acknowledgements.
         """
         self.ack_builder = ack_builder
         self.message_parser = message_parser
+        self.callbacks = callbacks
+        self.party_id = party_id
```
alibombaye: I don't quite understand the purpose of the `party_id` but do you feel that this is something that is required to construct the `async_response_handler` or is this something that could be passed in via the `post` method below?

mhs-reference-implementation/mhs/handler/tests/test_async_response_handler.py
```diff
@@ -37,10 +47,21 @@ def test_post(self, mock_get_uuid, mock_get_timestamp):
...
+        self.callbacks[REF_TO_MESSAGE_ID] = mock_callback
 
-        headers = {"Content-Type": 'multipart/related; boundary="--=_MIME-Boundary"'}
-        response = self.fetch("/", method="POST", body=request_body, headers=headers)
+        response = self.fetch("/", method="POST", body=request_body, headers=CONTENT_TYPE_HEADERS)
 
         self.assertEqual(response.code, 200)
         self.assertEqual(response.headers["Content-Type"], "text/xml")
         XmlUtilities.assert_xml_equal(expected_response, response.body)
+        mock_callback.assert_called_with(EXPECTED_MESSAGE)
```
alibombaye: Do you think it's work asserting that whatever the `_send_ack` is doing was also called?

mhs-reference-implementation/mhs/handler/tests/test_client_request_handler.py
```diff
@@ -6,33 +6,43 @@
...
         self.sender = Mock()
         return Application([
-            (r"/.*", ClientRequestHandler, dict(sender=self.sender))
+            (r"/.*", ClientRequestHandler, dict(sender=self.sender, callbacks={}, async_timeout=1))
         ])
 
-    def test_post(self):
-        interaction_name = "interaction"
-        request_body = "A request"
+    def test_post_synchronous_message(self):
```
alibombaye: Is there a test missing here to check the post of the asynchronous message. Which I think at the moment adds to the callable dictionary. Do you think this would be valuable to have?

mhs-reference-implementation/mhs/handler/client_request_handler.py
```diff
@@ -1,31 +1,81 @@
...
 
-            logging.debug("Message sent. Received response: %s", response)
+            interaction_is_asynchronous = async_message_id is not None
+            if interaction_is_asynchronous:
+                self.callbacks[async_message_id] = self._write_async_response
+                logging.debug("Added callback for asynchronous message with ID '%s'", async_message_id)
+
+            immediate_response = self.sender.send_message(interaction_name, message)
+
+            if interaction_is_asynchronous:
```
alibombaye: This if and the if above are the same? Is there a reason for this.

mhs-reference-implementation/mhs/sender/sender.py
```diff
@@ -22,54 +21,80 @@ def __init__(self, interaction_name):
...
         :param interaction_name: The name of the interaction the message is related to.
-        :param message_to_send: A string representing the message to be sent.
-        :return: The content of the response received when the message was sent.
+        :param content: The message content to be sent.
+        :return: A tuple containing the ID of the message (if an asynchronous response is expected) and the content of
+        the response received. The message ID will be None if no asynchronous response is expected.
+        """
+        interaction_details = self._get_interaction_details(interaction_name)
+
+        async_pattern = interaction_details[ASYNC_RESPONSE_EXPECTED]
```
alibombaye: I feel like the name of this variable doesn't give me an indication that what I am expecting here is a bool. Would something like `is_async` or `is_async_pattern` make it bit more clear?

mhs-reference-implementation/mhs/sender/sender.py
```diff
@@ -22,54 +21,80 @@ def __init__(self, interaction_name):
...
-        """Send the specified message for the interaction named.
+    def build_message(self, interaction_name: str, content: str) -> Tuple[str, str]:
+        """Build a message for the specified interaction. Wraps the provided content if required.
 
         :param interaction_name: The name of the interaction the message is related to.
-        :param message_to_send: A string representing the message to be sent.
-        :return: The content of the response received when the message was sent.
+        :param content: The message content to be sent.
+        :return: A tuple containing the ID of the message (if an asynchronous response is expected) and the content of
+        the response received. The message ID will be None if no asynchronous response is expected.
```
alibombaye: So the message_id is being used as a way of identifying whether the message is a sync or async message. Do you think perhaps returning back an additional flag in the return to indicate this would be beneficial?

mhs-reference-implementation/mhs/sender/sender.py
```diff
@@ -22,54 +21,80 @@ def __init__(self, interaction_name):
...
         """
 
         self.interactions_config = interactions_config
         self.message_builder = message_builder
         self.transport = transport
+        self.party_id = party_id
 
-    def send_message(self, interaction_name, message_to_send):
-        """Send the specified message for the interaction named.
+    def build_message(self, interaction_name: str, content: str) -> Tuple[str, str]:
```
alibombaye: Just a little confused. We have a `message_builder` that we use to build a message but then we also have this. Do you think it might be worth clearing as the difference here? From what I understand this looks like a wrapper and the actual message inside here is the one that the message_builder builds?

mhs-reference-implementation/main.py
gareth-allan: I did this because I wanted all requests to the 'supplier' interface to be of the form `/interaction-name`, so I didn't want anything without a leading slash to match. I'm not sure it would make a huge difference though.

mhs-reference-implementation/mhs/handler/async_response_handler.py
gareth-allan: The party ID is the identifier of the MHS instance, so it's something you're assigned when you're registered with Spine/Opentest/etc. In this case I thought it made more sense to have the response handler set up with the value than take it from the "to" field of the incoming messages.

mhs-reference-implementation/mhs/handler/tests/test_async_response_handler.py
gareth-allan: `_send_ack` builds and sends the acknowledgement message in response to the request, so I think the check that `expected_response` equals `response.body` covers that, doesn't it?

mhs-reference-implementation/mhs/handler/tests/test_client_request_handler.py
gareth-allan: Yes there is, and I do think it would be valuable, but this is the case that would test the asynchronous messaging pattern, as I mentioned in the description. Since this requires the asynchronous response to be returned in order for the post method to return I think this is going to require a more involved integration test, which I felt was beyond the scope of this task.

mhs-reference-implementation/mhs/handler/client_request_handler.py
gareth-allan: I wanted to store the callback (only if the message is asynchronous) before sending any requests (which is done for synchronous and asynchronous messages). The second if statement then decides whether to wait for an asynchronous response or return immediately.

mhs-reference-implementation/mhs/sender/sender.py
gareth-allan: Fair point, I'll go with `is_async`.

mhs-reference-implementation/mhs/sender/sender.py
gareth-allan: Yes, I think that's probably clearer, I'll update to do that.

mhs-reference-implementation/mhs/sender/sender.py
gareth-allan: As discussed offline, I'll update this method to be `prepare_message` and make it clearer in the docstring that it's preparing the message received for sending.

mhs-reference-implementation/main.py
alibombaye: ok sounds fair 👍

mhs-reference-implementation/mhs/handler/tests/test_async_response_handler.py
alibombaye: ah ok I see. So would the callback also generate another response? I can see you assert the callback is called. I think perhaps just making it clear that the expected_response and the response here are the acknowledgement responses and not the callback response.

gareth-allan: No, the callback doesn't return anything, which is why I'm only asserting that it's called. I'll update the name of the `expected_response` and `response` variables to (hopefully) make it clear that they are the acknowledgement responses.

mhs-reference-implementation/mhs/handler/tests/test_client_request_handler.py
alibombaye: I see is there not a "simple" unit test that can be added that will prove the callback dictionary is updated when it is determined that the request is of type `async`. I say "simple" I have no idea how to do it myself or how involved that would be. If you feel the integration test would best cover this then happy to leave this as is?

mhs-reference-implementation/mhs/handler/client_request_handler.py
alibombaye: As discussed offline. Wondering if the immediate_response we get back from `spine` can be used to determine whether it is the actual response of just an acknowledgment and using this to determine whether to pause or send the actual response. Could be one to investigate later.

gareth-allan: That's a good call. I'll add this to the PoC documentation around how we're going to expand it for more messaging patterns.

mhs-reference-implementation/mhs/handler/tests/test_client_request_handler.py
gareth-allan: As discussed offline, I haven't come up with a straightforward approach, so I feel that the integration test would be the best way to do this.

PR #21, opened 2019-05-15
Add an Example Supplier Application

gareth-allan:
This PR adds an example supplier application in the `supplier-example` directory. This is a simple flask application that provides the user with a web page they can submit some text to and it generates a GP summary update containing this and sends it to an instance of the MHS PoC.

I've also had to make a few tweaks to the `OneOneOne` and `SCR` projects to make sure they can be imported properly.


OneOneOne/reciever/message_checks/check.py
```diff
@@ -1,7 +1,7 @@
 import logging
 
 from builder.pystache_message_builder import PystacheMessageBuilder
-from OneOneOne.definitions import TEMPLATE_PATH, XML_PATH
+from definitions import TEMPLATE_PATH, XML_PATH
```
Philip-Woods: This change breaks the tests on my machine, it instead refers to the definitions file in the  nhais directory for some reason.

PR #22, opened 2019-06-28
Feature RT-66: SDS Client

Philip-Woods:
This PR contains SDS Lookup implementation for RT-66. Primarily includes the following:
- SDS Client: Handles LDAP lookup for accredited supplier details and MHS details of a supplier. 
- MHSAttributeLookup: Acts mainly as a placeholder wrapper for the SDS Client and what will be the caching of the SDS responses
- Testing for the above implementations, also added an async test wrapper to the common folder allowing us to execute asynchronous tests easily


mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
```diff
@@ -0,0 +1,68 @@
...
+    async def test_get_mhs_lookup_no_connection(self):
+        client = sds.SDSClient()
+        with self.assertRaises(RuntimeError):
+            await client._mhs_details_lookup(ODS_CODE, SERVICE_ID)
+
+    @async_test
+    async def test_tls(self):
+        pass
+        # client = sds.SDSClient(DEV_ADDRESS)
+        # result = await client.get_mhs_details(ODS_CODE, SERVICE_ID)
```
ismail-s: Commented-out code

.gitignore
```diff
@@ -108,3 +108,8 @@ venv.bak/
 
 # PyCharm settings
 .idea
+
+# Certs
```
ghost, line 112: 👍

common/utilities/test_utilities.py
```diff
@@ -0,0 +1,11 @@
+import asyncio
+
+
+def async_test(f):
```
ghost, line 4: As this is a general test utility it's probably worth a docstring to explain what the intent of this method is. Especially since it's not immediately obvious.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
```diff
@@ -0,0 +1,68 @@
...
+    @async_test
+    async def test_get_mhs_lookup_no_connection(self):
+        client = sds.SDSClient()
+        with self.assertRaises(RuntimeError):
+            await client._mhs_details_lookup(ODS_CODE, SERVICE_ID)
+
+    @async_test
+    async def test_tls(self):
+        pass
+        # client = sds.SDSClient(DEV_ADDRESS)
```
ghost: I appreciate that we've still to prove some of the TLS functionality, but what will this test?

mhs-reference-implementation/mhs/SDSLookup/tests/ldap_mocks.py
```diff
@@ -0,0 +1,33 @@
+from pathlib import Path
+
+from ldap3 import Server, Connection, MOCK_ASYNC
+import mhs.SDSLookup.sds as Sds
+from definitions import ROOT_DIR
+
+SERVER_INFO_PATH = str(Path(ROOT_DIR) / 'mhs' / 'SDSLookup' / 'tests' / 'data' / 'my_real_server_info.json')
```
ghost: Probably worth extracting the `mhs/SDSLookup/tests/data/` path to a constant.

mhs-reference-implementation/mhs/SDSLookup/tests/ldap_mocks.py
```diff
@@ -0,0 +1,33 @@
+from pathlib import Path
+
+from ldap3 import Server, Connection, MOCK_ASYNC
+import mhs.SDSLookup.sds as Sds
```
ghost: I think this import should be `sds` (i.e. lowercase). Otherwise it appears that `Sds` is a class, rather than a module.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
+import ssl
```
ghost: PEP-8 says:

> Python packages should also have short, all-lowercase names

So we shouldn't use `SDSLookup` here. Given we know this client will form part of the routing/reliability component, how about `mhs.routing` for the package name? I think `sds.py` is fine for the module name here though.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sds_handler.py
```diff
@@ -0,0 +1,36 @@
...
+
+MHS_OBJECT_CLASS = "nhsMhs"
+
+OPENTEST_SDS_URL = "192.168.128.11"
+PARTY_KEY = " A91461-9199084"
+SERVICE_ID = "urn:nhs:names:services:psis:MCCI_IN010000UK13"
+ORG_CODE = " A91461"
+
+
+class TestSDSHandler(TestCase):
```
ghost: For consistency, this class name should match the name of the class it's testing (`MHSAttributeLookupHandler`). In this instance, I'd be inclined to rename `MHSAttributeLookupHandler` to `SDSHandler`.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
```diff
@@ -0,0 +1,68 @@
+from unittest import TestCase
```
ghost: Following our coding standards, this file should be named `test_sds.py`, as it contains tests for `sds.py`. I'd be equally happy to rename `sds.py` to `sdsclient.py` though.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+# These need replacing as the appropriate TLS certs are received for PTL
+PRIVATE_KEY = Path(ROOT_DIR) / 'data' / 'certs' / 'client.key'
+LOCAL_CERT_FILE = Path(ROOT_DIR) / 'data' / 'certs' / 'client.cert'
+CA_CERTS_FILE = Path(ROOT_DIR) / 'data' / 'certs' / 'client.cert'
+
+
+class SDSClient:
+    mhs_attributes = [
+        'nhsEPInteractionType', 'nhsIDCode', 'nhsMhsCPAId', 'nhsMHSEndPoint', 'nhsMhsFQDN',
+        'nhsMHsIN', 'nhsMHSIsAuthenticated', 'nhsMHSPartyKey', 'nhsMHsSN', 'nhsMhsSvcIA', 'nhsProductKey',
```
ghost: I appreciate that these (appear) to have come from [the story](https://gpitbjss.atlassian.net/browse/RT-66), but is the case of these attributes correct? `nhsMHsSN` vs `nhsMhsSvcIA` looks a little strange?

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(self, ldap_address: str = None, tls: bool = False):
+        if ldap_address:
+            if tls:
+                load_tls = ldap3.Tls(local_private_key_file=PRIVATE_KEY,
+                                     local_certificate_file=LOCAL_CERT_FILE,
+                                     validate=ssl.CERT_REQUIRED, version=ssl.PROTOCOL_TLSv1,
+                                     ca_certs_file=CA_CERTS_FILE)
+
+                self.server = ldap3.Server(ldap_address, use_ssl=True, tls=load_tls)
```
ghost: Does `server` need to be an instance variable? It doesn't appear to be used outside of this constructor (only `connection` is).

Also, I think it might be cleaner to extract the `ldap3.Server(..)` call from the `if tls` block, and simply pass it `use_ssl=tls` and default `load_tls` to `None` . That way you don't have two separate calls to instantiate a `Server`.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+
+class SDSClient:
+    mhs_attributes = [
+        'nhsEPInteractionType', 'nhsIDCode', 'nhsMhsCPAId', 'nhsMHSEndPoint', 'nhsMhsFQDN',
+        'nhsMHsIN', 'nhsMHSIsAuthenticated', 'nhsMHSPartyKey', 'nhsMHsSN', 'nhsMhsSvcIA', 'nhsProductKey',
+        'uniqueIdentifier', 'nhsMHSAckRequested', 'nhsMHSActor', 'nhsMHSDuplicateElimination',
+        'nhsMHSPersistDuration', 'nhsMHSRetries', 'nhsMHSRetryInterval', 'nhsMHSSyncReplyMode'
+    ]
+
+    def __init__(self, ldap_address: str = None, tls: bool = False):
```
ghost: Needs a docstring to define what the arguments do.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+                self.server = ldap3.Server(ldap_address)
+
+            self.connection = ldap3.Connection(self.server, auto_bind=True, client_strategy=ldap3.REUSABLE)
+        else:
+            self.server = None
+            self.connection = None
+
+    async def get_mhs_details(self, nhs_id_code: str, service_id: str):
+        """
+        Returns the all end point details for the given org code and service id
```
ghost: Minor point, but this shouldn't say "the all". Also, we shouldn't refer to an NHS ID code with the variable name, and an org code with the doc string. I think we've tended to use the term "org code" so far.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+        :param nhs_id_code:
+        :param service_id:
+        :return:
+        """
+
+        accredited_system_lookup = await self._accredited_system_lookup(nhs_id_code, service_id)
+
+        if not accredited_system_lookup:
+            raise Exception('No response from accredited system lookup')
```
ghost: I think it's probably worth defining our own `RoutingException` type in this package and documenting that this is raised when no results are returned.

This does raise the question of whether we need to raise an exception here. What scenario does this occur in? Is it if nothing can be found for the id code and service ID provided, or when something has gone wrong with the query?

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # As per the spec exactly one result should be returned, manage error checking here
+        response = accredited_system_lookup[0]
+        party_key = response['attributes']['nhsMHSPartyKey']
```
ghost: Worth a constant for `nhsMHSPartyKey` at least.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+
+        if party_key:
```
ghost: What if `party_key` is none/false?

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+        :param service_id:
+        :return:
+        """
+
+        accredited_system_lookup = await self._accredited_system_lookup(nhs_id_code, service_id)
+
+        if not accredited_system_lookup:
+            raise Exception('No response from accredited system lookup')
+
+        # As per the spec exactly one result should be returned, manage error checking here
```
ghost: What does "manage error checking here" mean?

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+        response = accredited_system_lookup[0]
+        party_key = response['attributes']['nhsMHSPartyKey']
+
+        if party_key:
+            details = await self._mhs_details_lookup(party_key, service_id)
+            return details[0]
+
+    async def _accredited_system_lookup(self, org_code: str, service_id: str):
+        """
+            Used to find an accredited system, the result contains the ASID and nhsMhsPartyKey.
```
ghost: Not sure about the docstring indentation here. Also, given you're filtering the results, they won't contain the ASID, will they?

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # As per the spec exactly one result should be returned, manage error checking here
+        response = accredited_system_lookup[0]
+        party_key = response['attributes']['nhsMHSPartyKey']
+
+        if party_key:
+            details = await self._mhs_details_lookup(party_key, service_id)
+            return details[0]
+
+    async def _accredited_system_lookup(self, org_code: str, service_id: str):
```
ghost: I think the way you've implemented it is probably what we want, but the story still has "An accredited system lookup interface is exposed" and "An MHS lookup interface is exposed" as acceptance criteria and you've used an underscore here to denote that these are non-public methods. Probably worth confirming to make sure the story is in-line with what we want.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+            Used to find an accredited system, the result contains the ASID and nhsMhsPartyKey.
+            These can then be used to find an MHS endpoint
+        :return: endpoint details - filtered to only contain nhsMHSPartyKey
+        """
+        if not self.connection:
+            raise RuntimeError('No connection established')
```
ghost: Might make more sense as an `IOError`.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
+import ssl
+from pathlib import Path
```
ghost: Following the Google standards, this should probably just be `import pathlib`. The same with the `IOLoop` import below.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+        if not self.connection:
+            raise RuntimeError('No connection established')
+
+        search_filter = f"(&(nhsIDCode={org_code}) (objectClass={AS_OBJECT_CLASS}) (nhsAsSvcIA={service_id}))"
+
+        message_id = self.connection.search(search_base=NHS_SERVICES_BASE,
+                                            search_filter=search_filter,
+                                            attributes="nhsMHSPartyKey")
+
+        task = IOLoop.current().run_in_executor(None, self._get_query_result, message_id)
```
ghost: It might be worth extracting this to a 'run task asynchronously' function, as we might end up using it elsewhere, and it would also serve to explain what you're doing a bit more.

Also, I appreciate that this came from the example I provided, but I'm wondering if we could do this without needing a Tornado class? I think the Tornado functions are now a pretty thin wrapper around the built-in async IO functionality, so we should be able to do the same thing with asyncio. This package otherwise doesn't need to know anything about Tornado.

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+                                            search_filter=search_filter,
+                                            attributes=self.mhs_attributes)
+
+        task = IOLoop.current().run_in_executor(None, self._get_query_result, message_id)
+        response, result = await task
+        return response
+
+    def _get_query_result(self, message_id):
+        response, result = self.connection.get_response(message_id)
+        return response, result
```
ghost: Given we don't seem to care about `result`, is it worth returning it?

mhs-reference-implementation/mhs/SDSLookup/sds.py
```diff
@@ -0,0 +1,105 @@
...
+AS_OBJECT_CLASS = "nhsAs"
+
+# These need replacing as the appropriate TLS certs are received for PTL
+PRIVATE_KEY = Path(ROOT_DIR) / 'data' / 'certs' / 'client.key'
+LOCAL_CERT_FILE = Path(ROOT_DIR) / 'data' / 'certs' / 'client.cert'
+CA_CERTS_FILE = Path(ROOT_DIR) / 'data' / 'certs' / 'client.cert'
+
+
+class SDSClient:
+    mhs_attributes = [
```
ghost: Any reason these are a class variable, rather than a constant defined outside of the class?

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
```diff
@@ -0,0 +1,68 @@
...
+        server, connection = mocks.fake_ldap_server()
+
+        client = sds.SDSClient()
+        client.server = server
+        client.connection = connection
+
+        result = await client._mhs_details_lookup(PARTY_KEY, SERVICE_ID)
+
+        self.assertIsNotNone(result)
+        self.assertEqual(result[0]['attributes']['nhsMHSEndPoint'], ['https://vpn-client-1411.opentest.hscic.gov.uk/'])
```
ghost: I think this should probably test that the full list of attributes is returned, since it's one of the acceptance criteria on the story. Also, we need to verify that attributes not on the list are filtered.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
```diff
@@ -0,0 +1,68 @@
...
+    @async_test
+    async def test_accredited(self):
+        server, connection = mocks.fake_ldap_server()
+        client = sds.SDSClient()
+        client.server = server
+        client.connection = connection
+
+        result = await client._accredited_system_lookup(ODS_CODE, SERVICE_ID)
+        assert result
+        assert result[0]['attributes']['nhsMHSPartyKey'] == 'A91461-9199084'
```
ghost: Again, we want to verify that no other attributes have been returned.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
```diff
@@ -0,0 +1,68 @@
...
+PARTY_KEY = "A91461-9199084"
+SERVICE_ID = "urn:nhs:names:services:psis:MCCI_IN010000UK13"
+ODS_CODE = "A91461"
+
+
+class TestSDSClient(TestCase):
+
+    @async_test
+    async def test_mhs_details_lookup(self):
+        server, connection = mocks.fake_ldap_server()
```
ghost: You do this setup step several times. Might be worth extracting it to a method.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
```diff
@@ -0,0 +1,68 @@
...
+    @async_test
+    async def test_get_mhs_lookup(self):
+        server, connection = mocks.fake_ldap_server()
+        client = sds.SDSClient()
+        client.server = server
+        client.connection = connection
+
+        result = await client.get_mhs_details(ODS_CODE, SERVICE_ID)
+        self.assertIsNotNone(result)
+        self.assertEqual(result['attributes']['nhsMHSPartyKey'], 'A91461-9199084')
```
ghost: We need to make sure this result has the full set of attributes, as we want to verify it's done the correct two-stage lookup.

There are also some other bits of logic in `get_mhs_details` that we should cover with these tests. It filters results if there is more than one, so we should verify it does that as expected, and it also raises an exception if there are no results(?).

mhs-reference-implementation/mhs/SDSLookup/tests/test_sds_handler.py
```diff
@@ -0,0 +1,36 @@
...
+
+import mhs.SDSLookup.tests.ldap_mocks as mocks
+from utilities.test_utilities import async_test
+
+NHS_SERVICES_BASE = "ou=services, o=nhs"
+
+MHS_OBJECT_CLASS = "nhsMhs"
+
+OPENTEST_SDS_URL = "192.168.128.11"
+PARTY_KEY = " A91461-9199084"
```
ghost: These look like your actual party key - probably worth anonymising these values.

ghost (changes requested):
I appreciate there are quite a lot of comments here, but I've been fairly detailed with it being the first review, so we can make sure we're all on the same page in terms of approach. Generally, this looks good though!

One point I wanted to raise was regarding logging. I appreciate we've still got an outstanding task to define standards around levels, etc. and the required output format, but we should be trying to ensure that we include appropriate logging as we go, rather than having to include it afterwards. If you use the standard `logging` module with sensible log levels, it should mean it's relatively easy to convert to whatever format we settle on, and we won't have to actually add the log messages later. (CC @ismail-s) One good example here would be that I think we should log at warning level if we find more than one result for an SDS query - the spec says we shouldn't, and we can safely continue using the first result, but we should flag the fact that it's happened.

common/utilities/test_utilities.py
Philip-Woods: Good point, I've added a description of why it's there and how it's used to wrap async tests

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
Philip-Woods: Fair question, I originally wrote it as something I could run to test some TLS things. Since testing this functionality will require particular certs that won't be in the repo, is it safe to remove this test under the assumption it will be covered in an integration test?

mhs-reference-implementation/mhs/SDSLookup/tests/ldap_mocks.py
Philip-Woods: Completed as suggested

mhs-reference-implementation/mhs/SDSLookup/tests/ldap_mocks.py
Philip-Woods: yeah that's right, it fits the code standard like this as well, updated

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Updated module to be named `routing`

mhs-reference-implementation/mhs/SDSLookup/tests/test_sds_handler.py
Philip-Woods: I'll change the the test to match the class name since I think we discussed previously that routing/reliability probably doesn't need to know that the information comes from SDS, just that they're attributes of an MHS. Does that still seem reasonable?

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
Philip-Woods: Updated test module to be named `test_sds`

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: These attributes were picked from the full message whilst testing full LDAP queries and copied into this list/ the story so the casing should be exactly that which the LDAP query returns. When I update the tests to include all attributes I shall double check the attributes returned

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Good point, I've removed server as an instance variable and removed the assignments in the test code as well. Also rearranged the constructor as suggested.

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Added

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Updated docstring, I've also updated the parameter names in this class to be `ods_code` and `interaction_id` since this matches the tickets

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: I added under the assumption it might be possible for a bad ods code/interaction id pair and that we would want to throw an error rather than the exception occurring when we try to index `None` for the party key.
I'll add a custom routing exception for the time being.

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Updated

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: I'll remove this check as we're already checking earlier in the method if the `accredited_system_lookup` returned no response

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Removed, it was a reminder to myself to check for empty/more than 1 responses I believe

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Fixed both

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: The story was updated a couple of hours ago stating that these methods shouldn't be exposed to the rest of the system so I believe the implementation is now correct against the story.

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Updated

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Yeah that's right, I've updated this now.

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: I've swapped the tornado call to use the native async/await calls and as a result I've held off from moving the call into another method

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: binned it

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: Nope, I've moved them outside the class

mhs-reference-implementation/Pipfile
```diff
@@ -9,6 +9,7 @@ verify_ssl = true
 integration-adaptors-common = {editable = true,path = "./../common"}
 requests = "*"
 tornado = "*"
+ldap3 = "*"
```
ismail-s, line 12: General question not specific to this PR, but should we be specifying versions (ie the major version)?

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
Philip-Woods: Updated the test to check for all attributes, however because of the mhs attribute call returning 2 sets of attributes if I check the values for an exact match it will fail half the time (due to responses including a unique id). I've included all the attributes bar the two unique ones and still checked that they are present, just not their values. Does this seem reasonable?

Also added length checking to the attribute list to ensure it contains exactly the number we need

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
Philip-Woods: Check that exactly 1 attribute is returned (party key)

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
Philip-Woods: I'd actually already built a method that does all this in the ldap mocks but forgot to use it 🤦‍♂

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
Philip-Woods: Good point, I've updated this to match the other tests - it now checks for all the attributes and verifies no other attributes are included. 

I'm not totally sure we can verify it filters the multiple responses condition, since we can't reliably determine the order of the endpoints. I have added a test for the exception case where there are no responses though.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sds_handler.py
Philip-Woods: To my knowledge these are useless without my vpn creds alongside this, so this shouldn't be a concern. I've changed them to be fake values and replaced the data in the LDAP test data json's so the tests still work without opentest connection. I'll be more careful with what creds I'm committing next time!

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -0,0 +1,117 @@
...
+NHS_SERVICES_BASE = "ou=services, o=nhs"
+
+MHS_OBJECT_CLASS = "nhsMhs"
+AS_OBJECT_CLASS = "nhsAs"
+MHS_PARTY_KEY = 'nhsMHSPartyKey'
+
+# These need replacing as the appropriate TLS certs are received for PTL
+PRIVATE_KEY = pathlib.Path(ROOT_DIR) / 'data' / 'certs' / 'client.key'
+LOCAL_CERT_FILE = pathlib.Path(ROOT_DIR) / 'data' / 'certs' / 'client.cert'
+CA_CERTS_FILE = pathlib.Path(ROOT_DIR) / 'data' / 'certs' / 'client.cert'
```
ismail-s: If this is the same as the previous constant and may change, then simpler to do:
```suggestion
CA_CERTS_FILE = LOCAL_CERT_FILE
```
for the timebeing

common/utilities/test_utilities.py
```diff
@@ -0,0 +1,19 @@
...
+    By annotating a test method with `@async_test` it will cause the test to wait for asynchronous activities
+    to complete
+    :param f:
+    :return:
+    """
+    def wrapper(*args, **kwargs):
+        coro = asyncio.coroutine(f)
+        future = coro(*args, **kwargs)
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(future)
```
ismail-s: It looks like `asyncio.new_event_loop()` would [be better for test isolation](https://stackoverflow.com/a/23642269).

And then it looks like [`asyncio.run(future)`](https://docs.python.org/3.7/library/asyncio-task.html#asyncio.run) would [be better than that](https://github.com/python/asyncio/pull/465#issue-93620963).

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -0,0 +1,117 @@
...
+        party_key = response['attributes'][MHS_PARTY_KEY]
+
+        details = await self._mhs_details_lookup(party_key, interaction_id)
+
+        if len(details) > 1:
+            logging.warning(f"More than one mhs details  returned on inputs: "
+                            f"ods: {ods_code} - interaction: {interaction_id}")
+        return details[0]['attributes']
+
+    async def _accredited_system_lookup(self, ods_code: str, interaction_id: str):
```
ismail-s: As a general comment (not sure that we have/want a rule around whether we want type hints everywhere or not), if you want to type hint the return type, then you can do `async def blah() -> someType:`

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -0,0 +1,117 @@
...
+                            f"ods: {ods_code} - interaction: {interaction_id}")
+        # As per the spec exactly one result should be returned
+
+        response = accredited_system_lookup[0]
+        party_key = response['attributes'][MHS_PARTY_KEY]
+
+        details = await self._mhs_details_lookup(party_key, interaction_id)
+
+        if len(details) > 1:
+            logging.warning(f"More than one mhs details  returned on inputs: "
```
ismail-s: V. minor: double-space
```suggestion
            logging.warning(f"More than one mhs details returned on inputs: "
```

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -0,0 +1,117 @@
...
+MHS_OBJECT_CLASS = "nhsMhs"
+AS_OBJECT_CLASS = "nhsAs"
+MHS_PARTY_KEY = 'nhsMHSPartyKey'
+
+# These need replacing as the appropriate TLS certs are received for PTL
+PRIVATE_KEY = pathlib.Path(ROOT_DIR) / 'data' / 'certs' / 'client.key'
+LOCAL_CERT_FILE = pathlib.Path(ROOT_DIR) / 'data' / 'certs' / 'client.cert'
+CA_CERTS_FILE = pathlib.Path(ROOT_DIR) / 'data' / 'certs' / 'client.cert'
+
+mhs_attributes = [
```
ismail-s, line 16: How did you find out what ldap queries to do? Is there some documentation, or did you just make some ldap queries?

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
ghost: I think that's fair, yes.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sds_handler.py
ghost: Yes, that's a fair point, I'm happy to go with that.

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -0,0 +1,117 @@
...
+            if tls:
+                load_tls = ldap3.Tls(local_private_key_file=PRIVATE_KEY,
+                                     local_certificate_file=LOCAL_CERT_FILE,
+                                     validate=ssl.CERT_REQUIRED, version=ssl.PROTOCOL_TLSv1,
+                                     ca_certs_file=CA_CERTS_FILE)
+
+            server = ldap3.Server(ldap_address, use_ssl=tls, tls=load_tls)
+            self.connection = ldap3.Connection(server, auto_bind=True, client_strategy=ldap3.REUSABLE)
+        else:
+            self.connection = None
```
ismail-s: Why do you allow there to be no connection here, and then just throw exceptions in the other methods? Why not just throw the exception here, or is it not always required to have the SDS connection?

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -0,0 +1,117 @@
...
+        message_id = self.connection.search(search_base=NHS_SERVICES_BASE,
+                                            search_filter=search_filter,
+                                            attributes=mhs_attributes)
+
+        query_task = self._get_query_result(message_id)
+        response = await query_task
+        return response
+
+    async def _get_query_result(self, message_id):
+        response, result = self.connection.get_response(message_id)
```
ismail-s: AFAICT from looking at the source code of ldap3, this is a blocking operation that by default waits for upto 3 seconds before timing out. So there's a few things:
1. if we timeout, what should we do?
1. what should the timeout be?
1. can we make this method be called in some sort of async way?

mhs-reference-implementation/mhs/routing/sds_handler.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def __init__(self, sds_lookup_address: str = None):
+        self.cache = {}
+        self.sds_client = sds.SDSClient(sds_lookup_address)
+
+    async def retrieve_mhs_attributes(self, org_code, service_id):
+        # Check cache for hits, if no hit retrieve via sds client
+        endpoint_details = await self.sds_client.get_mhs_details(org_code, service_id)
+
+        # cache.add(endpoint_details)
```
ismail-s: Commented-out code

mhs-reference-implementation/mhs/routing/sds_handler.py
```diff
@@ -0,0 +1,16 @@
+
+import mhs.routing.sds as sds
+
+
+class MHSAttributeLookupHandler:
```
ismail-s, line 4: What is the purpose of this class? It looks like it's an incomplete caching layer to me.

mhs-reference-implementation/mhs/routing/tests/data/my_real_server_entries.json
```diff
@@ -0,0 +1,989 @@
...
+                    "top",
+                    "nhsMhs"
+                ],
+                "uniqueIdentifier": [
+                    "S918999410559"
+                ]
+            }
+        }
+    ]
+}
```
ismail-s: Newlines at end of file (and same for other test json files)?

mhs-reference-implementation/mhs/routing/tests/test_sds_handler.py
```diff
@@ -0,0 +1,63 @@
...
+    'nhsMHSRetries': '2',
+    'nhsMHSRetryInterval': 'PT1M',
+    'nhsMHSSyncReplyMode': 'MSHSignalsOnly',
+    'nhsMHsIN': 'MCCI_IN010000UK13',
+    'nhsMHsSN': 'urn:nhs:names:services:psis',
+    # 'nhsMhsCPAId': 'S918999410559', # Exclude this value since there seems to be two responses causing errors on this
+    'nhsMhsFQDN': 'vpn-client-1411.opentest.hscic.gov.uk',
+    'nhsMhsSvcIA': 'urn:nhs:names:services:psis:MCCI_IN010000UK13',
+    'nhsProductKey': '7374',
+    # 'uniqueIdentifier': ['S918999410559']
```
ismail-s: Commented-out code

mhs-reference-implementation/mhs/SDSLookup/tests/test_sds_handler.py
ghost: Yes, I don't think it's the end of the world, but probably good to avoid doing.

mhs-reference-implementation/mhs/routing/tests/test_sds.py
```diff
@@ -0,0 +1,96 @@
...
+    'nhsMHSEndPoint': ['https://vpn-client-1411.opentest.hscic.gov.uk/'],
+    'nhsMHSIsAuthenticated': 'transient',
+    'nhsMHSPartyKey': 'AP4RTY-K33Y',
+    'nhsMHSPersistDuration': 'PT5M',
+    'nhsMHSRetries': '2',
+    'nhsMHSRetryInterval': 'PT1M',
+    'nhsMHSSyncReplyMode': 'MSHSignalsOnly',
+    'nhsMHsIN': 'MCCI_IN010000UK13',
+    'nhsMHsSN': 'urn:nhs:names:services:psis',
+    # 'nhsMhsCPAId': 'S918999410559', # Exclude this value since there seems to be two responses causing errors on this
```
ismail-s: What do you mean by this?

mhs-reference-implementation/mhs/routing/tests/test_sds.py
```diff
@@ -0,0 +1,96 @@
...
+    'nhsMhsFQDN': 'vpn-client-1411.opentest.hscic.gov.uk',
+    'nhsMhsSvcIA': 'urn:nhs:names:services:psis:MCCI_IN010000UK13',
+    'nhsProductKey': '7374',
+    # 'uniqueIdentifier': ['S918999410559']
```
ismail-s: Commented-out code

mhs-reference-implementation/mhs/routing/tests/test_sds.py
```diff
@@ -0,0 +1,96 @@
+from unittest import TestCase
+import mhs.routing.sds as sds
+import mhs.routing.tests.ldap_mocks as mocks
+from utilities.test_utilities import async_test
+import mhs.routing.routing_exception as re
```
ismail-s, line 5: Minor: `re` is the name of the regex stdlib module, so it's a little confusing to use that name.

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -0,0 +1,117 @@
...
+    'uniqueIdentifier', 'nhsMHSAckRequested', 'nhsMHSActor', 'nhsMHSDuplicateElimination',
+    'nhsMHSPersistDuration', 'nhsMHSRetries', 'nhsMHSRetryInterval', 'nhsMHSSyncReplyMode'
+]
+
+
+class SDSClient:
+
+    def __init__(self, ldap_address: str = None, tls: bool = False):
+        """
+        :param ldap_address: The IP address of the directory containing data of the the MHSs
```
ghost: Can you add something to describe the behaviour when no `ldap_address` is provided? I.e. the client is expected to set the `connection` property themselves?

mhs-reference-implementation/mhs/routing/tests/test_sds_handler.py
```diff
@@ -0,0 +1,63 @@
...
+}
+
+
+class TestMHSAttributeLookupHandler(TestCase):
+    opentest_expected_reliability = {'nhsMHSAckRequested': 'always',
```
ismail-s: Is this variable being used?

ismail-s (changes requested):
Bear in mind that I didn't look at Gareth's comments before doing my review.

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -0,0 +1,117 @@
...
+        :return: endpoint details - filtered to only contain nhsMHSPartyKey
+        """
+        if not self.connection:
+            raise IOError('No connection established')
+
+        search_filter = f"(&(nhsIDCode={ods_code}) (objectClass={AS_OBJECT_CLASS}) (nhsAsSvcIA={interaction_id}))"
+
+        message_id = self.connection.search(search_base=NHS_SERVICES_BASE,
+                                            search_filter=search_filter,
+                                            attributes="nhsMHSPartyKey")
```
ghost: Should use `MHS_PARTY_KEY` here.

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -0,0 +1,117 @@
...
+        :return:
+        """
+
+        accredited_system_lookup = await self._accredited_system_lookup(ods_code, interaction_id)
+
+        if not accredited_system_lookup:
+            raise re.RoutingException('No response from accredited system lookup')
+
+        if len(accredited_system_lookup) > 1:
+            logging.warning(f"More than one accredited system details returned on inputs: "
```
ghost: No issues with the log statements added, but we also need to think about how we would trace successful flows through the code, etc. Given log aggregation tools are pretty standard nowadays, our logging approach should be "more is better". As long as we make use of the correct log levels to distinguish between what we're logging, an admin using Splunk, etc. should be able to filter our log messages easily enough.

mhs-reference-implementation/mhs/routing/tests/ldap_mocks.py
```diff
@@ -0,0 +1,33 @@
...
+import mhs.routing.sds as sds
+from definitions import ROOT_DIR
+
+MHS_TEST_DATA_PATH = Path(ROOT_DIR) / 'mhs' / 'routing' / 'tests' / 'data'
+SERVER_INFO_PATH = str(MHS_TEST_DATA_PATH / 'my_real_server_info.json')
+SCHEMA_PATH = str(MHS_TEST_DATA_PATH / 'my_real_server_schema.json')
+SERVER_ENTRIES_PATH = str(MHS_TEST_DATA_PATH / 'my_real_server_entries.json')
+
+
+def fake_ldap_server():
```
ghost: This method name should probably be `fake_ldap_connection` now.

mhs-reference-implementation/mhs/routing/tests/test_sds.py
```diff
@@ -0,0 +1,96 @@
...
+        client = mocks.mocked_sds_client()
+
+        attributes = await client.get_mhs_details(ODS_CODE, SERVICE_ID)
+
+        # check values present
+        for key, value in expected_mhs_attributes.items():
+            self.assertEqual(value, attributes[key])
+
+        # Assert exact number of attributes, minus the unique values
+        self.assertEqual(len(attributes) - 2, len(expected_mhs_attributes))
```
ghost: A minor point, but this logic is the reverse of the same thing from the `test_mhs_details_lookup` case above. It would be clearer for them both to do the same assertion.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
ghost: I think the fact that our dummy LDAP response contains multiple results for the same ODS code/party key is enough to cover the filtering.

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
ghost: I think this does the job, but isn't the MHS attribute call returning 2 sets of attributes because that's what we have in the dummy LDAP server data? Couldn't we edit it so that it doesn't return two sets for this query?

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
Philip-Woods: Removed in a previous commit

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Ultimately these will be different files, I'll update the paths to demonstrate this and I've added a constant holding the certs dir path

common/utilities/test_utilities.py
Philip-Woods: I've replaced the `loop = ...` and `loop.run_until_complete(future)` with just `asyncio.run(future)`. Is this the correct interpretation that we don't need `asyncio.new_event_loop()`?

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: 👍

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: 👍

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Nothing particularly special in terms of ldap queries, just filtering on the interaction id, object code and ods code/org code as suggested here:
https://nhsconnect.github.io/gpconnect/integration_spine_directory_service.html

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Perhaps this wasn't the correct approach but I allowed empty connections here so I could inject mocks as part of the tests without the tests erroring on instantiation. Any better ideas?

mhs-reference-implementation/mhs/SDSLookup/tests/test_sdsclient.py
Philip-Woods: This is a good point, I've removed the duplicate data and added all attribute checks in so the test should be complete now.

mhs-reference-implementation/mhs/routing/sds.py
ismail-s: I think we should start including `logger = logging.getLogger(__name__)` in files that we log in, so as to include the location of the file the log message came from. This is a standard line from [the Python website](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial).

ghost: Yes, agreed.

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Removed these completely actually

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: As discussed this has been altered so that the constructor takes a connection object and null value checking is completed there. A `connectionFactory` has been implemented to produce connections and make the implementation avoid having a constructor with empty values.

briandiggle:
I'm noticing in the code that there is a mixture of terminology used for the interation ID. It is called "Service ID" in places, and "interaction ID" in others. Can we standardise on "interaction ID" as this is commonly accepted term.

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: As discussed this call has been wrapped in a `loop.run_in_executor` call and awaited. Timeout has been added as a parameter to the constructor and the whole call has been wrapped an exception block to catch the timeout error with appropriate logging.

mhs-reference-implementation/mhs/routing/sds_handler.py
Philip-Woods: This was intended as a placeholder for the cache management and would be changed in the next pr, for the sake of completeness thought I've binned it

mhs-reference-implementation/mhs/routing/sds_handler.py
Philip-Woods: This is exactly what it is, in the next PR this handler manages the cache checking and calls to the sds client depending on what the cache returns

mhs-reference-implementation/mhs/routing/tests/data/my_real_server_entries.json
Philip-Woods: Added ending lines

mhs-reference-implementation/mhs/routing/tests/test_sds_handler.py
Philip-Woods: Fixed

mhs-reference-implementation/mhs/routing/tests/test_sds.py
Philip-Woods: The sds call should return exactly 1 value for this lookup, however my opentest instance seems to return 2, this had been copied to the test data and caused an issue due to this value being an unique identifier. I've deleted the second instance in the test data so I've readded these attributes

mhs-reference-implementation/mhs/routing/tests/test_sds.py
Philip-Woods: Fixed

mhs-reference-implementation/mhs/routing/tests/test_sds.py
Philip-Woods: Updated to just use the module name

mhs-reference-implementation/mhs/routing/tests/test_sds_handler.py
Philip-Woods: Good point, it's no longer being used since we're using the full attribute set defined out of the class

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Since the sds client now takes a connection object this no longer applies I believe.

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Updated

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: I've added the module reference Ismail suggested and added a couple extra logs around message IDs. Noted that we should try to consist logs throughout

mhs-reference-implementation/mhs/routing/tests/ldap_mocks.py
Philip-Woods: Done

mhs-reference-implementation/mhs/routing/tests/test_sds.py
Philip-Woods: Updated to do the assertions in the same order

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Added this

mhs-reference-implementation/mhs/SDSLookup/sds.py
Philip-Woods: As discussed I've wrapped the call within the `get_query_result` method to use the asyncio `run_in_executor` method

mhs-reference-implementation/mhs/routing/tests/test_sds.py
```diff
@@ -84,13 +81,7 @@ class TestSDSClient(TestCase):
             await client.get_mhs_details("fake code", "fake interaction")
 
     @async_test
-    async def test_accredited_no_connection(self):
-        client = sds.SDSClient()
-        with self.assertRaises(IOError):
-            await client._accredited_system_lookup(ODS_CODE, SERVICE_ID)
+    async def test__no_connection(self):
```
ghost: There's now a double underscore in this method name.

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -27,25 +23,17 @@
...
-            self.connection = None
-
-    async def get_mhs_details(self, ods_code: str, interaction_id: str):
+        if not sds_connection:
+            raise ValueError('sds_connection must not be null')
+
+        self.connection = sds_connection
+        self.timeout = timeout
+
+    async def get_mhs_details(self, ods_code: str, interaction_id: str) -> Dict:
```
ismail-s, line 36: Another point: `Dict` can be made more specific like eg `Dict[keyType, valueType]`. As can other container types eg `List[str]`

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -56,62 +44,70 @@ def __init__(self, ldap_address: str = None, tls: bool = False):
         accredited_system_lookup = await self._accredited_system_lookup(ods_code, interaction_id)
 
         if not accredited_system_lookup:
-            raise re.RoutingException('No response from accredited system lookup')
+            raise routing_exception.RoutingException('No response from accredited system lookup')
```
ghost, line 47: I think we need an error level log message here.

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -56,62 +44,70 @@ def __init__(self, ldap_address: str = None, tls: bool = False):
...
+            f" (nhsMhsSvcIA={interaction_id}))"
         message_id = self.connection.search(search_base=NHS_SERVICES_BASE,
                                             search_filter=search_filter,
                                             attributes=mhs_attributes)
 
+        logger.info(f'Message id - {message_id} - for query: party key - {party_key} '
+                    f': interaction id - {interaction_id}')
+
         query_task = self._get_query_result(message_id)
         response = await query_task
```
ghost: I think it's worth having a log message to identify that the query was successful here. This would apply to the `_accredited_system_lookup` method too.

ghost (commented):
Generally looks good, just a couple of comments on logging that I think we should address before this goes in.

mhs-reference-implementation/mhs/routing/tests/test_sds.py
Philip-Woods: Binned it

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Added this, I've also added empty response checking for the `_mhs_details_lookup` response with appropriate logging

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Added this

ghost (dismissed):

mhs-reference-implementation/mhs/routing/sds.py
```diff
@@ -27,91 +23,95 @@
...
+        logger.info(f'Found accredited supplier details for message_id: {message_id}')
 
-        query_task = self._get_query_result(message_id)
-        response = await query_task
         return response
 
-    async def _mhs_details_lookup(self, party_key: str, interaction_id: str):
+    async def _mhs_details_lookup(self, party_key: str, interaction_id: str) -> List:
         """
         Given a party key and a service, this will return an object containing the attributes of that party key,
```
ismail-s: I'm assuming interaction id and service mean the same thing (not familiar with all this terminology yet).

mhs-reference-implementation/mhs/routing/sds_connection_factory.py
```diff
@@ -0,0 +1,31 @@
+import ssl
+import ldap3
+
+
+def build_sds_connection(ldap_address: str) -> ldap3.Connection:
+    """
+    Given an ip address this will return a ldap3 connection object
```
ismail-s: Pedantic/minor: do you have to pass an ip address or any address?

ismail-s (dismissed):

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: Yep, I'd noticed this when looking at type hints, I refrained from using it here since the value type can be more than 1 type but I think in the future we should probably include the key/value types

mhs-reference-implementation/mhs/routing/sds.py
Philip-Woods: This is correct, I will change this as Brian has specified previously that we should try to stick to interaction id, must have missed this one!

mhs-reference-implementation/mhs/routing/sds_connection_factory.py
Philip-Woods: Looking at the ldap lib it seems like any valid ldap address with be accepted, I'll update to reflect this

ismail-s (approved):

mhs-reference-implementation/mhs/routing/sds.py
ismail-s: In that case, you could use `Any`, or `Union[someType, someOtherType]`, though that could get long-winded.

PR #25, opened 2019-07-04
Feature/RT55 - Cache Adapter

Philip-Woods:
This includes an interface for a cache used by the `MHSAttributeLookupHandler` to avoid constant lookups to the SDS. Changes include:
- AttributeLookupHandler now takes an `AbstractCacheAdaptor` as a part of the constructor.
- `retrieve_mhs_attributes` in the lookup handler now queries the queue before calling the sds client
- A dictionary-based implementation of the `AbstractCacheAdaptor` has been provided

### DIFF
--- a/mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
+++ b/mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
@@ -23,7 +23,7 @@ class AbstractMHSCacheAdaptor(abc.ABC):
         raise NotImplementedError()
 
     @abc.abstractmethod
-    def add_cache_value(self, ods_code: str, interaction_id: str, value) -> None:
+    async def add_cache_value(self, ods_code: str, interaction_id: str, value: Dict) -> None:
         """
         Adds a value to the cache, recording the input time used to determine when values have expired
         """
--- a/mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
+++ b/mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
@@ -1,33 +1,30 @@
 import abc
+from typing import Optional, Dict
+
+FIFTEEN_MINUTES_IN_SECONDS = 900
 
 
 class AbstractMHSCacheAdaptor(abc.ABC):
 
-    def __init__(self, expiry_time=15):
+    def __init__(self, expiry_time=FIFTEEN_MINUTES_IN_SECONDS):
         """
-        :param expiry_time: Time for a value to expire in minutes
+        :param expiry_time: Time for a value to expire in seconds
         """
+        if expiry_time < 0:
+            raise ValueError('Expiry time must not be non-negative')
         self.expiry_time = expiry_time
-        pass
 
     @abc.abstractmethod
-    async def retrieve_mhs_attributes_value(self, ods_code: str, interaction_id: str):
+    async def retrieve_mhs_attributes_value(self, ods_code: str, interaction_id: str) -> Optional[Dict]:
         """
         Given a key the cache should return either the cached value or None if the value is not found or if the value is
         out of date
-        :param ods_code:
-        :param interaction_id:
-        :return:
         """
         raise NotImplementedError()
 
     @abc.abstractmethod
-    def add_cache_value(self, ods_code: str, interaction_id: str, value):
+    def add_cache_value(self, ods_code: str, interaction_id: str, value) -> None:
         """
-
-        :param ods_code:
-        :param interaction_id:
-        :param value:
-        :return:
+        Adds a value to the cache, recording the input time used to determine when values have expired
         """
         raise NotImplementedError()
--- a/mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
+++ b/mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
@@ -1,0 +1,33 @@
+import abc
+
+
+class AbstractMHSCacheAdaptor(abc.ABC):
+
+    def __init__(self, expiry_time=15):
+        """
+        :param expiry_time: Time for a value to expire in minutes
+        """
+        self.expiry_time = expiry_time
+        pass
+
+    @abc.abstractmethod
+    async def retrieve_mhs_attributes_value(self, ods_code: str, interaction_id: str):
+        """
+        Given a key the cache should return either the cached value or None if the value is not found or if the value is
+        out of date
+        :param ods_code:
+        :param interaction_id:
+        :return:
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def add_cache_value(self, ods_code: str, interaction_id: str, value):
+        """
+
+        :param ods_code:
+        :param interaction_id:
+        :param value:
+        :return:
+        """
+        raise NotImplementedError()
--- a/mhs-reference-implementation/mhs/routing/dictionary_cache.py
+++ b/mhs-reference-implementation/mhs/routing/dictionary_cache.py
@@ -5,7 +5,7 @@ from typing import Dict, Optional
 FIFTEEN_MINUTES_IN_SECONDS = 900
 
 
-def _generate_key(ods_code: str, interaction_id: str):
+def _generate_key(ods_code: str, interaction_id: str) -> str:
     return ods_code + interaction_id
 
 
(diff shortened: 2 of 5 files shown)


mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
```diff
@@ -0,0 +1,30 @@
...
+    @abc.abstractmethod
+    async def retrieve_mhs_attributes_value(self, ods_code: str, interaction_id: str) -> Optional[Dict]:
+        """
+        Given a key the cache should return either the cached value or None if the value is not found or if the value is
+        out of date
+        """
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def add_cache_value(self, ods_code: str, interaction_id: str, value) -> None:
```
ismail-s: Is the type of `value` just going to always be `Dict`? If so, could add it as a type hint.

mhs-reference-implementation/mhs/routing/dictionary_cache.py
```diff
@@ -0,0 +1,59 @@
+import mhs.routing.abstract_cache_adapter as aca
+import time
+from typing import Dict, Optional
+
+FIFTEEN_MINUTES_IN_SECONDS = 900
+
+
+def _generate_key(ods_code: str, interaction_id: str) -> str:
+    return ods_code + interaction_id
```
ismail-s: What does an `ods_code` and an `interaction_id` look like? ie is there any chance that you could have different pairs of (ods_code, interaction_id) where the concatenation is the same (doubt it, but just double-checking)?

mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
```diff
@@ -0,0 +1,30 @@
...
+    def __init__(self, expiry_time=FIFTEEN_MINUTES_IN_SECONDS):
+        """
+        :param expiry_time: Time for a value to expire in seconds
+        """
+        if expiry_time < 0:
+            raise ValueError('Expiry time must not be non-negative')
+        self.expiry_time = expiry_time
+
+    @abc.abstractmethod
+    async def retrieve_mhs_attributes_value(self, ods_code: str, interaction_id: str) -> Optional[Dict]:
```
ismail-s, line 18: Why is this `async def` instead of just `def`? And same question for the methods in `dictionary_cache.py`. And `add_cache_value` is `def` in this file and `async def` in `dictionary_cache.py`.

mhs-reference-implementation/mhs/routing/tests/test_dictionary_cache.py
```diff
@@ -0,0 +1,77 @@
...
+
+        await cache.add_cache_value("code", "int", "check123")
+        value = await cache.retrieve_mhs_attributes_value("code", "int")
+
+        self.assertIsNotNone(value)
+        self.assertEqual(value, "check123")
+
+    @async_test
+    async def test_timeout(self):
+        cache = dict_cache.DictionaryCache(1)  # Set timeout to 1m
```
ismail-s:
```suggestion
        cache = dict_cache.DictionaryCache(1)  # Set timeout to 1s
```

mhs-reference-implementation/mhs/routing/tests/test_dictionary_cache.py
```diff
@@ -0,0 +1,77 @@
...
+        self.assertTrue(not cache.cache)
+
+        await cache.add_cache_value("code", "int", "check123")
+        value = await cache.retrieve_mhs_attributes_value("code", "int")
+
+        self.assertIsNotNone(value)
+        self.assertEqual(value, "check123")
+
+    @async_test
+    async def test_timeout(self):
```
ismail-s: Test name implies you're testing when a cache entry expires, when what you're actually testing is when the cache entry hasn't expired.

mhs-reference-implementation/mhs/routing/tests/test_dictionary_cache.py
```diff
@@ -0,0 +1,77 @@
...
+        self.assertIsNotNone(value)
+        self.assertEqual(value, "check123")
+
+    @async_test
+    async def test_message_should_timeout(self):
+        half_second = 0.5
+        cache = dict_cache.DictionaryCache(half_second)
+        await cache.add_cache_value("code", "int", "check123")
+
+        time.sleep(1)
```
ismail-s: Sleeping in these tests, relying on system time, feels a bit naff to me. Faster/safer would be to use the `patch` decorator, see [here](https://docs.python.org/3/library/unittest.mock.html#quick-guide) for a quick guide.

mhs-reference-implementation/mhs/routing/tests/test_sds_handler.py
```diff
@@ -39,17 +41,75 @@
 
 class TestMHSAttributeLookupHandler(TestCase):
 
+    def setUp(self) -> None:
+        self.cache = dc.DictionaryCache()
+        self.mocked_client = mocks.mocked_sds_client()
```
ismail-s: Unused variable

mhs-reference-implementation/mhs/routing/tests/test_sds_handler.py
```diff
@@ -39,17 +41,75 @@
...
+            sds.MHSAttributeLookupHandler(mocks.mocked_sds_client(), None)
+
+    @async_test
+    async def test_value_added_to_cache(self):
+        handler = sds.MHSAttributeLookupHandler(mocks.mocked_sds_client(), self.cache)
+        handler.sds_client = mocks.mocked_sds_client()
+        future = asyncio.Future()
+        future.set_result(None)
+
+        handler.cache.add_cache_value = MagicMock()
```
ismail-s: Why not just have a `MagicMock` cache, instead of setting all the parts of the cache to be `MagicMock`s. `MagicMock` recursively creates more mocks so you can do `MagicMock().this.iss.still.a.mock`

mhs-reference-implementation/mhs/routing/tests/test_sds_handler.py
```diff
@@ -39,17 +41,75 @@
...
+        client = MagicMock()
+        future = asyncio.Future()
+        future.set_result({'attributes': 'testData'})
+
+        client.get_mhs_details.return_value = future
+        handler = sds.MHSAttributeLookupHandler(client, cache)
+
+        await handler.cache.add_cache_value("check", "not", "added")  # value in cache for 3 seconds
+
+        time.sleep(2)
```
ismail-s: Again, better to avoid sleeping in tests https://github.com/nhsconnect/integration-adaptors/pull/25#discussion_r300331419.

ismail-s (changes requested):

mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
Philip-Woods: Added the type hint

mhs-reference-implementation/mhs/routing/dictionary_cache.py
Philip-Woods: In theory it's possible, I think it would be super unlikely in reality but I will add a seperating `-` just to ensure this won't happen.

mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
Philip-Woods: I've updated the `add_cache_value` in the abstract class to be async.

My original thought was that they should all be `async`, since depending on the implementation of the cache it could involve blocking calls.  Does this seem a valid assumption?

mhs-reference-implementation/mhs/routing/tests/test_dictionary_cache.py
Philip-Woods: 👍

mhs-reference-implementation/mhs/routing/tests/test_dictionary_cache.py
Philip-Woods: Updated the name appropriately

mhs-reference-implementation/mhs/routing/tests/test_sds_handler.py
Philip-Woods: By default `MagicMock` doesn't work with async and can't be awaited, so I have to specify the methods and that they return a future.

mhs-reference-implementation/mhs/routing/tests/test_sds_handler.py
Philip-Woods: Fixed as suggested

mhs-reference-implementation/mhs/routing/abstract_cache_adapter.py
ismail-s: I think so.

ismail-s (dismissed):

ismail-s (approved):

PR #26, opened 2019-07-04
RT-54 vpn setup

ismail-s:
Create a Python [Fabric](https://www.fabfile.org/) script that connects to an EC2 instance and sets up a VPN connection to Opentest.

At the moment, the idea is that this script would be run manually, but it shouldn't be too difficult to automate this if necessary.


pipeline/scripts/setup-ec2-openvpn/README.md
```diff
@@ -0,0 +1,9 @@
+# EC2 OpenVPN Setup Script
+
+This is a Fabric script that can be used to setup an EC2 instance (specifically, one that has been created by ECS) with an OpenVPN connection.
```
ghost: Worth adding the link to [Fabric](https://www.fabfile.org/) in this readme?

pipeline/scripts/setup-ec2-openvpn/openvpn-down.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+# Copyright (c) 2007-2008 Roy Marples <roy@marples.name>
+# Amended by Kristian Gunstone for manjaro-openrc
+# Released under the 2-clause BSD license.
```
ghost: Where have these scripts come from? https://opensource.org/licenses/BSD-2-Clause implies that there should be more to this license statement.

ghost (commented):
Not been through this in-depth yet, but a couple of comments from a quick look.

pipeline/scripts/setup-ec2-openvpn/openvpn-down.sh
ismail-s: I took the files from https://gist.github.com/gammy/b821ff02abfb23d0e41524f96d8a1b0a and https://gist.github.com/gammy/23c071e62f6f93ea8d1b42e0c110d8c9, as mentioned at the bottom of each file.

Looking online a bit more, I think a better source of these files is [here](https://github.com/OpenRC/openrc/tree/fee2ffe559bc39beec16585daf557b902a53137b/support/openvpn). The up.sh/down.sh files there specifically reference the [LICENSE file at the root of that repo](https://github.com/OpenRC/openrc/blob/fee2ffe559bc39beec16585daf557b902a53137b/LICENSE), which is the BSD-2 license.

If this all seems ok with you, I'll switch to using those files instead.

ghost: Sorry, I missed the notes at the bottom of those files! It sounds like it's worth switching to those files.

ismail-s: I've made these changes. I'll test them tomorrow morning.

ismail-s: I've tested the changes in the new AWS account we've got, and they work as expected (as long as the VPN connection is setup before running the task, and as long as the up/down bash scripts have LF line endings, not CRLF).

Philip-Woods (approved):

PR #28, opened 2019-07-09
RT-52 Setup Jenkins server

ismail-s:
This is quite a dense PR. It contains scripts that build the Jenkins server and worker that we're using.

- [x] Add appropriate documentation
- [x] Re-organise the files in this PR

PR #29, opened 2019-07-10
Feature/deploy containers

Philip-Woods:
This PR contains all the necessary packer, Jenkins and Terraform scripting in order to publish our builds to ECR, run the build as part of our pipeline and stand up the MHS in a ECS instance that we can run our integration tests against. Finally the MHS build is pulled down (again with Terraform)


Jenkinsfile
```diff
@@ -20,5 +22,42 @@ pipeline {
...
+                        """
+                }
+            }
+        }
+
+        stage('Integration Tests') {
+            steps {
+                // Wait for MHS container to fully stand up
+                // TODO: Run actual integration tests.
+                sh label: 'Ping MHS', script: 'sleep 20; curl ${MHS_ADDRESS}'
```
ismail-s, line 45: A better way of waiting for MHS to stand up is:
```bash
until curl -s -o /dev/null ${MHS_ADDRESS}; do sleep 1; done
```
Note that I haven't tested that exact command. And no, I didn't come up with this, I noticed it in [this article](https://aws.amazon.com/blogs/compute/amazon-ecs-and-docker-volume-drivers-amazon-ebs/).

Actually, hold on a second, that would just mean that we could have a job running forever. To fix that, could add a [timeout](https://stackoverflow.com/a/38106655) on the command.

Or alternatively just leave it as-is, as it's probably good enough for the timebeing.

.gitignore
```diff
@@ -109,10 +109,17 @@ venv.bak/
...
+# Terraform
+.terraform
+*.tfstate*
+
+
 # Certs
 **/*.cert
 **/*.key
 **/*.pem
+**/certs/party_key.txt
```
ismail-s, line 121: Optional: perhaps we should just ignore the folder as a whole

Jenkinsfile
```diff
@@ -20,5 +22,42 @@ pipeline {
...
         }
+
+        stage('Deploy') {
+            steps {
+                dir('pipeline/terraform/test-environment') {
+                    sh label: 'Initialising Terraform', script: 'terraform init -input=false'
+                    sh label: 'Applying Terraform configuration', script: """
+                            terraform apply -auto-approve \
+                            --var cluster_id=${CLUSTER_ID} \
+                            -var ecs_address=${DOCKER_REPOSITORY} \
```
ismail-s: Minor: Does it make any difference doing `-var` vs `--var`? Otherwise, for consistency (and nothing else), could just do them all in the same way (along with the `-auto-approve`).

pipeline/terraform/test-environment/variables.tf
```diff
@@ -0,0 +1,8 @@
+variable "region" {
+  default = "eu-west-2"
+}
+
+variable "cluster_id" {}
+variable "task_execution_role" {}
+variable "build_id" {}
+variable "ecs_address" {}
```
ismail-s: I think this should be:
```suggestion
variable "ecr_address" {}
```

pipeline/terraform/test-environment/test-environment.tf
```diff
@@ -0,0 +1,63 @@
+provider "aws" {
+  profile = "default"
+  region = var.region
+}
+
+resource "aws_ecs_task_definition" "test-environment-mhs-task" {
+  family = "${var.build_id}-mhs-task"
```
ismail-s: Optional/v. minor: could look a bit nicer with the build id at the end instead of at the beginning.

pipeline/terraform/test-environment/test-environment.tf
```diff
@@ -0,0 +1,63 @@
...
+  cpu = "256"
+  memory = "512"
+  requires_compatibilities = [
+    "EC2"
+  ]
+  execution_role_arn = var.task_execution_role
+}
+
+resource "aws_ecs_service" "test-environment-service" {
+  name = "${var.build_id}-service"
```
ismail-s, line 58: Service name would be a bit nicer with `mhs` in it somewhere.

pipeline/terraform/test-environment/README.md
```diff
@@ -0,0 +1,35 @@
...
+## Pre-requisites
+In order to function, this configuration expects several components to exist in the AWS account being used: 
+
+- An IAM role for the 'Elastic Container Service Task' trusted entity with the built-in
+  `AmazonECSTaskExecutionRolePolicy`. You can use the `ecsTaskExecutionRole` created for you automatically by AWS when
+  creating a task definition from the console, if available. This ARN of this role is passed as the
+  `task_execution_role` input variable.
+- An IAM role for the EC2 instance/etc. running the build pipeline. This must grant the `ecs:RegisterTaskDefinition`
+  ,`ecs:CreateService`, `iam:PassRole`, `ecs:DescribeTaskDefinition`, `ecs:DeregisterTaskDefinition`, 
+  `ecs:DescribeServices`, `ecs:UpdateService` & `ecs:DeleteService` permissions to the build machine.
```
ismail-s, line 18: What's the difference between this list and the one in `pipeline/README.md`, both say they're for the build pipeline worker thing.

ismail-s (commented):
Only realised now, but in #28, I've also made changes to `pipeline/README.md`, so there'll be a bit of merge conflict resolving needed to make it read nicely.

Jenkinsfile
Philip-Woods: Yeah I tried a few different things with Jenkin's `waitUntil` but it got a little messy, since jenkins steps will fail completely if the script fails anyway (curl will throw an exit code 7 if port 80 is refused, causing the step to fail). 

I probably could have wrapped the script in try catch but for the time being I think this will suffice, I think we'll have to modify this if we stand up multiple services (caches / queues /different components) but it works for the time being

.gitignore
Philip-Woods: I think in the future this will probably be removed as the party key will be passed in as part of the config manager, I'll leave this for the time being since regardless of where the certs are we will always want to ignore them.

pipeline/terraform/test-environment/variables.tf
Philip-Woods: Updated

pipeline/terraform/test-environment/test-environment.tf
Philip-Woods: Moved

pipeline/terraform/test-environment/test-environment.tf
Philip-Woods: updated

pipeline/terraform/test-environment/README.md
Philip-Woods: This list I think Garreth added specifically for Terraform to start, the permissions in `pipeline/README.md` also contains the permission for Packer to publish to ECR.

PhilSmithBJSS (approved):

PR #30, opened 2019-07-12
Feature/rt58 test execution script

PhilSmithBJSS:


Jenkinsfile
```diff
@@ -40,9 +40,12 @@ pipeline {
 
         stage('Integration Tests') {
             steps {
-                // Wait for MHS container to fully stand up
-                // TODO: Run actual integration tests.
-                sh label: 'Ping MHS', script: 'sleep 20; curl ${MHS_ADDRESS}'
+                dir('mhs-reference-implementation') {
+                    // Wait for MHS container to fully stand up
+                    // TODO: Run actual integration tests.
```
Philip-Woods: Can probably remove this TODO line now!

mhs-reference-implementation/selenium_tests/page_objects/methods.py
```diff
@@ -1,19 +1,20 @@
...
+    # The asid should be set in the 'Environment variables' section of the Run/Debug Configurations
+    # ...if this is not set, it will default to '1234567890', which will cause test failures!)
+    return os.environ.get('ASID', os.environ.get('ASID', 1234567890))
 
-    return asid
+
+def get_hostname():
+    # The hostname should be set in the 'Environment variables' section of the Run/Debug Configurations
+    # ...if this is not set, it will default to 'http://localhost/'
+    return "http://" + os.environ.get('MHS_ADDRESS', 'localhost') + "/"
```
Philip-Woods, line 17: Can I suggest we record the details (and maybe an example for pycharm) about which environment variables we need and the format they should take (so mhs address should be a 'bare' url with no http prefix or '/' postfix) in a README for the integration tests. Just a bit easier to find than in the code

Philip-Woods (commented):
Just a couple things on documentation/comments but all good

Philip-Woods (approved):

PR #31, opened 2019-07-12
Feature - RT76 Timing decorator

Philip-Woods:
This PR adds several decorator methods that can be used to surround method calls with a timing framework that allows us to measure the time taken to execute a method. Additional methods are provided to add decorators specific to tornado to log the inbound messages.

Each timing method is duplicated with an asynchronous version to handle cases of `async` defined methods


common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,159 @@
...
+
+        time.sleep(2)
+        await task
+        self.assertEqual(len(log_mock.call_args_list), 1)
+
+        call = log_mock.call_args_list[0]
+        args, kwargs = call
+        input_dict = args[0]
+
+        self.assertTrue(math.isclose(input_dict["duration"], 1, rel_tol=0.15))
```
Philip-Woods: I know in previous PRs we've mentioned to avoid the usage of actual `sleep` calls in favor of patching the library, however I found there was varying behavior of the timing decorator between real sleeps and patched sleeps so it felt wrong not to include a full unit test with the realistic/non patched scenario. Any better ideas would be appreciated

common/utilities/timing.py
```diff
@@ -0,0 +1,127 @@
...
+import time
+from functools import wraps
+
+
+class StopWatch:
+
+    start_time = None
+
+    def start_timer(self):
+        self.start_time = time.time()
```
ghost: I think we'd be better using [`time.perf_counter`](https://docs.python.org/3/library/time.html#time.perf_counter) here (and in `stop_timer`), as it's aimed at this use-case.

common/utilities/timing.py
```diff
@@ -0,0 +1,127 @@
...
+            duration = round(duration, 3)
+            logging.info({
+                'method': method.__name__,
+                'duration': duration
+            })
+
+    return invoke_method_with_timer
+
+
+def log_request(method):
```
ghost: Probably worth naming this method (and `log_request_async`) to reflect the fact that we're doing timing. Perhaps `time_request`/`time_request_async`, for consistency with the other methods provided?

ghost: Do we need these Tornado-specific methods? From a quick check of the existing MHS code, Tornado already logs the following for me when a request completes:

```
404 POST /foo (::1) 1.29ms
```

Is the goal that we can use these to ensure a consistent logging format for request handler methods?

common/utilities/timing.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # Gets the 'self' of the caller
+        handler = args[0]
+
+        log_details = {
+            'handler': handler.__class__.__name__,
+            'method': handler.request.method.lower()
+        }
+
+        if handler.request.body:
```
ghost: I'm not sure we should be logging the full body of requests. I can foresee that these may contain patient data that we don't want to log. Also, if we do keep this, we should presumably have the same `if handler.request.body` block in `log_request` above.

common/utilities/timing.py
```diff
@@ -0,0 +1,127 @@
...
+    def start_timer(self):
+        self.start_time = time.time()
+
+    def stop_timer(self):
+        end_time = time.time()
+        diff = end_time - self.start_time
+        return diff
+
+
+def time_method(method):
```
ghost: There's quite a lot of duplication between these methods (especially in the blocks that calculate the duration of the timer and print it to the log). I think it would be worth trying to extract some of this logic into separate reusable methods.

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,159 @@
...
+        return Application([
+            (r"/.*", FakeRequestHandler, {})
+        ])
+
+    @patch('time.sleep')
+    @patch('common.utilities.timing.StopWatch.stop_timer')
+    @patch('logging.info')
+    def test_post_synchronous_message(self, log_mock, time_mock, sleep_mock):
+        time_mock.return_value = 5
+        expected_response = "Hello world!"
```
ghost: This looks like code taken from another test class.

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,159 @@
...
+
+class TestHTTPWrapperTimeUtilities(AsyncHTTPTestCase):
+
+    def get_app(self):
+        self.sender = Mock()
+        return Application([
+            (r"/.*", FakeRequestHandler, {})
+        ])
+
+    @patch('time.sleep')
```
ghost: I don't think this mock is required.

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,159 @@
...
+        log = {'handler': 'FakeRequestHandler', 'method': 'post', 'requestBody': b"{'test': 'tested'}", 'duration': 5}
+        log_mock.assert_called_with(log)
+
+    @patch('logging.info')
+    def test_get_asynchronous_message(self, log_mock):
+        """
+        Had to enable a full wait here as the behaviour was different when the time
+        response was mocked
+        """
+        expected_response = "Hello world!"
```
ghost: Again, these three lines don't appear relevant to the rest of the tests.

common/utilities/tests/test_timing.py
ghost: What was the varying behaviour?

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,159 @@
...
+from unittest.mock import patch, Mock
+
+from tornado.testing import AsyncHTTPTestCase
+from tornado.web import Application, RequestHandler
+
+from common.utilities import timing
+from common.utilities.test_utilities import async_test
+
+
+class TestTimeUtilities(TestCase):
```
ghost: In terms of coverage, you've got tests for lots of different scenarios for `time_method`, but none for the same scenarios on the other methods, and they all implement the logic to handle these scenarios separately. If these methods are refactored to re-use some more code then this should take care of most of this coverage, but worth bearing in mind if certain bits of logic remain within the individual methods.

ghost (changes requested):

common/utilities/tests/test_timing.py
Philip-Woods: It's a bit difficult to describe, but it's what lead me to realise we needed async versions as well, because if I patch the time module and call an `async` method without the `async` decorator it would look like the decorator is working correctly. However without the patch it would log that the async method would take 0s (makes sense since it's a coroutine and will return instantly). So it feels like the non patched test should be included to cover this case

common/utilities/timing.py
Philip-Woods: Updated to use perf_counter

common/utilities/timing.py
Philip-Woods: 👍

Philip-Woods: As discussed these were included as something similar to what spine has. I think if tornado is logging things we probably want to ensure it also complies with the format we're wanting to log with?

ghost: Yeah, I think as we discussed offline it is worth having these methods to allow us to log details of HTTP requests in the same format as the rest of our messages.

common/utilities/tests/test_timing.py
ghost: Does that not point at the fact that your tests aren't checking that the values returned from the decorated method are as expected? That's an important consideration for the decorator, and doing the check would have highlighted the problem you describe, wouldn't it?

Philip-Woods: Good point, and I will update them to ensure they're returning the expected value as well. Do you think this should negate the need to run with an actual sleep?

common/utilities/timing.py
Philip-Woods: I did wonder about this as spine does something similar, however as a part of the logging module particular patient data is hidden but it seems quite difficult, should we just avoid logging anything that might contain any data just as a catch all?

common/utilities/tests/test_timing.py
ghost: If think it would, yes.

common/utilities/timing.py
ismail-s: Minor/optional: instead of saying that you're wrapping methods, it would be more general to say you're wrapping functions.

common/utilities/timing.py
```diff
@@ -0,0 +1,127 @@
...
+            duration = round(duration, 3)
+            logging.info({
+                'method': method.__name__,
+                'duration': duration
+            })
+
+    return invoke_method_with_timer
+
+
+def time_method_async(method):
```
ismail-s: I don't think separate functions for async/non-async are required, you can just use `inspect.iscoroutinefunction` to determine whether the function being decorated is async or not, and wrap it accordingly (yes, this does involve some reflection, but this is only at module import time).

common/utilities/timing.py
```diff
@@ -0,0 +1,127 @@
+import logging
+import time
+from functools import wraps
+
+
+class StopWatch:
```
ismail-s: Minor/optional: Users of this class have to know/remember to call the start method before the stop one. A possibly simpler alternative is to start the stopwatch in the constructor and have a method to get the current time on the stopwatch (I'm basing this on Google Guava's [Stopwatch](https://github.com/google/guava/blob/8000dc992d0124372972729b20df40a42b27a849/guava/src/com/google/common/base/Stopwatch.java) class).

Also, pedantic, but I think only the first letter needs to be capitalised.

common/utilities/timing.py
```diff
@@ -0,0 +1,127 @@
...
+    @wraps(method)
+    def invoke_method_with_timer(*args, **kwargs):
+        stopwatch = StopWatch()
+        stopwatch.start_timer()
+        try:
+            return method(*args, **kwargs)
+        finally:
+            duration = stopwatch.stop_timer()
+            duration = round(duration, 3)
+            logging.info({
```
ghost: Just looking at this again, this will produce output like:

```
2019-07-16 10:38:40,836 - INFO: {'method': 'foo', 'duration': 12.345}
```

This doesn't tally with our [logging guidance](https://gpitbjss.atlassian.net/wiki/spaces/RTDel/pages/1560608910/KDD-001+-+Logging+Strategy+and+Format) around log line format (in the Log Line Format section):

> Log lines should be human readable strings, not blobs of JSON or XML

This should log something more like `Method name=foo took duration=12.345 seconds`.

common/utilities/timing.py
```diff
@@ -0,0 +1,127 @@
...
+    async def invoke_method_with_timer(*args, **kwargs):
+        stopwatch = StopWatch()
+        stopwatch.start_timer()
+        try:
+            return await method(*args, **kwargs)
+        finally:
+            duration = stopwatch.stop_timer()
+            duration = round(duration, 3)
+            logging.info({
+                'method': method.__name__,
```
ismail-s: When logging, don't you have to have a format string as the first argument, and then params? Or is this a different way of logging?

common/utilities/timing.py
Philip-Woods: I've pulled the duplicate pieces out into their own methods, maybe went a bit overkill but the individual timing methods are fairly concise now.

ghost (changes requested):

common/utilities/tests/test_timing.py
Philip-Woods: Binned it since it wasn't really been used here, also I've added actual return values to the handler class so now the response is checked fully as well

common/utilities/tests/test_timing.py
Philip-Woods: Deleted

common/utilities/timing.py
Philip-Woods: This is correct, but that is something that is expected to be handled as apart of the logging module that is to be written next. I think when we started this ticket it was under the expectation that the logging would change/be updated following the proper logging implementation

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,159 @@
...
+    def test_invoke_with_time_rounding(self, log_mock, time_mock):
+        time_mock.return_value = 5.1236
+        self.sleep_method()
+        log_mock.assert_called_with({'method': 'sleep_method', 'duration': 5.124})
+
+    @patch('common.utilities.timing.StopWatch.stop_timer')
+    @patch('logging.info')
+    def test_invoke_with_time_parameters(self, log_mock, time_mock):
+        time_mock.return_value = 5
+        self.take_parameters("whew", 1, [2], {3: 3})
```
ismail-s: Minor/optional: could also test that varargs/keyword args get passed through.

common/utilities/timing.py
ghost: I think logging the full request would also violate the "Do not log extremely verbose data" guidance in the "Other Logging Good Practice" section of our [logging strategy](https://gpitbjss.atlassian.net/wiki/spaces/RTDel/pages/1560608910/KDD-001+-+Logging+Strategy+and+Format), so I'm inclined to say we should avoid it. Having this method that extracts key data would seem to meet the spirit of that section. Probably one to confirm with Dave.

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,159 @@
...
+
+class FakeRequestHandler(RequestHandler):
+
+    @timing.log_request
+    def post(self):
+        pass
+
+    @timing.async_log_request
+    async def get(self):
+        time.sleep(0.5)
```
ismail-s: Would be better to use an async sleep instead.

common/utilities/timing.py
ghost: I'd suggest that it's probably better to write the log message in the expected output format now, rather than having to come back and re-write it when the logging module is plugged in.

ismail-s (changes requested):

common/utilities/timing.py
Philip-Woods: I wrote this originally as a format inspired by what spine does where they just pass in a dictionary which is formatted by the logging module. As Gareth mentioned though we should be trying to use human readable strings so this has now changed anyway

common/utilities/tests/test_timing.py
Philip-Woods: This is removed now anyway

common/utilities/timing.py
Philip-Woods: Added this, definitely makes the usage of the decorators neater 👍

common/utilities/timing.py
Philip-Woods: valid point, I think I'll leave it as I prefer to lean to more explicit usage and to me calling `stop` without ever calling `start` feels kinda strange. I've updated the name though

common/utilities/tests/test_timing.py
Philip-Woods: So I've reorganised a bit now, most tests now include 2 sub tests, one sync and one async version, this should improve the general coverage and make the tests a bit cleaner as well. 

This is in addition to a bit of refactoring to pull out duplicate code

common/utilities/timing.py
```diff
@@ -0,0 +1,92 @@
...
+    duration = round(duration, 3)
+    logging.info(f'Method name={name} took duration={duration}')
+
+
+def _log_tornado_time(duration, handler, method):
+    duration = round(duration, 3)
+    logging.info(f'method={method} from handler={handler} took duration={duration}')
+
+
+def time_function(method):
```
ismail-s:
```suggestion
def time_function(func):
```

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,207 @@
...
+            self.assertEqual("whew12345", res)
+        with self.subTest("Async"):
+            time_mock.return_value = 5
+            res = await self.var_parameters_async("whew", 1, "three", 4)
+            log_mock.assert_called_with('Method name=var_parameters_async took duration=5')
+            self.assertEqual("whew1three4", res)
+
+    @timing.time_function
+    def default_method(self):
+        return "slept"
```
ismail-s: Why not also have this as:
```suggestion
        return 5
```

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,207 @@
...
+        time_mock.return_value = 5
+        with self.subTest("Sync version"):
+            res = self.default_method()
+            log_mock.assert_called_with(5, 'default_method')
+            self.assertEqual("slept", res)
+
+        with self.subTest("Async version"):
+            res = await self.default_method_async()
+            log_mock.assert_called_with(5, 'default_method_async')
+            self.assertEqual(5, res)
```
ismail-s: A bit confusing here to have 5 as both the return value and the time taken

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,207 @@
...
+    def take_parameters(self, check, one, two, three):
+        assert check is not None
+        assert one is not None
+        assert two is not None
+        assert three is not None
+        return check + str(one)
+
+    @timing.time_function
+    async def take_parameters_async(self, check, one, two, three):
+        assert check is not None
```
ismail-s: Can simplify this method down to `return self.take_parameters(check, one, two, three)` (and same for the other async-y methods, at least the more complicated ones)

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,207 @@
...
+    @patch('common.utilities.timing.Stopwatch.stop_timer')
+    @patch('common.utilities.timing._log_time')
+    @async_test
+    async def test_exception_thrown_whilst_timing(self, log_mock, time_mock):
+        time_mock.return_value = 5
+        with self.subTest("Sync"):
+            try:
+                self.throw_error_method()
+            except ValueError:
+                pass
```
ismail-s: It might make more sense to do `with self.assertRaises(ValueError): self.throw_error_method()`, as then the test fails if the exception wasn't thrown as expected.

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,207 @@
...
+    @timing.time_request
+    async def get(self):
+        self.write("hello")
+
+    @timing.time_request
+    def put(self):
+        try:
+            raise ValueError("Whew")
+        finally:
+            self.write("put")
```
ismail-s, line 162: What's the point of this line?

ismail-s (dismissed):

common/utilities/tests/test_timing.py
```diff
@@ -0,0 +1,207 @@
...
+        stopwatch.start_timer()
+        time_mock.return_value = 5.0
+        result = stopwatch.stop_timer()
+
+        self.assertEqual(result, 5)
+
+    @patch('logging.info')
+    def test_invoke_with_time_rounding(self, log_mock):
+        timing._log_time(5.1236, "yes")
+        log_mock.assert_called_with('Method name=yes took duration=5.124')
```
ghost: Probably worth a sub-test to verify `_log_tornado_time` also rounds as expected.

ghost (dismissed):

common/utilities/tests/test_timing.py
Philip-Woods: Made them both into strings and made it a constant since it was used in a few places, this should also address your later comment stating it's a bit unclear having the response value matching the value I set the time as too

common/utilities/tests/test_timing.py
Philip-Woods: I was trying to make it so all the methods I was calling return an actual value so that we can verify end result of the wrapped method is completed/returned, it was mentioned in a previous comment from Gareth that it would indicate that the async methods were correctly being called

ghost (approved):

ismail-s (approved):

PR #33, opened 2019-07-17
RT-91 Add workaround for certs with old signature algs

ismail-s:
This PR is meant to get the Jenkins builds working again. It seems to deal with the issue we've been having with SSL connection issues. However, there's now an issue where the asynchronous reply from Spine doesn't come.

I've tried debugging this, but haven't made much headway. I've managed to get it working once after rebooting the AWS EC2 instance, but then as soon as I tried re-running the integration tests (without restarting anything, MHS or EC2), then the integration tests started failing again. I've checked and updated the ASID environment variable, that hasn't fixed this. I've tried running the integration tests locally, they work fine. Any other ideas as to what this could be being caused by are welcome.

### DIFF
--- a/pipeline/packer/mhs.json
+++ b/pipeline/packer/mhs.json
@@ -40,6 +40,12 @@
     {
       "type": "shell",
       "inline": [
+        "sed -i 's/SECLEVEL=2/SECLEVEL=1/' /etc/ssl/openssl.cnf # Temporarily lower security to workaround opentest certs with SHA1 signatures"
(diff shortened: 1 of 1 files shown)


ismail-s:
After a slew of investigation, I figured out that switching the certs used in the test environment with my certs got the build to work. So we'll go with that for the timebeing, and try to figure out separately if we can get the original certs fixed somehow.

Philip-Woods (approved):

PR #34, opened 2019-07-17
Fix/common tests

Philip-Woods:
The pipeline doesn't currently run the unit tests in the `common` folder, this PR adds an extra step to do this. This also includes a few test changes in order to ensure the module paths are correct. The unit tests fully run on the pipeline

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,19 @@ pipeline {
     }
 
     stages {
+        stage('Common Module Unit Tests') {
+            steps {
+                dir('common') {
+                    sh label: 'Installing dependencies', script: 'pipenv install --dev --deploy --ignore-pipfile'
+                    sh label: 'Running unit tests', script: 'pipenv run unittests-cov'
+                    sh label: 'Displaying code coverage report', script: 'pipenv run coverage-report'
+                    sh label: 'Exporting code coverage report', script: 'pipenv run coverage-report-xml'
+                    cobertura coberturaReportFile: '**/coverage.xml'
+               }
+            }
+        }
+
+
         stage('Unit Tests') {
             steps {
                 dir('mhs-reference-implementation') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,22 +11,15 @@ pipeline {
         stage('Common Module Unit Tests') {
             steps {
                 dir('common') {
-                    sh label: 'Installing dependencies', script: 'pipenv install --dev --deploy --ignore-pipfile'
-                    sh label: 'Running unit tests', script: 'pipenv run unittests-cov'
-                    sh label: 'Displaying code coverage report', script: 'pipenv run coverage-report'
-                    sh label: 'Exporting code coverage report', script: 'pipenv run coverage-report-xml'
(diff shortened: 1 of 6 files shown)


Jenkinsfile
```diff
@@ -8,6 +8,19 @@ pipeline {
...
+                    sh label: 'Running unit tests', script: 'pipenv run unittests-cov'
+                    sh label: 'Displaying code coverage report', script: 'pipenv run coverage-report'
+                    sh label: 'Exporting code coverage report', script: 'pipenv run coverage-report-xml'
+                    cobertura coberturaReportFile: '**/coverage.xml'
+               }
+            }
+        }
+
+
         stage('Unit Tests') {
```
ismail-s: Minor/optional: could we do:
```suggestion
         stage('MHS Unit Tests') {
```

common/Pipfile
```diff
@@ -13,3 +15,9 @@ python_version = "3.7"
 
 [scripts]
 tests = "python -m unittest -v"
```
ghost: Probably worth removing this `tests` script now, as it duplicates `unittests`.

Jenkinsfile
```diff
@@ -8,6 +8,19 @@ pipeline {
...
 
     stages {
+        stage('Common Module Unit Tests') {
+            steps {
+                dir('common') {
+                    sh label: 'Installing dependencies', script: 'pipenv install --dev --deploy --ignore-pipfile'
+                    sh label: 'Running unit tests', script: 'pipenv run unittests-cov'
+                    sh label: 'Displaying code coverage report', script: 'pipenv run coverage-report'
+                    sh label: 'Exporting code coverage report', script: 'pipenv run coverage-report-xml'
+                    cobertura coberturaReportFile: '**/coverage.xml'
```
ismail-s: Optional: seeing as this is the same as the MHS unit tests block, you possibly could [extract as a function](https://stackoverflow.com/a/47631522), though I'm not sure how well supported that functionality is.

ghost: Looking at the Jenkins build for this branch/PR, it appears that it's picking up the same coverage report (for the MHS) twice, rather than reporting coverage for the common module.

ghost (changes requested):

ismail-s (commented):
I was thinking about doing this, good job I didn't start doing so.

You're missing a `.coveragerc` file (see the one in the mhs folder), atm the coverage report is including tornado, pystache and your unittests.

And for some reason, in [Jenkins](https://nhs-red-team-803592372.eu-west-2.elb.amazonaws.com/job/Test2/job/PR-34/1/), the 2 coverage reports are the same. I think potentially just run the cobertura line just once in the base folder, it should pick up both.

Jenkinsfile
Philip-Woods: I've added this in since we'll probably have similar commands for the SCR adapter as well it makes sense to have it

Philip-Woods: Updated the Jenkins file so that it now has a single coverage stage which combines both reports

ismail-s (approved):

ghost (approved):

PR #35, opened 2019-07-18
Feature/RT75-Logger

Philip-Woods:
This PR adds a logging module to wrap the python logging module in particular giving us the following features:

* Fixed message format (messages are generated with a key=value pairs) - this also includes enforcing the rules on key=value pairs (No spaces in key, quotes around values if it contains spaces)
* The overall log format (Time at the start, processId, log level at the end ect)
* Prevents log levels outside of our designated log levels 
* Adds an `audit` level log

An example of an audit level log produced by the logging module:

`[2019-07-19T11:12:35.998Z] ThereWillBeNoSpacesToday="wow qwe" CorrelationId=2 pid=27296 LogLevel=AUDIT ProcessKey=TES000`

### DIFF
--- a/common/utilities/integration_adapters_logger.py
+++ b/common/utilities/integration_adapters_logger.py
@@ -2,24 +2,33 @@ import logging
 import sys
 import time
 
-# Set a custom log level
 AUDIT = 25
-logging.addLevelName(AUDIT, "AUDIT")
 
 
-def audit(self, message, *args, **kws):
-    if self.isEnabledFor(AUDIT):
-        self._log(AUDIT, message, args, **kws)
+def _add_audit_level_log():
+    """
+    Adds a new log level to the standard python logging module, this level is the `audit` level and is defined as
+    a high level than INFO but lower than  WARNING
+    """
+    logging.addLevelName(AUDIT, "AUDIT")
 
+    def audit(self, message, *args, **kws):
+        if self.isEnabledFor(AUDIT):
+            self._log(AUDIT, message, args, **kws)
 
-logging.Logger.audit = audit
+    logging.Logger.audit = audit
 
 
 def load_global_log_config():
+    """
+    A general method to load the overall config of the system, specifically it modifies the root handler to output
+    to sysout and sets the default log levels and format. This is expected to be called once at the start of a
+    system.
+    """
+    _add_audit_level_log()
     logger = logging.getLogger()
     logger.setLevel(logging.NOTSET)
     handler = logging.StreamHandler(sys.stdout)
-    handler.setLevel(logging.NOTSET)
     logging.Formatter.converter = time.gmtime
     formatter = logging.Formatter('[%(asctime)s.%(msecs)03d] '
                                   '%(message)s pid=%(process)d LogLevel=%(levelname)s ProcessKey=%(processKey)s',
@@ -29,6 +38,10 @@ def load_global_log_config():
 
 
 def _format_values_in_map(dict_values: dict) -> dict:
+    """
+    Replaces the values in the map with key=value so that the key in a string can be replaced with the correct
+    log format, also surrounds the value with quotes if it contains spaces
+    """
     new_map = {}
     for key, value in dict_values.items():
         if ' ' in value:
@@ -39,11 +52,14 @@ def _format_values_in_map(dict_values: dict) -> dict:
 
 
 def _formatted_string(message: str, dict_values: dict) -> str:
+    """
+    Populates the string with the correctly formatted dictionary values
+    """
     formatted_values = _format_values_in_map(dict_values)
     return message.format(**formatted_values)
 
 
-class Logger:
+class IntegrationAdaptersLogger:
 
     def __init__(self, log_ref: str = "SYS"):
         self.process_key_tag = log_ref
@@ -55,6 +71,9 @@ class Logger:
         self.logger.log(level, message, extra={'processKey': self.process_key_tag + process_key_num})
 
     def _format_and_write(self, message, values, process_key_num, request_id, correlation_id, level):
+        """
+        Formats the string and appends the appropriate values if they are included before writing the log
+        """
         message = _formatted_string(message, values)
 
         if request_id:
@@ -66,30 +85,25 @@ class Logger:
 
     def info(self, message: str, values: dict = None, process_key_num: str = None,
              request_id: str = None, correlation_id=None):
-        if values is None:
-            values = {}
+        values = values if values is not None else {}
         self._format_and_write(message, values, process_key_num, request_id, correlation_id, logging.INFO)
 
     def audit(self, message: str, values: dict = None, process_key_num: str = None,
               request_id: str = None, correlation_id=None):
-        if values is None:
-            values = {}
+        values = values if values is not None else {}
         self._format_and_write(message, values, process_key_num, request_id, correlation_id, AUDIT)
 
     def warning(self, message: str, values: dict = None, process_key_num: str = None,
                 request_id: str = None, correlation_id=None):
-        if values is None:
-            values = {}
+        values = values if values is not None else {}
         self._format_and_write(message, values, process_key_num, request_id, correlation_id, logging.WARNING)
 
     def error(self, message: str, values: dict = None, process_key_num: str = None,
               request_id: str = None, correlation_id=None):
-        if values is None:
-            values = {}
+        values = values if values is not None else {}
         self._format_and_write(message, values, process_key_num, request_id, correlation_id, logging.ERROR)
 
     def critical(self, message: str, values: dict = None, process_key_num: str = None,
                  request_id: str = None, correlation_id=None):
-        if values is None:
-            values = {}
+        values = values if values is not None else {}
         self._format_and_write(message, values, process_key_num, request_id, correlation_id, logging.CRITICAL)
--- a/common/utilities/tests/test_logger.py
+++ b/common/utilities/tests/test_logger.py
@@ -8,6 +8,9 @@ from utilities import integration_adaptors_logger as log
 
 class TestLogger(TestCase):
 
+    def tearDown(self) -> None:
+        logging.getLogger().handlers = []
+
     def test_dictionary_formatting(self):
         # Tests both removing the spaces and surrounding values with quotes if needed
         input_dict = {
--- a/common/utilities/tests/test_logger.py
+++ b/common/utilities/tests/test_logger.py
@@ -1,0 +1,10 @@
+from unittest import TestCase
+
+from common.utilities import logger
+
+
+class TestLogger(TestCase):
+
+    def test_thing(self):
+        logger.Logger("", log_level=0)\
+            .audit('{There Will Be No Spaces Today}', {'There Will Be No Spaces Today': 'wow qwe'}, correlation_id=2)
--- a/common/utilities/tests/test_logger.py
+++ b/common/utilities/tests/test_logger.py
@@ -19,7 +19,7 @@ class TestLogger(TestCase):
             'Key With Space': 'KeyWithSpace="value with space"',
             'EasyKey': 'EasyKey=EasyValue'
         }
-        output = log.IntegrationAdaptorsLogger()._format_values_in_map(input_dict)
+        output = log.IntegrationAdaptorsLogger('SYS')._format_values_in_map(input_dict)
         self.assertEqual(output, expected_output)
 
     @patch('sys.stdout', new_callable=io.StringIO)
@@ -38,7 +38,7 @@ class TestLogger(TestCase):
     @patch('sys.stdout', new_callable=io.StringIO)
     def test_format_and_write(self, mock_std):
         log.configure_logging()
-        log.IntegrationAdaptorsLogger("SYS")._format_and_write(
+        log.IntegrationAdaptorsLogger('SYS')._format_and_write(
             message='{yes} {no} {maybe}',
             values={'yes': 'one', 'no': 'two', 'maybe': 'three'},
             request_id=10,
@@ -61,7 +61,7 @@ class TestLogger(TestCase):
     @patch('sys.stdout', new_callable=io.StringIO)
     def test_format_and_write_empty_vals(self, mock_std):
         log.configure_logging()
-        log.IntegrationAdaptorsLogger("SYS")._format_and_write(
+        log.IntegrationAdaptorsLogger('SYS')._format_and_write(
             message='{yes} {no} {maybe}',
             values={'yes': 'one', 'no': 'two', 'maybe': 'three'},
             request_id=None,
@@ -80,7 +80,7 @@ class TestLogger(TestCase):
         self.assertNotIn('RequestId=', output)
 
     def test_log_levels(self):
-        logger = log.IntegrationAdaptorsLogger()
+        logger = log.IntegrationAdaptorsLogger('SYS')
         logger._format_and_write = MagicMock()
         with self.subTest('INFO'):
             logger.info('100', '{yes}', {'yes': 'no'}, 'REQ', 313)
@@ -105,7 +105,7 @@ class TestLogger(TestCase):
     @patch('sys.stdout', new_callable=io.StringIO)
     def test_empty_values(self, mock_stdout):
         log.configure_logging()
-        logger = log.IntegrationAdaptorsLogger()
+        logger = log.IntegrationAdaptorsLogger('SYS')
         with self.subTest('Empty info log'):
             logger.info('100', 'I can still log info strings without values!')
             output = mock_stdout.getvalue()
@@ -129,7 +129,7 @@ class TestLogger(TestCase):
 
     def test_write_throws_error_on_bad_params(self):
         with self.assertRaises(ValueError):
-            log.IntegrationAdaptorsLogger()._write(logging.INFO, 'no processkey', None)
+            log.IntegrationAdaptorsLogger('SYS')._write(logging.INFO, 'no processkey', None)
 
     def test_undefined_log_ref_throws_error(self):
         with self.assertRaises(ValueError):
@@ -137,4 +137,4 @@ class TestLogger(TestCase):
 
     def test_2(self):
         log.configure_logging()
-        log.IntegrationAdaptorsLogger('YES').info("100", "qweqwe", {})
+        log.IntegrationAdaptorsLogger('YES').info('100', 'qweqwe', {})
--- a/common/utilities/tests/test_logger.py
+++ b/common/utilities/tests/test_logger.py
@@ -3,7 +3,7 @@ import os
 import logging
 from unittest import TestCase
 from unittest.mock import patch, MagicMock
-from common.utilities import integration_adaptors_logger as log
+from utilities import integration_adaptors_logger as log
 
 
 class TestLogger(TestCase):
--- a/common/utilities/tests/test_logger.py
+++ b/common/utilities/tests/test_logger.py
@@ -2,12 +2,7 @@ import io
 import logging
 from unittest import TestCase
 from unittest.mock import patch, MagicMock
-import re
-
 from common.utilities import logger as log
-
-# This is the regex use to extract the key=value formatted log strings into dictionaries for testing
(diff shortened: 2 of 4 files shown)


common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+    Adds a new log level to the standard python logging module, this level is the `audit` level and is defined as
+    a high level than INFO but lower than  WARNING
+    """
+    logging.addLevelName(AUDIT, "AUDIT")
+
+    def audit(self, message, *args, **kws):
+        if self.isEnabledFor(AUDIT):
+            self._log(AUDIT, message, args, **kws)
+
+    logging.Logger.audit = audit
```
ismail-s: Possibly better to use [`getLoggerClass`](https://docs.python.org/3.7/library/logging.html#logging.getLoggerClass):
```suggestion
    logging.getLoggerClass().audit = audit
```

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+    to sysout and sets the default log levels and format. This is expected to be called once at the start of a
+    system.
+    """
+    _add_audit_level_log()
+    logger = logging.getLogger()
+    logger.setLevel(logging.NOTSET)
+    handler = logging.StreamHandler(sys.stdout)
+    logging.Formatter.converter = time.gmtime
+    formatter = logging.Formatter('[%(asctime)s.%(msecs)03d] '
+                                  '%(message)s pid=%(process)d LogLevel=%(levelname)s ProcessKey=%(processKey)s',
```
ismail-s: Is ProcessKey meant to be a unique id for each log message? If so, then I think it would be better called something like LogKey or LogId

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+
+def _format_values_in_map(dict_values: dict) -> dict:
+    """
+    Replaces the values in the map with key=value so that the key in a string can be replaced with the correct
+    log format, also surrounds the value with quotes if it contains spaces
+    """
+    new_map = {}
+    for key, value in dict_values.items():
+        if ' ' in value:
+            value = f"\"{value}\""
```
ismail-s: Trivial/optional: I think it would look cleaner to use `'` instead of `"`. From the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#310-strings):

> Be consistent with your choice of string quote character within a file. Pick ' or " and stick with it. It is okay to use the other quote character on a string to avoid the need to \\ escape within the string.

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+
+def _formatted_string(message: str, dict_values: dict) -> str:
+    """
+    Populates the string with the correctly formatted dictionary values
+    """
+    formatted_values = _format_values_in_map(dict_values)
+    return message.format(**formatted_values)
+
+
+class IntegrationAdaptersLogger:
```
ismail-s:
```suggestion
class IntegrationAdaptorsLogger:
```
I know this is being pedantic, but I feel like might as well be consistent in our repo (and also to avoid any confusion). And yes, I did Google [the difference between the 2 words](https://grammarist.com/spelling/adapter-adaptor/): apparently both of them are fine.

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def __init__(self, log_ref: str = "SYS"):
```
ismail-s: Why have the default value?

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+        self.process_key_tag = log_ref
+        self.logger = logging.getLogger()
```
ismail-s: So if I understand correctly, everything would be using the same root logger. And then we distinguish by `log_ref + process_key_num`.

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _write(self, level, message, process_key_num: str):
+        if not process_key_num:
+            process_key_num = "000"
```
ismail-s: Should we instead just make `process_key_num` be mandatory ie what's the benefit to providing this default value?

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+import sys
+import time
+
+AUDIT = 25
+
+
+def _add_audit_level_log():
+    """
+    Adds a new log level to the standard python logging module, this level is the `audit` level and is defined as
+    a high level than INFO but lower than  WARNING
```
ghost: Minor typo:
```suggestion
    a higher level than INFO but lower than WARNING
```

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    logging.addLevelName(AUDIT, "AUDIT")
+
+    def audit(self, message, *args, **kws):
```
ghost: Do you actually use this audit method? It looks like all logging is done via the `log(level, ...)` method.

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+            self._log(AUDIT, message, args, **kws)
+
+    logging.Logger.audit = audit
+
+
+def load_global_log_config():
+    """
+    A general method to load the overall config of the system, specifically it modifies the root handler to output
+    to sysout and sets the default log levels and format. This is expected to be called once at the start of a
+    system.
```
ghost: Perhaps slightly clearer to say:
```suggestion
    application.
```

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+        if self.isEnabledFor(AUDIT):
+            self._log(AUDIT, message, args, **kws)
+
+    logging.Logger.audit = audit
+
+
+def load_global_log_config():
+    """
+    A general method to load the overall config of the system, specifically it modifies the root handler to output
+    to sysout and sets the default log levels and format. This is expected to be called once at the start of a
```
ghost:
```suggestion
    to stdout and sets the default log levels and format. This is expected to be called once at the start of a
```

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+    system.
+    """
+    _add_audit_level_log()
+    logger = logging.getLogger()
+    logger.setLevel(logging.NOTSET)
+    handler = logging.StreamHandler(sys.stdout)
+    logging.Formatter.converter = time.gmtime
+    formatter = logging.Formatter('[%(asctime)s.%(msecs)03d] '
+                                  '%(message)s pid=%(process)d LogLevel=%(levelname)s ProcessKey=%(processKey)s',
+                                  '%Y-%m-%d %H:%M:%S')
```
ghost: I think it's worth using the `T` separator between date and time, as specified in the [logging principles document](https://gpitbjss.atlassian.net/wiki/spaces/RTDel/pages/1560608910/KDD-001+-+Logging+Strategy+and+Format#KDD-001-LoggingStrategyandFormat-Timestamp), [since it appears that omitting it needs to be something both parties agree on](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations):

```suggestion
                                  '%Y-%m-%dT%H:%M:%S')
```

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+    A general method to load the overall config of the system, specifically it modifies the root handler to output
+    to sysout and sets the default log levels and format. This is expected to be called once at the start of a
+    system.
+    """
+    _add_audit_level_log()
+    logger = logging.getLogger()
+    logger.setLevel(logging.NOTSET)
+    handler = logging.StreamHandler(sys.stdout)
+    logging.Formatter.converter = time.gmtime
+    formatter = logging.Formatter('[%(asctime)s.%(msecs)03d] '
```
ghost: I think we should include the `Z` suffix to indicate this timestamp is in UTC, as [otherwise the standard says the time is assumed to be local time](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators).

```suggestion
    formatter = logging.Formatter('[%(asctime)s.%(msecs)03dZ] '
```

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+    logging.addLevelName(AUDIT, "AUDIT")
+
+    def audit(self, message, *args, **kws):
+        if self.isEnabledFor(AUDIT):
+            self._log(AUDIT, message, args, **kws)
+
+    logging.Logger.audit = audit
+
+
+def load_global_log_config():
```
ghost: Maybe:
```suggestion
def configure_logging():
```

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+                                  '%(message)s pid=%(process)d LogLevel=%(levelname)s ProcessKey=%(processKey)s',
+                                  '%Y-%m-%d %H:%M:%S')
+    handler.setFormatter(formatter)
+    logger.addHandler(handler)
+
+
+def _format_values_in_map(dict_values: dict) -> dict:
+    """
+    Replaces the values in the map with key=value so that the key in a string can be replaced with the correct
+    log format, also surrounds the value with quotes if it contains spaces
```
ghost: And removes spaces from the key name.

common/utilities/integration_adapters_logger.py
ghost: Should this always use the root logger? I think it would make sense to take the logger's name as a constructor parameter, so the modules using this utility can pass `__name__`.

common/utilities/integration_adapters_logger.py
ghost: Could use a docstring to explain what `log_ref` is.

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class IntegrationAdaptersLogger:
+
+    def __init__(self, log_ref: str = "SYS"):
+        self.process_key_tag = log_ref
+        self.logger = logging.getLogger()
+
+    def _write(self, level, message, process_key_num: str):
+        if not process_key_num:
```
ghost: I'm not sure there's much value in defaulting to using `000` here. Perhaps just raise an error indicating the argument was invalid?

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+        if request_id:
+            message += f' RequestId={request_id}'
+        if correlation_id:
+            message += f' CorrelationId={correlation_id}'
+
+        self._write(level, message, process_key_num)
+
+    def info(self, message: str, values: dict = None, process_key_num: str = None,
+             request_id: str = None, correlation_id=None):
+        values = values if values is not None else {}
```
ghost: You could extract the implementation of all of these log methods to a single method that just takes the log level.

ghost (changes requested):
Can you include an example of the log strings that are output when using this utility please? It would make it easier to compare to our desired logging standards, etc.

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+    handler = logging.StreamHandler(sys.stdout)
+    logging.Formatter.converter = time.gmtime
+    formatter = logging.Formatter('[%(asctime)s.%(msecs)03d] '
+                                  '%(message)s pid=%(process)d LogLevel=%(levelname)s ProcessKey=%(processKey)s',
+                                  '%Y-%m-%d %H:%M:%S')
+    handler.setFormatter(formatter)
+    logger.addHandler(handler)
+
+
+def _format_values_in_map(dict_values: dict) -> dict:
```
ghost: This function (and `_formatted_string`) might make more sense as static methods on the `IntegrationAdaptersLogger` class.

common/utilities/integration_adapters_logger.py
```diff
@@ -0,0 +1,109 @@
...
+        message = _formatted_string(message, values)
+
+        if request_id:
+            message += f' RequestId={request_id}'
+        if correlation_id:
+            message += f' CorrelationId={correlation_id}'
+
+        self._write(level, message, process_key_num)
+
+    def info(self, message: str, values: dict = None, process_key_num: str = None,
```
ghost: I think `process_key_num` shouldn't default to `None`. We want to be specifying this each time we write a log statement, to ensure they're unique.

common/utilities/tests/test_logger.py
```diff
@@ -0,0 +1,124 @@
...
+
+        expected_output = {
+            'Key With Space': 'KeyWithSpace="value with space"',
+            'EasyKey': 'EasyKey=EasyValue'
+        }
+        output = log._format_values_in_map(input_dict)
+        self.assertEqual(output, expected_output)
+
+    @patch('sys.stdout', new_callable=io.StringIO)
+    def test_audit(self, mock_std):
```
ghost: I'm not sure this test name makes it clear what you're testing.

common/utilities/integration_adapters_logger.py
Philip-Woods: Good point, I thought it had to be there in order for the logger to print that log level but it seems you only need the `logging.addLevelName` which I'll just move to the global config method.

common/utilities/tests/test_logger.py
```diff
@@ -0,0 +1,124 @@
...
+            level=logging.INFO,
+            process_key_num="100"
+        )
+
+        output = mock_std.getvalue()
+        self.assertIn(' LogLevel=INFO ', output)
+        self.assertIn(' ProcessKey=SYS100', output)
+        self.assertIn(' maybe=three ', output)
+        self.assertIn(' pid=', output)
+        self.assertIn(' yes=one ', output)
```
ghost, line 81: Probably worth a couple of assertions to verify that the `RequestId` and `CorrelationId` keys are not present in the string.

ismail-s:
Your Jenkins build is failing with error message _ModuleNotFoundError: No module named 'common'_.

common/utilities/tests/test_logger.py
```diff
@@ -0,0 +1,124 @@
...
+        expected_output = {
+            'Key With Space': 'KeyWithSpace="value with space"',
+            'EasyKey': 'EasyKey=EasyValue'
+        }
+        output = log._format_values_in_map(input_dict)
+        self.assertEqual(output, expected_output)
+
+    @patch('sys.stdout', new_callable=io.StringIO)
+    def test_audit(self, mock_std):
+        log.load_global_log_config()
```
ismail-s: In a few of these tests, you call this function. And each time you do, you are adding another handler to the root logger. I mean, the tests are all passing (after fixing the import issue), but it doesn't feel clean/proper.

common/utilities/tests/test_logger.py
```diff
@@ -0,0 +1,124 @@
...
+        output = mock_std.getvalue()
+
+        # Check that each value has spaces
+        self.assertIn(' CorrelationId=5 ', output)
+        self.assertIn(' LogLevel=INFO ', output)
+        self.assertIn(' ProcessKey=SYS100', output)
+        self.assertIn(' RequestId=10 ', output)
+        self.assertIn(' maybe=three ', output)
+        self.assertIn(' yes=one ', output)
+        self.assertIn(' pid=', output)
```
ismail-s: You could just do
```suggestion
        self.assertIn(f' pid={os.getpid()} ', output)
```

common/utilities/tests/test_logger.py
```diff
@@ -0,0 +1,124 @@
...
+            correlation_id=None,
+            level=logging.INFO,
+            process_key_num="100"
+        )
+
+        output = mock_std.getvalue()
+        self.assertIn(' LogLevel=INFO ', output)
+        self.assertIn(' ProcessKey=SYS100', output)
+        self.assertIn(' maybe=three ', output)
+        self.assertIn(' pid=', output)
```
ismail-s: https://github.com/nhsconnect/integration-adaptors/pull/35#discussion_r305317957

common/utilities/tests/test_logger.py
```diff
@@ -0,0 +1,124 @@
...
+
+        with self.subTest("CRITICAL"):
+            logger.critical("{yes}", {'yes': 'no'}, "100", "REQ", 313)
+            logger._format_and_write.assert_called_with("{yes}", {'yes': 'no'}, "100", "REQ", 313, logging.CRITICAL)
+
+    @patch('sys.stdout', new_callable=io.StringIO)
+    def test_empty_values(self, mock_stdout):
+        log.load_global_log_config()
+        logger = log.IntegrationAdaptersLogger()
+        with self.subTest("Empty info log"):
```
ismail-s: Optional: for this and the previous test, you could do some sort of parametrising-y thing to refactor out the common parts in each test. Though this probably/possibly would make the test less readable, so not sure if it's worth it.

common/utilities/integration_adapters_logger.py
Philip-Woods: As discussed this has been updated so that the config applies to all logger and now each instance of `IntegrationAdaptorsLogger` uses `getLogger(log_ref)` as it's logger

common/utilities/integration_adapters_logger.py
Philip-Woods: Changed this in a few other places to make it a bit more consistent, I'll try to stick to ' from now on as well

common/utilities/integration_adapters_logger.py
Philip-Woods: This has now changed, we set the root logger to have the correct message format and each module will have it's own logger name based on the log ref, this seemed to be what the python logging docs suggested

common/utilities/tests/test_logger.py
Philip-Woods: Yeah would definitely save lines, though I'll keep it as it is if you think it might make it harder to read

common/utilities/tests/test_logger.py
Philip-Woods: Yeah I had an issue with this actually, I wanted to run it once before all but I ran into issues with the mocking because adding it before the test seems to not mock stdout as expected, running it at the start of each test was the only consistent solution. Other ideas would be good if you have any

common/utilities/integration_adapters_logger.py
Philip-Woods: Yep this is now mandatory and throws a value error without it

common/utilities/integration_adapters_logger.py
Philip-Woods: Removed it and added a `ValueError`

common/utilities/integration_adapters_logger.py
Philip-Woods: Yeah it's to identify where the log came from in the code, `ProcessKey` was chosen as to be consistent with the [logging docs](https://gpitbjss.atlassian.net/wiki/spaces/RTDel/pages/1560608910/KDD-001+-+Logging+Strategy+and+Format)

Edit: More specifically it's meant to be a unique id for each log message _type_, it's not a value unique to each log instance

common/utilities/integration_adapters_logger.py
Philip-Woods: This has been removed anyway

common/utilities/integration_adapters_logger.py
Philip-Woods: Removed the default value and moved the process key to be the first element inserted to abide by pythons rules about default values and parameter orders

common/utilities/integration_adapters_logger.py
ghost: As discussed offline, this won't really make things more readable, so I'm happy to leave this as it is.

ghost (dismissed):

ghost (dismissed):

ismail-s (approved):

ghost (approved):

PR #37, opened 2019-07-19
RT-28 - Application Refactor

ghost:
This PR is for the proposed package, class and module structure for the MHS, which aims to match the currently defined architecture (visible at the bottom of this [Confluence page](https://gpitbjss.atlassian.net/wiki/spaces/RTDel/pages/1558085689/MHS+architecture+workshop+05+July+2019)).

I've created the new application structure in the `mhs` directory (and removed the existing `mhs-reference-implementation` directory). At a high-level, I have divided the package structure into `inbound`, `outbound` and `message_configuration` to correlate with the separate application components that have been identified as part of the architecture, with a `common` package for the components that will be shared between high-level components. Within each of these top-level packages there are sub-packages that correlate to the sub-components (white boxes) within the architecture diagram. **One key caveat is that this application still runs as a single Tornado application, and the decomposition into separate applications that communicate via queue has not yet been performed.**

I've also updated the build scripts to ensure that the integration tests still run successfully against this new application structure, and manually verified (using the [`supplier-example` application](https://github.com/nhsconnect/integration-adaptors/tree/develop/supplier-example)) that I can still successfully make a GP summary upload request to Opentest's instance of Spine.


mhs/mhs/common/messages/envelope.py
```diff
@@ -0,0 +1,26 @@
...
+    def serialize(self):
+        """Produce a serialised representation of this message.
+
+        :return: The serialized representation of this message.
+        """
+        pass
+
+    @classmethod
+    @abc.abstractmethod
+    def from_string(cls, headers: typing.Dict[str, str], message: str):
```
ismail-s: Optional: If you want to do:
```suggestion
    def from_string(cls, headers: typing.Dict[str, str], message: str) -> Envelope:
```
then you could do `from __future__ import annotations`. Or you could do:
```suggestion
    def from_string(cls, headers: typing.Dict[str, str], message: str) -> 'Envelope':
```

And same for other classes in this subpackage. Or I could do this as part of RT-72.

mhs/mhs/common/state/persistance_adapter.py
```diff
@@ -0,0 +1,8 @@
+"""
+This module defines the state adapter interface, used to allow support for multiple state database implementations.
+"""
+
+
+class PersistanceAdapter:
```
ismail-s:
```suggestion
class PersistenceAdapter:
```
Would also require a renaming the file.

mhs/README.md
```diff
@@ -1,12 +1,12 @@
-# MHS Reference Implementation
+# MHS
 
-This package contains a reference implementation of an MHS.
+This package contains a pre-assured implementation of an MHS.
```
ismail-s: Not quite belonging to this PR, but what does MHS stand for? Can we mention that here, or somewhere in this repo?
This might be better done as part of RT-128.

mhs/README.md
```diff
@@ -1,12 +1,12 @@
...
 
-## Installation
```
ismail-s, line 5: Probs worth mentioning that `pipenv install` is enough to run MHS.

mhs/mhs/message_configuration/lookup/cache_adapter.py
```diff
@@ -14,11 +18,10 @@ def __init__(self, expiry_time=FIFTEEN_MINUTES_IN_SECONDS):
             raise ValueError('Expiry time must not be non-negative')
         self.expiry_time = expiry_time
 
-    @abc.abstractmethod
```
ismail-s: Why did you remove this line?

ghost: Must have been an accidental deletion, thanks!

ismail-s (approved):

PR #38, opened 2019-07-22
RT-79 config component

ismail-s:
- Create a common config component and use it in mhs
- Configure the build pipeline to set the log level for the integration tests
- Change the logger to get the log level from config
  - I originally did this as 3ef57d4, then changed it to cb7719b. If people want, I can switch back to 3ef57d4, otherwise I'll leave it as-is.
- Add some docs for Terraform variables

### DIFF
--- a/common/utilities/config.py
+++ b/common/utilities/config.py
@@ -1,0 +1,24 @@
+"""
+Config
+
+This module holds the config used by an application. To use this module, first
+call `setup_config` to populate `config`. Then, just get any required config
+from `config`.
+"""
+import os
+from typing import Dict
+
+config: Dict[str, str] = {}
+
+
+def setup_config(component_name: str):
+    """
+    Populate the `config` variable in this module
+
+    :param component_name: name of the component, used to find the relevant
+    environment variables to populate `config` with.
+    """
+    prefix = component_name + "_"
+    for k, v in os.environ.items():
+        if k.startswith(prefix):
+            config[k[len(prefix):]] = v
--- a/common/utilities/config.py
+++ b/common/utilities/config.py
@@ -3,8 +3,9 @@ Config
 
 This module holds the config used by an application. To use this module, first
 call `setup_config` to populate `config`. Then, just get any required config
-from `config`.
+using `get_config` (or directly use `config` for optional config).
 """
+import logging
 import os
 from typing import Dict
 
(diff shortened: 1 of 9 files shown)


common/utilities/tests/test_config.py
```diff
@@ -0,0 +1,26 @@
+import unittest
+from unittest.mock import patch
+
+import utilities.config as config
+
+
+@patch("utilities.config.config", new_callable=dict)
```
ghost: No major issues with the approach here, but it looks like [`patch.dict`](https://docs.python.org/3/library/unittest.mock.html#patch-dict) would also be an option. It might make things a little more readable?

common/utilities/config.py
```diff
@@ -0,0 +1,24 @@
+"""
+Config
+
+This module holds the config used by an application. To use this module, first
+call `setup_config` to populate `config`. Then, just get any required config
+from `config`.
```
ghost: The design decision on configuration says:

> there is no need to write code to handle the cases where the config item hasn't been provided, simply erroring with a clear message detailing which required configuration item is unavailable is sufficient

Should we implement a `get_config` function that handles logging a standard message if the requested configuration value is not available, or was that deliberately ruled out?

common/utilities/tests/test_config.py
```diff
@@ -24,3 +27,29 @@ def test_setup_config_filters_by_prefix(self, mock_environ):
         config.setup_config("PREFIX")
 
         self.assertEqual({}, config.config)
+
+    def test_get_config_success(self, mock_environ):
+        mock_environ["PREFIX_TEST"] = "123"
+        mock_environ["PREFIX_LOG_LEVEL"] = "INFO"
```
ghost: This isn't used.

common/utilities/tests/test_config.py
```diff
@@ -0,0 +1,55 @@
...
+        config.setup_config("PREFIX")
+
+        self.assertEqual("123", config.get_config("TEST"))
+
+    @patch('sys.stdout', new_callable=io.StringIO)
+    def test_get_config_no_config_variable_found(self, mock_stdout, mock_environ):
+        mock_environ["PREFIX_LOG_LEVEL"] = "INFO"
+        config.setup_config("PREFIX")
+
+        def rename_logging_handler():
```
ghost:
```suggestion
        def remove_logging_handler():
```

common/utilities/tests/test_config.py
```diff
@@ -0,0 +1,55 @@
...
+            logging.getLogger().handlers = []
+
+        self.addCleanup(rename_logging_handler)
```
ghost:
```suggestion
        self.addCleanup(remove_logging_handler)
```

ghost (approved):

PR #42, opened 2019-07-26
RT-93 message packer generation

ismail-s:
- Throw error when missing tag when generating messages
- Improve logging around message generation
- Add more unittests around message generation
- Add a `messageId` optional query param that can be used to specify the message id so as to use the same message id in the message header and HL7 body.
  - Add an integration test to test this (in addition to unittest)
- Fix (in e3b8a80) a bug in configuration manager (which I hit with this PR) where the same config object was being returned instead of a copy (so any mutation to the config object would get shared).


common/builder/pystache_message_builder.py
```diff
@@ -20,4 +26,16 @@ def build_message(self, message_dictionary):
...
-        return self._renderer.render(self._parsed_template, message_dictionary)
+        try:
+            return self._renderer.render(self._parsed_template, message_dictionary)
+        except KeyNotFoundError as e:
+            logger.error('0001', 'Failed to find key when generating message from {TemplateFile} . {ErrorMessage}',
+                         {'TemplateFile': self.template_file, 'ErrorMessage': str(e)})
+            raise MessageGenerationSOAPFaultError()
+
+
+class MessageGenerationSOAPFaultError(Exception):
```
ismail-s: Thinking of changing this to something like `MessageGenerationError` or some more specific thing like `MessageGenerationMissingTagError`. The reason I bunged `SOAPFault` in the name originally was because it was mentioned in [Confluence](https://gpitbjss.atlassian.net/wiki/spaces/RTDel/pages/1562149077/ebXML+SOAP+MessageHeader+Ack+to+Spine?atlOrigin=eyJpIjoiODg4MDEwOTk3MDYxNGI2Y2FmZmU5YzBmNjRjMjIwOTkiLCJwIjoiYyJ9).

common/builder/pystache_message_builder.py
```diff
@@ -1,6 +1,11 @@
 """A module that defines a Pystache-based MessageBuilder."""
 import pystache
 from pystache import Renderer
+from pystache.common import MissingTags
+from pystache.context import KeyNotFoundError
+from utilities.integration_adaptors_logger import IntegrationAdaptorsLogger
```
Philip-Woods: Don't know if we're following/enforcing this so probably optional, but the google style guide says to avoid importing specific classes in favor of the module

common/builder/pystache_message_builder.py
Philip-Woods: Might suffice to use `MessageGenerationError` and pass in a string specifying it's due to a missing tag?

common/builder/tests/test_pystache_message_builder.py
```diff
@@ -1,7 +1,7 @@
 import os
 from unittest import TestCase
 
-from builder.pystache_message_builder import PystacheMessageBuilder
+from builder.pystache_message_builder import PystacheMessageBuilder, MessageGenerationSOAPFaultError
```
Philip-Woods: (Maybe optional) Again modules over classes

common/builder/tests/test_pystache_message_builder.py
```diff
@@ -19,3 +19,7 @@ def test_build_message(self):
 
         self.assertEqual("Hello, world!", message.strip(),
                          "Message returned should be the rendered string returned by Pystache")
+
+    def test_build_message_errors_on_missing_tag(self):
+        with self.assertRaises(MessageGenerationSOAPFaultError):
+            self.builder.build_message({})
```
Philip-Woods, line 25: As a part of the Gp summary upload there are optional sections, just to check that we want that behavior where this throws when an optional tag isn't present?

mhs/mhs/common/messages/ebxml_envelope.py
```diff
@@ -55,11 +56,14 @@ def serialize(self) -> typing.Tuple[str, str]:
...
-        message_id = message_utilities.MessageUtilities.get_uuid()
-        ebxml_message_dictionary[MESSAGE_ID] = message_id
+        message_id = ebxml_message_dictionary.get(MESSAGE_ID)
+        if not message_id:
+            message_id = message_utilities.MessageUtilities.get_uuid()
+            ebxml_message_dictionary[MESSAGE_ID] = message_id
         timestamp = message_utilities.MessageUtilities.get_timestamp()
         ebxml_message_dictionary[TIMESTAMP] = timestamp
-        logging.debug("Creating ebXML message with message ID '%s' and timestamp '%s'", message_id, timestamp)
+        logger.info('0001', 'Creating ebXML message with {MessageId} and {Timestamp}',
```
Philip-Woods, line 65: Just an idea but is it worth doing logging codes in the hundreds per method/ series of activities, this way when we add/remove a log call we don't have to update the whole file to shift the values.

mhs/mhs/common/workflow/sync_async.py
```diff
@@ -33,18 +33,24 @@ def __init__(self, config_manager: configuration_manager.ConfigurationManager,
         self.transmission = transmission
         self.party_id = party_id
 
-    def prepare_message(self, interaction_name: str, content: str) -> typing.Tuple[bool, str, str]:
+    def prepare_message(self, interaction_name: str, content: str, message_id: typing.Optional[str] = None) \
+            -> typing.Tuple[bool, str, str]:
```
Philip-Woods: As I was saying earlier google suggests importing modules not classes, however there is an exemption for the typing  [module](http://google.github.io/styleguide/pyguide.html#22-imports)

mhs/mhs/common/workflow/sync_async.py
```diff
@@ -33,18 +33,24 @@ def __init__(self, config_manager: configuration_manager.ConfigurationManager,
...
+        The purpose of this param is to be used for duplicate elimination when sending async messages.
         :return: A tuple containing:
         1. A flag indicating whether this message should be sent asynchronously
         2. the ID of the message, if an asynchronous response is expected, otherwise None
         3. The message to send
         """
         interaction_details = self._get_interaction_details(interaction_name)
 
+        if message_id is not None:
+            interaction_details[ebxml_envelope.MESSAGE_ID] = message_id
```
Philip-Woods, line 51: This will apply if `message_id` is an empty string, would there ever be a situation where we want an empty message id or should we just not have the pair?

mhs/mhs/outbound/request/synchronous/handler.py
```diff
@@ -28,15 +28,16 @@ def initialize(self, workflow: sync_async_workflow.SyncAsyncWorkflow,
         self.async_response_received = tornado.locks.Event()
         self.async_timeout = async_timeout
 
-    async def post(self):
+    async def post(self, interaction_name):
         logging.debug("Client POST received: %s", self.request)
```
Philip-Woods: (Optional as maybe it's worth refactoring it all at a later date) I know you didn't write this but we maybe we should try to update old logging statements as we go?

common/builder/tests/test_pystache_message_builder.py
ismail-s: If I'm thinking about this correctly:
If we want the optional tags to default to False then it seems that this would have to be done at a point where it is known what the template is (so that we know what the optional tags are), and then set defaults for those tags specifically. I haven't done this, I think we only need to do this if we find that there's a decent chunk of time we're spending specifying these default tags.

mhs/mhs/common/messages/ebxml_envelope.py
ismail-s: This feels like something to discuss with @dshipp-bjss.

mhs/mhs/common/workflow/sync_async.py
ismail-s: I doubt we'd want an empty message id, but I don't think we should do any validation here as this isn't the interface we expose to the outside world, and it feels like it might be overkill here. However, I think we should validate _somewhere_. I'm not sure if this needs to be part of this task, last week @gareth-allan mentioned to me that we've got some other task for sorting out the HTTP interface, so I'd imagine we'd do HTTP request validation at that time?

mhs/mhs/outbound/request/synchronous/handler.py
ismail-s: I agree (and @gareth-allan mentioned this to me as a general point as well).

common/builder/tests/test_pystache_message_builder.py
Philip-Woods: Seems reasonable to leave it as it is, like you say we can always rethink it if we're defaulting a lot of tags

mhs/mhs/common/workflow/sync_async.py
Philip-Woods: Fair point, we'll have to remember to do this somewhere else that's appropriate

ismail-s: I've mentioned this to Colin to put into the relevant story (as more of a general thing, not just for message id).

Philip-Woods (approved):

PR #44, opened 2019-08-01
Feature/RT-81 Persistence Adapter

flavgj:
Please ignore the branch name typo.


mhs/mhs/common/state/dynamo_persistence_adapter.py
```diff
@@ -0,0 +1,111 @@
+import json
+import traceback
+
+import boto3
+import utilities.integration_adaptors_logger as log
+
+from mhs.common.state.persistence_adapter import PersistenceAdapter
```
Philip-Woods: The style guide we're following is to import modules over individual classes, think these applies in a few other files as well

mhs/mhs/common/state/dynamo_persistence_adapter.py
```diff
@@ -0,0 +1,111 @@
...
+
+        dynamo_db = boto3.resource('dynamodb')
+
+        self.tables = {
+            state_table: dynamo_db.Table(state_arn),
+            sync_async_table: dynamo_db.Table(sync_async_arn)
+        }
+
+    def add(self, table_name, key, item):
+        try:
```
Philip-Woods: The other methods seem to be logging at the start which this is missing

mhs/mhs/common/state/dynamo_persistence_adapter.py
```diff
@@ -0,0 +1,111 @@
...
+            if 'Attributes' not in response:
+                logger.info("000", 'No previous record found: {key}', {"key": key})
+                return None
+            return json.loads(response.get('Attributes', {}).get('data'))
+        except Exception as e:
+            logger.error("001", 'Error creating record: {exception}', {"exception": traceback.format_exc()})
+            raise RecordCreationError from e
+
+    def get(self, table_name, key):
+        logger.info("002", 'Getting record for key {key}', {"key": key})
```
Philip-Woods: This will resolve as 'Getting record for key key=qweqweqwe', doesn't matter too much but it might look better as 
`logger.info("000", 'Getting record for {key}')`.

mhs/mhs/common/state/dynamo_persistence_adapter.py
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+    def add(self, table_name, key, item):
+        try:
+            response = self._resolve_table(table_name).put_item(
+                Item={"key": key, "data": json.dumps(item)},
+                ReturnValues='ALL_OLD'
+            )
+            if 'Attributes' not in response:
+                logger.info("000", 'No previous record found: {key}', {"key": key})
```
Philip-Woods: There's a few logs here that jump between single and double quotes, the style guide says to try to stick to one type and I think generally we're using single quotes for strings

mhs/mhs/common/state/persistence_adapter.py
```diff
@@ -1,8 +1,37 @@
 """
 This module defines the state adapter interface, used to allow support for multiple state database implementations.
 """
+from typing import Optional
```
Philip-Woods: I said above we're avoiding importing specific classes in favor of whole modules but just as a note this doesn't apply to the typing class so this is fine

mhs/mhs/common/state/persistence_adapter.py
```diff
@@ -1,8 +1,37 @@
...
 
 
 class PersistenceAdapter:
     """An adapter that provides a common interface to the state management database."""
     pass
```
Philip-Woods: Does this `pass` want removing?

mhs/mhs/common/state/persistence_adapter.py
```diff
@@ -1,8 +1,37 @@
...
+
+    def add(self, table_name, key, item: dict) -> Optional[dict]:
```
Philip-Woods: Think generally we were trying to include type hints for all methods

mhs/mhs/common/state/tests/test_dynamo_persistence_adaoter.py
```diff
@@ -0,0 +1,145 @@
...
+TEST_ITEM = {"key": TEST_KEY, "data": TEST_DATA}
+TEST_GET_RESPONSE = {"Item": TEST_ITEM}
+TEST_GET_EMPTY_RESPONSE = {}
+TEST_DELETE_RESPONSE = {"Attributes": TEST_ITEM}
+TEST_ADD_RESPONSE = {"Attributes": TEST_ITEM}
+TEST_EXCEPTION = Exception()
+TEST_SIDE_EFFECT = Mock(side_effect=TEST_EXCEPTION)
+
+
+class TestDynamoPersistenceAdapter(unittest.TestCase):
```
Philip-Woods: For consistency I think we're using the `Adaptor` spelling

mhs/mhs/common/state/tests/test_dynamo_persistence_adaoter.py
```diff
@@ -0,0 +1,145 @@
+import unittest
```
Philip-Woods: Typo in this files name

mhs/mhs/common/state/tests/test_dynamo_persistence_adaoter.py
flavgj: 50:50 split, the original interface was called persistence_adapter...

mhs/mhs/common/state/persistence_adaptor.py
```diff
@@ -0,0 +1,36 @@
+"""
+This module defines the state adaptor interface, used to allow support for multiple state database implementations.
+"""
+import typing
+
+
+class PersistenceAdaptor:
```
Philip-Woods: Optional - Should we make this subclass from `abc.ABC` just to show it's an interface?

Philip-Woods (commented):
Looks good to me, will give it a 👍 when it goes through the pipeline

mhs/mhs/common/state/dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,112 @@
...
+        pass
+
+
+class RecordRetrievalError(RuntimeError):
+    def __init__(self):
+        pass
+
+
+class InvalidTableError(RuntimeError):
+    def __init__(self):
```
ismail-s: This could probably be simplified to:
```python
class InvalidTableError(RuntimeError}:
    pass
```
If you really want to confuse people, you can actually replace `pass` with `...`

See [this SO answer](https://stackoverflow.com/a/1319675).

mhs/mhs/common/state/dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,112 @@
...
+        :class:`PersistenceAdaptor <mhs.common.state.persistence_adaptor.PersistenceAdaptor>`.
+        The kwargs provided should contain the following information:
+          * state_table: The name used to refer to the table containing state items.
+          * state_arn: The Amazon Resource Name used to identify the dynamo table containing state items.
+          * sync_async_table: The name used to refer to the table containing sync_async message items.
+          * sync_async_arn: The Amazon Resource Name used to identify the dynamo table containing sync_async message
+            items.
+        :param kwargs: The key word arguments required for this constructor.
+        """
+        state_table = kwargs.get('state_table', 'state')
```
ismail-s: Can mention default parameters in docstring. And tbh, do you even need to do this as `kwargs`, why not just have these as actual parameters. Then you can specify defaults there and they'd probably appear in the `Ctrl+q` docs Pycharm shows without you having to write them manually in the docstring.

Also, `state_table` and `sync_async_table`, I doubt we'd be setting them to different things in different parts of the codebase. So a simpler thing would be to just have them as constants in this class (or create an [enum](https://docs.python.org/3/library/enum.html) class), then whoever calls the different methods just passes the constant/enum value.

mhs/mhs/common/state/persistence_adaptor.py
```diff
@@ -0,0 +1,36 @@
...
+    """An adaptor that provides a common interface to the state management database."""
+
+    async def add(self, table_name: str, key: str, item: dict) -> typing.Optional[dict]:
```
ismail-s: According to [Google Python style guide](https://google.github.io/styleguide/pyguide.html#22-imports), import individual things from `typing`.

mhs/mhs/common/state/persistence_adaptor.py
ismail-s: I think that should be done, along with `abstractmethod` decorators, unless we have a good reason not to.

mhs/mhs/common/state/dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(self):
+        pass
+
+
+class InvalidTableError(RuntimeError):
+    def __init__(self):
+        pass
+
+
+class DynamoPersistenceAdaptor(mhs.common.state.persistence_adaptor.PersistenceAdaptor):
```
ismail-s, line 29: Have you tried the code in this file against an actual DynamoDB instance?

mhs/mhs/common/state/dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+    async def add(self, table_name, key, item):
+        logger.info('011', 'Adding {record} for {key}', {'record': item, 'key': key})
+        try:
+            response = await self._resolve_table(table_name).put_item(
+                Item={'key': key, 'data': json.dumps(item)},
+                ReturnValues='ALL_OLD'
+            )
+            if 'Attributes' not in response or response.get('Attributes', {}).get('data') is None:
```
ismail-s: What causes the 2 diffferent parts of this or?

mhs/mhs/common/state/dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,112 @@
...
+                logger.info('000', 'No previous record found: {key}', {'key': key})
+                return None
+            return json.loads(response.get('Attributes', {}).get('data'))
+        except Exception as e:
+            logger.error('001', 'Error creating record: {exception}', {'exception': traceback.format_exc()})
```
ismail-s, line 61: The logging package has some sort of functionality for including exception info, could that potentially be used? Also, I think we'd want this sort of functionality in an easy-to-use way in the logger component, as this is a common pattern. Also, looking at the Jenkins build output, the log message is being split over multiple lines, which I doubt Splunk would like, I think our logger component should somehow get rid of newlines in some way.

Parts of this comment could be done in a separate PR/task.

mhs/mhs/common/state/dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,112 @@
...
+
+        dynamo_db = aioboto3.resource('dynamodb')
+
+        self.tables = {
+            state_table: dynamo_db.Table(state_arn),
+            sync_async_table: dynamo_db.Table(sync_async_arn)
+        }
+
+    async def add(self, table_name, key, item):
+        logger.info('011', 'Adding {record} for {key}', {'record': item, 'key': key})
```
ismail-s: Can you include `table_name` in these log messages. Or just have a log message in `_resolve_table` that says which table has been gotten.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,176 @@
...
+from mhs.common.state import dynamo_persistence_adaptor
+from mhs.common.state.dynamo_persistence_adaptor import DynamoPersistenceAdaptor, RecordCreationError, \
+    InvalidTableError, RecordRetrievalError, RecordDeletionError
+
+TEST_STATE_ARN = "TEST STATE ARN"
+TEST_ASYNC_ARN = "TEST ASYNC ARN"
+TEST_ASYNC_TABLE = "TEST ASYNC TABLE"
+TEST_STATE_TABLE = "TEST STATE TABLE"
+TEST_INVALID_TABLE = "TEST INVALID TABLE"
+TEST_DATA = "{\"test_attribute\": \"test_value\"}"
```
ismail-s: Optional:
```suggestion
TEST_DATA = '{"test_attribute": "test_value"}'
```

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,176 @@
...
+TEST_DELETE_EMPTY_RESPONSE = {}
+TEST_ADD_EMPTY_RESPONSE = {"Attributes": {}}
+TEST_ADD_RESPONSE = {"Attributes": TEST_ITEM}
+TEST_EXCEPTION = Exception()
+TEST_SIDE_EFFECT = Mock(side_effect=TEST_EXCEPTION)
+
+
+class TestDynamoPersistenceAdaptor(unittest.TestCase):
+
+    @patch.object(dynamo_persistence_adaptor.aioboto3, "resource")
```
ismail-s: I think (double-check this though) that this decorator is only going to do the mocking for this `setUp` method. If that's not what you want (ie if you want the mock to persist through the test as well, then either move the decorator to decorate the class (but then you won't have access to the mock in the `setUp` methods, or use the [`start` and `stop` methods](https://docs.python.org/3/library/unittest.mock.html#patch-methods-start-and-stop).

I think the reason this is working as-is is because you're currently only referencing `aioboto3` in the `__init__` method of your adaptor, and then in this `setUp` method, you're maintaining a reference to your mock, so it doesn't get garbage-collected when the patch decorator exits.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,176 @@
...
+    def setUp(self, mock_boto3_resource):
+        self.mock_dynamodb = MagicMock()
+        mock_boto3_resource.return_value = self.mock_dynamodb
```
ismail-s: You shouldn't have to create `MagicMock`'s and attach them onto mocks, they magically get created, so you can do eg `self.mock_dynamodb = mock_boto3_resource()`.

ismail-s:
@flavgj https://github.com/pypa/pip/issues/2984 probably was causing this memory issue. Though I expect that the cache helps us a bit as we run `pipenv` multiple times.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,176 @@
...
+
+        response = await self.service.add(TEST_STATE_TABLE, TEST_KEY, TEST_ITEM)
+
+        self.assertIsNone(response)
+
+    @async_test
+    async def test_add_success_async(self):
+        self.__setFutureResponse(self.mock_table.put_item, TEST_ADD_EMPTY_RESPONSE)
+
+        response = await self.service.add(TEST_ASYNC_TABLE, TEST_KEY, TEST_ITEM)
```
ismail-s: Optional: This is virtually just the same as the previous test. Could combine the 2 tests as 2 subtests.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
```diff
@@ -0,0 +1,176 @@
...
+
+        self.assertIsInstance(ex.exception.__cause__, InvalidTableError)
+
+    @async_test
+    async def test_add_overwrite(self):
+        self.__setFutureResponse(self.mock_table.put_item, TEST_ADD_RESPONSE)
+
+        response = await self.service.add(TEST_STATE_TABLE, TEST_KEY, TEST_ITEM)
+
+        self.assertIsInstance(response, dict)
```
ismail-s: Couldn't you check that your response is whatever you expected it to be, rather than just checking the type? And same comment for the other tests where you do this.

ismail-s (changes requested):

mhs/mhs/common/state/dynamo_persistence_adaptor.py
flavgj: Yes, and no.

I have used very similar code previously. But I have not run this specific version of it. I will do so before actually merging this branch into 'develop' so that any required changes can be made.

mhs/mhs/common/state/dynamo_persistence_adaptor.py
flavgj: prefer to use a **kwargs here, as it means the contract for the constructors match as well.

mhs/mhs/common/state/dynamo_persistence_adaptor.py
flavgj: I'll add this.

mhs/mhs/common/state/dynamo_persistence_adaptor.py
flavgj: probably doesn't need to be split now that the right side is null safe.

mhs/mhs/common/state/persistence_adaptor.py
flavgj: done

mhs/mhs/common/state/dynamo_persistence_adaptor.py
flavgj: done

mhs/mhs/common/state/persistence_adaptor.py
flavgj: Done

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
flavgj: yes, as intended.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
flavgj: I prefer top-level given-when-then formatting in unit tests as it removes co-dependence of individual tests.

therefore any helper methods should be constrained to performing a single action, and not executing several operations within a test.

(purely preference, possibly worth further discussion.)

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
flavgj: This assert is validating the journey, and that the data flows through the service under test successfully. Testing actual data values passed in via mocks, would only validate the contents of those mocks. 

Additionally testing that json.loads() parses data into the correct format would be outside the scope of this unit test, and is not functionality we are responsible for.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
ismail-s: In this case, I mean [like this](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests), so you wouldn't have that codependence thing, ie if one subtest fails, then the other(s) should still get run.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
flavgj: done

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
flavgj: leaving as discussed.

mhs/mhs/common/state/dynamo_persistence_adaptor.py
flavgj: done

mhs/mhs/common/state/dynamo_persistence_adaptor.py
flavgj: RT-157 - Common logging package enhancements. 
https://gpitbjss.atlassian.net/browse/RT-157

ismail-s (dismissed):

Philip-Woods (dismissed):

mhs/mhs/common/state/dynamo_persistence_adaptor.py
```diff
@@ -10,57 +11,47 @@
...
+        :param data: The item to store in persistence.
+        :return: The previous version of the item which has been replaced. (None if no previous item)
+        """
+        logger.info('011', 'Adding {record} for {key}', {'record': data, 'key': key})
         try:
-            response = await self._resolve_table(table_name).put_item(
-                Item={'key': key, 'data': json.dumps(item)},
-                ReturnValues='ALL_OLD'
-            )
+            async with aioboto3.resource('dynamodb', region_name='eu-west-2') as dynamo_resource:
```
ismail-s: Extract common string. Actually, a better way is to extract the common with invocation. I think this can be done something like:
```python
@contextlib.asynccontextmanager
async def create_dynamo_table(self):
    async with aioboto3.resource('dynamodb', region_name='eu-west-2') as dynamo_resource:
        yield dynamo_resource.Table(self.table_name)
```
Possibly in here, you can log the table that's being used as well.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
```diff
@@ -30,146 +28,213 @@
...
-    #   TESTING ADD METHOD
-    @async_test
-    async def test_add_success_state(self):
-        self.__setFutureResponse(self.mock_table.put_item, TEST_ADD_EMPTY_RESPONSE)
-
-        response = await self.service.add(TEST_STATE_TABLE, TEST_KEY, TEST_ITEM)
-
-        self.assertIsNone(response)
-
+    @patch.object(dynamo_persistence_adaptor.aioboto3, "resource")
```
ismail-s: I believe that this decorator can be used as a class decorator instead of having to repeat it for each test in this class. Actually, you're also repeating `self.__configure_mocks` in all of these tests, so maybe instead don't do it as a class decorator, and do something like:
```python
def setUp(self):
    ...
    patcher = patch.object(...)
    self.mock = patcher.start()
    self.addCleanup(patcher.stop)
    self.__configure_mocks(mock)
```

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
```diff
@@ -30,146 +28,213 @@
...
+    """
+    A wrapper for asynchronous tests.
+    By default unittest will not wait for asynchronous tests to complete even if the async functions are awaited.
+    By annotating a test method with `@async_test` it will cause the test to wait for asynchronous activities
+    to complete. This will only use a single event loop for all tests with this annotation.
+    :param f: The coroutine which will be executed.
+    :return: a function which will execute an asynchronous unit test.
+    """
+
+    def wrapper(*args, **kwargs):
```
ismail-s, line 169: Optional: This is being pedantic, and you could argue isn't really required for a test, but can you stick `@wraps(f)` as a decorator for this function. [Python docs here](https://docs.python.org/3/library/functools.html#functools.wraps). It's the default way that decorators are created in the Python world.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
```diff
@@ -30,146 +28,213 @@
...
+        """
+        coro = asyncio.coroutine(f)
+        future = coro(*args, **kwargs)
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(future)
+
+    return wrapper
+
+
+@unittest.skip("This has been configured for a specific test dynamo instance which may not exist.")
```
ismail-s, line 185: So what is required to not skip these tests? Do we have a task covering unblocking them? Or should that be done as part of this task?

ismail-s (dismissed):
I think enough's done for me to approve. Comments have some suggestions though.

mhs/mhs/common/state/dynamo_persistence_adaptor.py
flavgj: Yes,

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
flavgj: These tests are just component integration tests to enable the verification the dynamodb code works with a real dynamodb. Unfortunately, this requires a real dynamodb and config to connect to it, and as component integration tests are not part of our test strategy, these have been ignored so as not to create test failures if their dependencies do not exist or have not been set up.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
flavgj: done.

mhs/mhs/common/state/dynamo_persistence_adaptor.py
flavgj: Done.

mhs/mhs/common/state/tests/test_dynamo_persistence_adaptor.py
flavgj: done.

ismail-s (approved):

PR #46, opened 2019-08-02
RT-94 message packer parsing

ismail-s:
- Separate better the parsing code that is specific to request/responses and acks
- Refactor code a little
- Set flags to False when elements don't exist, instead of not setting them
- Error when mandatory stuff isn't found during parsing
- Add more tests
- Chain an exception that was added in #42.
- Use [`defusedxml`](https://pypi.org/project/defusedxml/) to add some level of protection against parsing of malformed xml
  - Potentially we don't actually need this, if we accept Spine as a trusted source of xml, but this feels a bit iffy to me, I feel like we should try and make the MHS protected.
- [x] Decide on headers, whether it makes sense to do them in this task or not
- [x] Go through MIME code
- [x] Check code coverage, see if more tests are required
- [x] Type hints

### DIFF
--- a/mhs/mhs/common/messages/ebxml_envelope.py
+++ b/mhs/mhs/common/messages/ebxml_envelope.py
@@ -2,7 +2,7 @@
 import copy
 import pathlib
 import xml.etree.ElementTree as ElementTree
-from typing import Dict, Tuple
+from typing import Dict, Tuple, Any, Optional
 
 import builder.pystache_message_builder as pystache_message_builder
 import definitions
@@ -45,7 +45,7 @@ class EbxmlEnvelope(envelope.Envelope):
         {'name': RECEIVED_MESSAGE_ID, 'element_name': 'RefToMessageId', 'parent': 'MessageData'}
     ]
 
-    def __init__(self, template_file, message_dictionary: Dict[str, str]):
+    def __init__(self, template_file: str, message_dictionary: Dict[str, Any]):
         """Create a new EbxmlEnvelope that populates the specified template file with the provided dictionary.
 
         :param template_file: The template file to populate with values.
@@ -94,7 +94,7 @@ class EbxmlEnvelope(envelope.Envelope):
         return extracted_values
 
     @staticmethod
-    def _path_to_ebxml_element(name, parent=None):
+    def _path_to_ebxml_element(name: str, parent: str = None) -> str:
         path = ".//"
 
         if parent is not None:
@@ -105,7 +105,8 @@ class EbxmlEnvelope(envelope.Envelope):
         return path
 
     @staticmethod
-    def _extract_ebxml_value(xml_tree: ElementTree.Element, element_name, parent=None, required=False):
+    def _extract_ebxml_value(xml_tree: ElementTree.Element, element_name: str, parent: str = None,
+                             required: bool = False) -> Optional[ElementTree.Element]:
         xpath = EbxmlEnvelope._path_to_ebxml_element(element_name, parent=parent)
         value = xml_tree.find(xpath, namespaces=NAMESPACES)
         if value is None and required:
@@ -115,7 +116,8 @@ class EbxmlEnvelope(envelope.Envelope):
         return value
 
     @staticmethod
-    def _extract_ebxml_text_value(xml_tree: ElementTree.Element, element_name, parent=None, required=False):
+    def _extract_ebxml_text_value(xml_tree: ElementTree.Element, element_name: str, parent: str = None,
+                                  required: bool = False) -> Optional[str]:
         value = EbxmlEnvelope._extract_ebxml_value(xml_tree, element_name, parent, required)
         text = None
 
@@ -125,7 +127,8 @@ class EbxmlEnvelope(envelope.Envelope):
         return text
 
     @staticmethod
-    def _extract_attribute(xml_tree, element_name, attribute_namespace, attribute_name, values_dict, key):
+    def _extract_attribute(xml_tree: ElementTree.Element, element_name: str, attribute_namespace: Dict[str, str],
+                           attribute_name: str, values_dict: Dict[str, Any], key: str):
         xpath = EbxmlEnvelope._path_to_ebxml_element(element_name)
         element = xml_tree.find(xpath, NAMESPACES)
         if element is not None:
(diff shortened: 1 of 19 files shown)


mhs/mhs/common/messages/tests/test_ebxml_ack_envelope.py
```diff
@@ -61,11 +65,28 @@ def test_from_string(self):
...
     def test_from_string_with_no_values(self):
         message = file_utilities.FileUtilities.get_file_string(str(self.message_dir / "ebxml_header_empty.xml"))
 
-        parsed_message = ebxml_ack_envelope.EbxmlAckEnvelope.from_string({}, message)
+        with self.assertRaisesRegex(ebxml_envelope.EbXmlParsingError, "Weren't able to find required element"):
+            ebxml_ack_envelope.EbxmlAckEnvelope.from_string({}, message)
+
+    @patch.object(message_utilities.MessageUtilities, "get_timestamp")
+    @patch.object(message_utilities.MessageUtilities, "get_uuid")
+    def test_roundtrip(self, mock_get_uuid, mock_get_timestamp):
```
ismail-s, line 85: The reason I haven't added a roundtrip test to `test_ebxml_request_envelope.py` is because on the initial request, we don't send a `RefToMessageId`, but on the inbound request from Spine, we do get a `RefToMessageId` element (this makes sense as we send only the initial message id on our request to Spine, and when Spine sends a request back to us, it sends us a message id for the individual request and the message id we sent in previously). It's not that it wouldn't have been possible, but it would have required me to fiddle with the MIMEd xml message string.

mhs/mhs/common/messages/ebxml_envelope.py
```diff
@@ -68,67 +76,49 @@ def serialize(self) -> Tuple[str, str]:
...
 
     @staticmethod
-    def _extract_ebxml_text_value(xml_tree, element_name, parent=None):
-        value = EbxmlEnvelope._extract_ebxml_value(xml_tree, element_name, parent)
+    def _extract_ebxml_text_value(xml_tree: ElementTree.Element, element_name: str, parent: str = None,
+                                  required: bool = False) -> Optional[str]:
+        value = EbxmlEnvelope._extract_ebxml_value(xml_tree, element_name, parent, required)
         text = None
 
         if value is not None:
```
ismail-s, line 134: Here and [here](https://github.com/nhsconnect/integration-adaptors/pull/46/files#diff-f28d44ca274c270cb3ef934a33e38dd2R150), the unittests I've got don't test when the if statement is False. The reason why is that atm we don't parse any templates with optional text values. I considered creating some sort of contrived test for this scenario, but `TestEbxmlEnvelope` is used as a base class, so I'd have to somehow fiddle with that. I could maybe change the interface, but that feels a bit odd as well as sometime in the future we may want to optionally extract a text value.

Anyone got any thoughts as to what they'd prefer I do?

Philip-Woods: Maybe I don't understand fully what's going on here but would it suffice just to test this method and `parse_message` in isolation to ensure it can handle optional values? This should give us _some_ confidence it handles optionals without having to build a contrived test? Then maybe at a later date when we have optional text values we can build more suitable tests if needed

mhs/mhs/common/messages/ebxml_envelope.py
```diff
@@ -37,7 +33,19 @@
...
+        {'name': FROM_PARTY_ID, 'element_name': 'PartyId', 'parent': 'From'},
+        {'name': TO_PARTY_ID, 'element_name': 'PartyId', 'parent': 'To'},
+        {'name': CPA_ID, 'element_name': 'CPAId', 'parent': None},
+        {'name': CONVERSATION_ID, 'element_name': 'ConversationId', 'parent': None},
+        {'name': SERVICE, 'element_name': 'Service', 'parent': None},
+        {'name': ACTION, 'element_name': 'Action', 'parent': None},
+        {'name': MESSAGE_ID, 'element_name': 'MessageId', 'parent': 'MessageData'},
+        {'name': TIMESTAMP, 'element_name': 'Timestamp', 'parent': 'MessageData'},
+        {'name': RECEIVED_MESSAGE_ID, 'element_name': 'RefToMessageId', 'parent': 'MessageData'}
+    ]
```
Philip-Woods, line 51: Might be worth putting these key strings into variables?

mhs/mhs/common/messages/ebxml_envelope.py
```diff
@@ -68,67 +76,49 @@ def serialize(self) -> Tuple[str, str]:
...
-        EbxmlEnvelope._add_flag_if_present(extracted_values, ACK_REQUESTED,
-                                           EbxmlEnvelope._extract_ebxml_value(xml_tree, "AckRequested"))
-        EbxmlEnvelope._extract_attribute(xml_tree, "AckRequested", SOAP_NAMESPACE, "actor", extracted_values,
-                                         ACK_SOAP_ACTOR)
+        for element_to_extract in EbxmlEnvelope._elements_to_extract_when_parsing:
+            EbxmlEnvelope._add_if_present(extracted_values, element_to_extract['name'],
+                                          EbxmlEnvelope._extract_ebxml_text_value(xml_tree,
+                                                                                  element_to_extract['element_name'],
+                                                                                  parent=element_to_extract['parent'],
+                                                                                  required=True))
```
Philip-Woods, line 102: We discussed refactoring this during the PoC so it's good to see it being updated 👍

mhs/mhs/common/messages/ebxml_request_envelope.py
```diff
@@ -60,10 +77,14 @@ def _parse_mime_message(headers: Dict[str, str], message: str) -> email.message.
         :return: a Message that represents the message received.
         """
         content_type = headers[CONTENT_TYPE_HEADER_NAME]
-        content_type_header = CONTENT_TYPE_HEADER_NAME + ": " + content_type + "\r\n"
+        content_type_header = CONTENT_TYPE_HEADER_NAME + ": " + content_type + "\r\n\r\n"
```
Philip-Woods: Think somewhere else you were converting string concatenations to f-strings, probably makes sense to do this here as well.

Philip-Woods (commented):
Couple very minor points but looks good to me

mhs/mhs/common/messages/ebxml_envelope.py
ismail-s: Started to do this, then felt like namedtuples might be a bit nicer. See d6144df. If you don't like it, I can change it back, but I think it makes it more readable, though I half-think I've made it too confusing with this.

Philip-Woods (approved):

PR #47, opened 2019-08-02
Convert to strings when logging

ismail-s:
Convert values to strings when logging, so we don't have to do this in our log statements.


Philip-Woods (approved):
Seems fine to me

PR #48, opened 2019-08-05
Feature/RT82-State-Management

Philip-Woods:
Creates a state store which loads and updates a local copy of a work description.

This should be ready for review, though it should be noted it is dependent on PR #44 so any changes to the interface of the persistence adaptor might affect this PR.


mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+        self.version = data[VERSION_KEY]
+        self.timestamp = data[TIMESTAMP]
+        self.status = data[STATUS]
+
+    async def publish(self):
+        logger.info('011', 'Attempting to publish work description {key}', {'key': self.message_key})
+        logger.info('012', 'Retrieving latest work description to check version')
+        latest_data = await self.persistence_store.get(self.table_name, self.message_key)
+        if latest_data is not None:
+            latest_version = latest_data[DATA][VERSION_KEY]
```
ismail-s, line 125: Optional: could log whether `latest_data is None` or not

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+import utilities.integration_adaptors_logger as log
+
+
+logger = log.IntegrationAdaptorsLogger('STATE_MANAGER')
+
+
+class MessageStatus:
+    RECEIVED = 1
+    STARTED = 2
+    IN_OUTBOUND_WORKFLOW = 3
```
ismail-s, line 14: Are these just dummy values for the timebeing?

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+        serialised = self._serialise_data()
+
+        old_data = await self.persistence_store.add(self.table_name, serialised)
+        logger.info('014', 'Successfully updated work description to state store for {key}', {'key': self.message_key})
+        return old_data
+
+    def _serialise_data(self):
+        data = {
+            DATA_KEY: self.message_key,
+            'DATA': {
```
ismail-s: Why not use the constant here?

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+
+class OutOfDateVersionError(RuntimeError):
+    pass
+
+
+class EmptyWorkDescriptionError(RuntimeError):
+    pass
+
+
+class WorkDescription:
```
ismail-s: Generally, missing docstrings on stuff.

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+DATA = 'DATA'
+STATUS = 'STATUS'
+
+
+class OutOfDateVersionError(RuntimeError):
+    pass
+
+
+class EmptyWorkDescriptionError(RuntimeError):
+    pass
```
ismail-s, line 32: Optional: possibly have a custom base exception for both of these?

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+                TIMESTAMP: self.timestamp,
+                VERSION_KEY: self.version,
+                STATUS: self.status
+            }
+        }
+
+        return json.dumps(data)
+
+
+class WorkDescriptionFactory:
```
ismail-s: Optional: I don't think these functions need to be in a class. I'm thinking more in terms of imports, though if I'm thinking that, then maybe our import rule is giving us too verbose code.

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+            logger.error('003', 'Persistence store returned empty value for {key}', {'key': key})
+            raise EmptyWorkDescriptionError(f'Failed to find a value for key id {key}')
+
+        return WorkDescription(persistence_store, table_name, json_store_data)
+
+    @staticmethod
+    def create_new_work_description(persistence_store,
+                                    table_name: str,
+                                    key: str,
+                                    timestamp: str,
```
ismail-s: What is timestamp, is it creation time or last modified time? And could it be managed by this file?

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+        if status is None:
+            logger.error('007', 'Failed to build new work description, status should not be null')
+            raise ValueError('Expected status to not be None')
+
+        work_description_map = {
+            DATA_KEY: key,
+            DATA: {
+                TIMESTAMP: timestamp,
+                STATUS: status,
+                VERSION_KEY: 1
```
ismail-s: So is the idea behind this version key that it gets incremented on all (committed) modifications to this work description object? If so, could it be maintained by this file instead?

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+
+        json_store_data = await persistence_store.get(table_name, key)
+        if json_store_data is None:
+            logger.error('003', 'Persistence store returned empty value for {key}', {'key': key})
+            raise EmptyWorkDescriptionError(f'Failed to find a value for key id {key}')
+
+        return WorkDescription(persistence_store, table_name, json_store_data)
+
+    @staticmethod
+    def create_new_work_description(persistence_store,
```
ismail-s: So if I understand correctly, wherever this function is used, it's called, and then the resulting object has the `publish` method invoked on it?

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+        self.timestamp = data[TIMESTAMP]
+        self.status = data[STATUS]
+
+    async def publish(self):
+        logger.info('011', 'Attempting to publish work description {key}', {'key': self.message_key})
+        logger.info('012', 'Retrieving latest work description to check version')
+        latest_data = await self.persistence_store.get(self.table_name, self.message_key)
+        if latest_data is not None:
+            latest_version = latest_data[DATA][VERSION_KEY]
+            if latest_version > self.version:
```
ismail-s: And what about if the versions are equal?

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+                logger.error('013', 'Failed to update message {key}, local version out of date',
+                             {'key': self.message_key})
+                raise OutOfDateVersionError(f'Failed to update message {self.message_key}: local version out of date ')
```
ismail-s: V. v. minor:
```suggestion
                raise OutOfDateVersionError(f'Failed to update message {self.message_key}: local version out of date')
```

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+
+        serialised = self._serialise_data()
+
+        old_data = await self.persistence_store.add(self.table_name, serialised)
```
ismail-s: I think code to handle failure is missing here.

mhs/mhs/common/state/tests/test_work_description.py
```diff
@@ -0,0 +1,157 @@
...
+            wd.DATA_KEY: 'aaa-aaa-aaa',
+            wd.DATA: {
+                wd.VERSION_KEY: 1,
+                wd.TIMESTAMP: '11:00',
+                wd.STATUS: wd.MessageStatus.IN_OUTBOUND_WORKFLOW
+            }
+        }
+
+
+class TestWorkDescription(unittest.TestCase):
```
ismail-s, line 30: Looking at code coverage reports, it looks like tests are missing for:
- In `publish`, `latest_data is None`
- In `get_work_description_from_store`, `json_store_data is None`
- In `create_new_work_description`, `persistence_store  is None`

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,122 @@
...
+        data = store_data[DATA]
+        self.message_key = store_data[DATA_KEY]
+        self.version = data[VERSION_KEY]
+        self.timestamp = data[TIMESTAMP]
+        self.status = data[STATUS]
+
+    async def publish(self):
+        logger.info('011', 'Attempting to publish work description {key}', {'key': self.message_key})
+        logger.info('012', 'Retrieving latest work description to check version')
+        latest_data = await self.persistence_store.get(self.table_name, self.message_key)
```
ismail-s: So a get is done first, then the version check is done, then based on that, the db is/isn't updated. If multiple things are doing this at the same time, then we've got a race condition. We just need to be careful that if we want to have a failsafe around multiple updates, then we're fine with this possible race condition, otherwise we should move the version check onto the database.

ismail-s (changes requested):

mhs/mhs/common/state/work_description.py
Philip-Woods: I'm going to say yes, we didn't have specific values for these to be set to I don't think so I just treated it as a temporary store for status values

mhs/mhs/common/state/work_description.py
Philip-Woods: Should be updated to be more descriptive now

mhs/mhs/common/state/work_description.py
Philip-Woods: Would this give us any particular benefit?

mhs/mhs/common/state/work_description.py
Philip-Woods: Yeah that's a good point, I can move them out of the class but does this raise the question as to if we should change that import rule?

mhs/mhs/common/state/work_description.py
Philip-Woods: I believe it's meant to be the last modified, yes it could be managed by this file but I figured I would allow complete control to the workflow manager as to the details of this state object. 

I can update it if you think it's preferential to have this updated automatically?

mhs/mhs/common/state/work_description.py
Philip-Woods: Updated to raise exception if theyre equal as well

mhs/mhs/common/state/work_description.py
Philip-Woods: Do we want to handle failure here? The persistence store has the possibility to raise a record creation exception, I don't know what we would do to handle this here, this feels like an exception to be handled by the caller?

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,156 @@
...
+            logger.info('015', 'No previous version found, continuing attempt to publish new version')
+        serialised = self._serialise_data()
+
+        old_data = await self.persistence_store.add(self.table_name, serialised)
+        logger.info('016', 'Successfully updated work description to state store for {key}', {'key': self.message_key})
+        return old_data
+
+    def _serialise_data(self):
+        """
+        A simple serialization method that turns the data local data into a json string which can be stored in the
```
ismail-s:
```suggestion
        A simple serialization method that turns the local data into a json string which can be stored in the
```

mhs/mhs/common/state/tests/test_work_description.py
Philip-Woods: Tried to parameterize the null checks in `create_new_work_description` but the parameterize library failed for `None` so I just had to make the test a bit longer

mhs/mhs/common/state/work_description.py
ismail-s: Yeah, not massively. In my head, it means that other places could catch the base exception rather than these ones, but I'm not sure how we're going to be handling these exceptions (if explicitly catching them in the first place), so not that fussed about this one.

mhs/mhs/common/state/work_description.py
Philip-Woods: Good point, I guess we'll have to remember to place the check on the database as well.

mhs/mhs/common/state/work_description.py
ismail-s: I think it is better to have it updated automatically as we're going to have to update it somewhere, might as well be here.

Last modified or created? What about both? Probably a question for Dave.

mhs/mhs/common/state/work_description.py
ismail-s: Ignore me. When looking at #44, I saw that the `add` method could return `None`. But looking closer, that's only when there was no previous record, which you're already checking for.

mhs/mhs/common/state/work_description.py
ismail-s: @flavgj @dshipp-bjss any thoughts?

mhs/mhs/common/state/work_description.py
ismail-s: I don't see this as having a check in 2 places. If we wanted to do it, I think we'd have to use [something like this](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html). This then gets more complicated as I don't know how this plays with DynamoDBs eventual consistency. Feels like a Dave question.

mhs/mhs/common/state/tests/test_work_description.py
```diff
@@ -4,6 +4,7 @@
 import mhs.common.state.work_description as wd
 from unittest.mock import MagicMock, patch
 from utilities.test_utilities import async_test
+import parameterized
```
ismail-s: Should this be here?

mhs/mhs/common/state/work_description.py
dshipp-bjss: I’d envisaged that they’d be static members of the WorkDescription class rather than in their own class. They could also be stand alone functions - I’m not precious

mhs/mhs/common/state/work_description.py
dshipp-bjss: Do both

mhs/mhs/common/state/work_description.py
Philip-Woods: I'll make them as standalone functions simply since the code guidelines say to try and avoid static methods where possible

mhs/mhs/common/state/work_description.py
```diff
@@ -13,7 +13,8 @@ class MessageStatus:
 
 DATA_KEY = 'MESSAGE_KEY'
 VERSION_KEY = 'VERSION'
-TIMESTAMP = 'TIMESTAMP'
+CREATED_TIMESTAMP = 'CREATED'
+LAST_MODIFIED_TIMESTAMP = 'TIMESTAMP'
```
ismail-s: Bit of a naff name for this one.

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,155 @@
...
+        logger.info('011', 'Attempting to publish work description {key}', {'key': self.message_key})
+        logger.info('012', 'Retrieving latest work description to check version')
+
+        latest_data = await self.persistence_store.get(self.message_key)
+        if latest_data is not None:
+            logger.info('013', 'Retrieved previous version, comparing versions')
+            latest_version = latest_data[DATA][VERSION_KEY]
+            if latest_version == self.version:
+                logger.info('017', 'Local version matches remote, incrementing local version number')
+                self.version = self.version + 1
```
ismail-s: Minor/optional:
```suggestion
                self.version += 1
```

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,155 @@
+import json
+import utilities.integration_adaptors_logger as log
+import datetime
+
+logger = log.IntegrationAdaptorsLogger('STATE_MANAGER')
+
+
+class MessageStatus:
```
ismail-s, line 11: You could actually use [enum](https://docs.python.org/3.7/library/enum.html) (should have thought of this when first reviewing this).

mhs/mhs/common/state/work_description.py
```diff
@@ -0,0 +1,155 @@
...
+    """Exception thrown when no work description is found for a given key"""
+    pass
+
+
+def get_time():
+    """Returns UTC time in the appropriate format """
+    return datetime.datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S.%f')
+
+
+async def get_work_description_from_store(persistence_store, key: str):
```
ismail-s:
```suggestion
async def get_work_description_from_store(persistence_store, key: str) -> WorkDescription:
```
Might have to do that as a string or do `from __future__ import annotations`

mhs/mhs/common/state/tests/test_work_description.py
```diff
@@ -72,26 +75,52 @@ def test_constructor(self):
...
+
+        persistence = MagicMock()
+        persistence.get.return_value = future
+        persistence.add.return_value = future
+        work_description = wd.WorkDescription(persistence, input_data)
+        await work_description.publish()
+
+        updated = copy.deepcopy(input_data)
+        updated[wd.DATA][wd.VERSION_KEY] = 2
+        print(updated)
```
ismail-s: What's this printing about, is it meant to be here?

ismail-s (dismissed):
Approving as mostly there. Still got build failures.

mhs/mhs/common/state/work_description.py
Philip-Woods: I actually used enum at first, but it didn't serialize to json without a custom serialization method which seemed a bit long winded/ overkill considering I wasn't fully sure how it was going to be used in the code yet, if we find it might be useful/cleaner to define it as an enum I'm happy update it

Philip-Woods:
I've addressed all the current comments so should be ok for approval again, it probably makes sense for #44 to be merged first then I can update the type hints for the persistence adaptor parameters

mhs/mhs/common/state/work_description.py
```diff
@@ -14,7 +15,7 @@ class MessageStatus:
 DATA_KEY = 'MESSAGE_KEY'
 VERSION_KEY = 'VERSION'
 CREATED_TIMESTAMP = 'CREATED'
-LAST_MODIFIED_TIMESTAMP = 'TIMESTAMP'
+LATEST_TIMESTAMP = 'TIMESTAMP'
```
ismail-s: Ah, what I actually meant by https://github.com/nhsconnect/integration-adaptors/pull/48#discussion_r311479442 was the string part.

ismail-s (dismissed):

mhs/mhs/common/state/work_description.py
ismail-s: I was about to suggest `IntEnum`, but then that wouldn't be that readable in the database. Only other thing I can think of is to just use the value only when serialising to json, but meh, we can look at this later in the tasks that use this code.

ismail-s: Since writing the above comment, I found what I was looking for: According to [here](https://docs.python.org/3.7/library/enum.html#others), you can create a string enum pretty simply. And I expect that that would serialise in the manner you want it to. Then you can use the [`unique`](https://docs.python.org/3.7/library/enum.html#ensuring-unique-enumeration-values) decorator as well.

ismail-s (approved):

PR #49, opened 2019-08-06
Feature/RT-144 queue adaptor implementation for publish

flavgj:


common/local-env/rabbit/dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM rabbitmq:3.7-management
+RUN rabbitmq-plugins enable --offline rabbitmq_amqp1_0
```
ismail-s: Not sure if this will add 2 newlines or one, but:
```suggestion
RUN rabbitmq-plugins enable --offline rabbitmq_amqp1_0

```

common/setup.py
```diff
@@ -13,6 +13,6 @@
     description='Common utilities used by the NHS integration adaptors projects.',
     install_requires=[
         'pystache',
-        'lxml'
+        'lxml', 'python-qpid-proton', 'tornado'
```
ismail-s: V. minor/optional: formatting could be a bit nicer.

common/Pipfile
```diff
@@ -20,3 +21,7 @@ unittests-cov = 'coverage run -m xmlrunner -o test-reports -v'
 coverage-report = 'coverage report'
 coverage-report-xml = 'coverage xml'
 coverage-report-html = 'coverage html'
+local-env-prep = 'docker build -t rabbit-amqp1 local-env/rabbit/'
+local-env-start = 'docker run --rm -d --hostname localhost --name rabbit-amqp-1 -p 15672:15672 -p 5672:5672 rabbit-amqp1'
+local-env-stop = 'docker stop rabbit-amqp-1'
+local-int-test = 'python'
```
ismail-s: Readme should be updated to explain usage of these scripts. Also, I can't see any changes to the Jenkinsfile to run these integration tests, and more importantly, this integration test script is wrong, it's just running a Python interpreter. Also, I still feel like it would be better to encapsulate all of these scripts in a single Python test, then it's just a single script here to run that manages building/running/stopping Docker for you.

common/comms/proton_queue_adaptor.py
```diff
@@ -0,0 +1,105 @@
...
+        message_id = str(uuid.uuid4())
+        logger.info('001', 'Constructing message with {id} for {body}', {'id': message_id, 'body': message})
+        return proton.Message(id=message_id, body=message)
+
+    def __send(self, message: proton.Message) -> None:
+        """
+        Performs a synchronous send of a message, to the host defined when this adaptor was constructed.
+        :param message: The message to be sent.
+        """
+        proton.reactor.Container(ProtonMessagingHandler(self.host, message)).run()
```
ismail-s, line 63: If I understand correctly, you're establishing a new connection each time as opposed to reusing an existing connection. Not amazing, but I'm not going to say that it's an issue per se as performance testing hasn't been done on it.

common/Pipfile
```diff
@@ -7,9 +7,10 @@ verify_ssl = true
 coverage = "~=4.5"
 tornado = "*"
 unittest-xml-reporting = "~=2.5"
+python-qpid-proton = "*"
```
Philip-Woods: Think this wants to be in the `[packages]` section since it's needed in live code not just tests?

common/Pipfile
Philip-Woods: What's this `local-int-test`?

common/comms/queue_adaptor.py
```diff
@@ -0,0 +1,20 @@
+"""Module for generic queue adaptor functionality"""
+import abc
+
+
+class QueueAdaptor(abc.ABC):
+    """Interface for a message queue adaptor."""
+
+    async def send_async(self, message: str) -> None:
```
ismail-s, line 9: Possibly missing `abstractmethod` decorators.

common/comms/queue_adaptor.py
```diff
@@ -0,0 +1,20 @@
...
+        """
+        Sends a message with awaits using the async flow.
+        :param message: The message content to send.
+        """
+        pass
+
+    def send_sync(self, message: str) -> None:
```
ismail-s, line 17: I don't see why this is part of the interface, when would we ever use it?

common/comms/proton_queue_adaptor.py
```diff
@@ -0,0 +1,106 @@
...
+    """The connection to the Message Queue ended before sending of the message had been done."""
+    pass
+
+
+class ProtonQueueAdaptor(comms.queue_adaptor.QueueAdaptor):
+    """Proton implementation of a queue adaptor."""
+
+    def __init__(self, **kwargs) -> None:
+        """
+        Constrcut a Proton implementation of a :class:`QueueAdaptor <comms.queue_adaptor.QueueAdaptor>`.
```
ismail-s:
```suggestion
        Construct a Proton implementation of a :class:`QueueAdaptor <comms.queue_adaptor.QueueAdaptor>`.
```

common/comms/proton_queue_adaptor.py
```diff
@@ -0,0 +1,106 @@
...
+        The kwargs provided should contain the following information:
+          * host: The host of the Message Queue to be interacted with.
+        :param kwargs: The key word arguments required for this constructor.
+        """
+        super(ProtonQueueAdaptor, self).__init__()
```
ismail-s: [I don't think you need to pass any arguments](https://docs.python.org/3/library/functions.html#super):
```suggestion
        super().__init__()
```

common/comms/proton_queue_adaptor.py
```diff
@@ -0,0 +1,106 @@
...
+class ProtonMessagingHandler(proton.handlers.MessagingHandler):
+    """Implementation of a Proton MessagingHandler which will send a single message."""
+
+    def __init__(self, host, message: proton.Message) -> None:
+        """
+        Constructs a MessagingHandler which will send a specified message to a specified host.
+        :param host: The host to send the message to.
+        :param message: The message to be sent to the host.
+        """
+        super(ProtonMessagingHandler, self).__init__()
```
ismail-s:
```suggestion
        super().__init__()
```

common/comms/proton_queue_adaptor.py
```diff
@@ -0,0 +1,106 @@
...
+        self.sender = event.container.create_sender(self.host)
+
+    def on_sendable(self, event):
+        if event.sender.credit:
+            if not self.sent:
+                event.sender.send(self.message)
+                logger.info('003', 'Message sent to {host}.', {'host': self.host})
+                self.sent = True
+        else:
+            logger.error('004', 'Failed to send message as no available credit.')
```
ismail-s, line 92: What is credit?

common/comms/tests/test_proton_queue_adaptor.py
```diff
@@ -0,0 +1,155 @@
...
+import asyncio
+import unittest.mock
+
+import proton
+
+import comms.proton_queue_adaptor
+import utilities.test_utilities
+
+TEST_MESSAGE = "TEST MESSAGE"
+TEST_PROTON_MESSAGE = unittest.mock.Mock()
```
ismail-s, line 9: Imports are a bit long.

common/comms/tests/test_proton_queue_adaptor.py
```diff
@@ -0,0 +1,155 @@
...
+    def setUp(self) -> None:
+        self.service = comms.proton_queue_adaptor.ProtonQueueAdaptor(host=TEST_QUEUE_HOST)
+
+    # TESTING SEND ASYNC METHOD
+    @unittest.mock.patch.object(proton.reactor, "Container")
+    @utilities.test_utilities.async_test
+    async def test_send_async_success(self, mock_container):
+        """Test happy path of send_async."""
+        mock_container.run.return_value = asyncio.Future()
+        await self.service.send_async(TEST_MESSAGE)
```
ismail-s, line 31: Feels like for this sort of test you'd want to use the `AsyncTestCase` provided by Tornado, as you're using the Tornado `run_in_executor` method.

Also, these tests aren't asserting that anything has been done.

ismail-s (changes requested):

common/comms/proton_queue_adaptor.py
```diff
@@ -0,0 +1,106 @@
...
+        self.__send(self.__construct_message(message))
+
+    @staticmethod
+    def __construct_message(message: str) -> proton.Message:
+        """
+        Build a message with a generated uuid, and specified message body.
+        :param message: The message body to be wrapped.
+        :return: The Message in the correct format with generated uuid.
+        """
+        message_id = str(uuid.uuid4())
```
Philip-Woods: There's a method in `message_utilities` that will generate uuids for us, it's only 1 line so I'm not that fussed but I figured it makes sense to get uuids from the same place

common/comms/proton_queue_adaptor.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        Constructs a MessagingHandler which will send a specified message to a specified host.
+        :param host: The host to send the message to.
+        :param message: The message to be sent to the host.
+        """
+        super(ProtonMessagingHandler, self).__init__()
+        self.host = host
+        self.message = message
+        self.sender = None
+        self.sent = False
```
Philip-Woods: Just to check but are any of these modified externally? Just wondering if they should be prefixed with `_`

common/local-env/rabbit/dockerfile
flavgj: done locally.

common/setup.py
flavgj: done

common/Pipfile
flavgj: Done, I think the approach for this may change in the future, but these provide a start-point.

common/comms/proton_queue_adaptor.py
flavgj: Yes, not ideal, but as this has to run in a separate executor, and there doesn't appear to be a way to pass an existing connection, connection pool, or connection provider, and the container closes the connection after completion of the send, without significant effort to 'make proton do what we want' this seems the simplest implementation. (further work may be required if performance is not good enough, but at the moment we don't need to spend that effort) - proton may be replaced by pika as the library of choice if a newer version of pika supports amqp 1.0. Additionally there is a library named uamqp which could be used to provide an async alternative to proton.

common/comms/queue_adaptor.py
flavgj: I exposed this as part of the interface in cases where an alternative means than using a tornado executor to cause the sync functionality to be run async was desired.

common/comms/proton_queue_adaptor.py
flavgj: Done locally.

common/comms/proton_queue_adaptor.py
flavgj: Done locally

common/comms/proton_queue_adaptor.py
flavgj: Done locally

common/comms/proton_queue_adaptor.py
flavgj: Capacity within the broker for this publisher to send messages on the queue.

common/comms/tests/test_proton_queue_adaptor.py
flavgj: unfortunately, as we rely on proton to provide the actual functionality, and the only modified state would be external to the application there is very little to assert other than the container has been run.

common/Pipfile
flavgj: Done

common/Pipfile
flavgj: Gone.

common/comms/proton_queue_adaptor.py
flavgj: Thanks. Done.

common/comms/proton_queue_adaptor.py
flavgj: Done.

common/Pipfile
```diff
@@ -5,12 +5,12 @@ verify_ssl = true
 
 [dev-packages]
 coverage = "~=4.5"
-tornado = "*"
 unittest-xml-reporting = "~=2.5"
-python-qpid-proton = "*"
 
 [packages]
 integration-adaptors-common = {editable = true, path = "."}
+tornado = "*"
```
ismail-s, line 12: `Pipfile.lock` needs regenerating.

flavgj: Did

ismail-s (approved):

PR #50, opened 2019-08-07
RT137 - Logging configuration

Philip-Woods:
This PR updates the main file to load the correct logging config, this fixes the incorrect date format. Top level exception handling has also been added in order to ensure any unhandled exceptions are logged appropriately


mhs/main.py
```diff
@@ -112,4 +107,9 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except Exception as e:
+        logger.critical('001', 'Fatal exception in main application: {exception}', {'exception': e})
+    finally:
+        logger.warning('002', 'Exiting application')
```
ismail-s: I think this should just be INFO, as you could want to exit the application deliberately.

common/utilities/integration_adaptors_logger.py
```diff
@@ -14,15 +14,29 @@ def configure_logging():
...
     logger.setLevel(config.get_config('LOG_LEVEL'))
     handler = logging.StreamHandler(sys.stdout)
-    logging.Formatter.converter = time.gmtime
-    formatter = logging.Formatter('[%(asctime)s.%(msecs)03dZ] '
-                                  '%(message)s pid=%(process)d LogLevel=%(levelname)s ',
-                                  '%Y-%m-%dT%H:%M:%S')
+
+    formatter = _CustomFormatter(
+        fmt='[%(asctime)sZ] %(message)s pid=%(process)d LogLevel=%(levelname)s ',
+        datefmt='%Y-%m-%dT%H:%M:%S.%f')
```
ismail-s, line 36: Why didn't you have to update the unittests?

Philip-Woods: The tests didn't test the whole string, usually just either that the `log.write` was called with particular parameters or that the output log string contains particular attributes. For completeness I just added a test that asserts the format of the time in the output log

common/utilities/tests/test_logger.py
```diff
@@ -162,3 +163,13 @@ def test_write_throws_error_on_bad_params(self):
     def test_undefined_log_ref_throws_error(self):
         with self.assertRaises(ValueError):
             log.IntegrationAdaptorsLogger('')
+
+    @patch('sys.stdout', new_callable=io.StringIO)
+    def test_correct_time_format(self, mock_stdout):
```
ismail-s, line 168: Don't forget to raise some techdebt for figuring out the issue where this test name actually affects the test passing/failing.

ismail-s (approved):

PR #51, opened 2019-08-09
RT-150 client request handler

ismail-s:
This is one of the first parts in supporting an async express workflow. This PR is focussed on getting the outbound sync request handler up to scratch.

- HTTP request from supplier now can have message id, interaction id and correlation id as headers
- Use [`contextvars`](https://docs.python.org/3/library/contextvars.html) to simplify including message id & correlation id in our logs
- Lookup interaction details in handler, not in workflow. And then use the interaction details lookup to decide which workflow to execute
- Do some basic validation of HTTP request
- Remove all the sync-async stuff from the handler. This will be re-implemented again later, when we work on the sync-async workflow.

# Notes
- I haven't separated code between the outbound sync request handler and common outbound, as I see no benefit in doing so (it looks like we'll only have the sync request handler, not an async one).
- I haven't put any prefix on the header names. I'm not sure if we should (I considered doing so, but then though maybe not, as this would have to make sense with the adaptors and the supplier system).

### DIFF
--- a/mhs/main.py
+++ b/mhs/main.py
@@ -107,6 +107,6 @@ if __name__ == "__main__":
     try:
         main()
     except Exception as e:
-        logger.critical('001', 'Fatal exception in main application: {exception}', {'exception': e})
+        logger.critical('002', 'Fatal exception in main application: {exception}', {'exception': e})
     finally:
-        logger.info('002', 'Exiting application')
+        logger.info('003', 'Exiting application')
--- a/mhs/main.py
+++ b/mhs/main.py
@@ -44,21 +44,18 @@ def load_party_key(data_dir: pathlib.Path) -> str:
     return party_key
 
 
-def initialise_workflow(data_dir: pathlib.Path, certs_dir: pathlib.Path,
+def initialise_workflow(certs_dir: pathlib.Path,
                         party_key: str) -> sync_async_workflow.SyncAsyncWorkflow:
-    """Initialise the
+    """Initialise the workflow
 
-    :param data_dir: The directory to load interactions configuration from.
     :param certs_dir: The directory containing certificates/keys to be used to identify this MHS to a remote MHS.
     :param party_key: The party key to use to identify this MHS.
     :return: The workflow that can be used to handle messages.
     """
-    interactions_config_file = str(data_dir / "interactions" / "interactions.json")
-    config_manager = configuration_manager.ConfigurationManager(interactions_config_file)
 
     transmission = outbound_transmission.OutboundTransmission(str(certs_dir))
 
-    workflow = sync_async_workflow.SyncAsyncWorkflow(config_manager, transmission, party_key)
+    workflow = sync_async_workflow.SyncAsyncWorkflow(transmission, party_key)
 
     return workflow
 
@@ -68,10 +65,11 @@ def configure_logging() -> None:
     logging.basicConfig(format="%(asctime)s - %(levelname)s: %(message)s", level=config.config["LOG_LEVEL"])
 
 
-def start_tornado_servers(certs_file: str, key_file: str, workflow: sync_async_workflow.SyncAsyncWorkflow,
+def start_tornado_servers(data_dir: pathlib.Path, certs_file: str, key_file: str, workflow: sync_async_workflow.SyncAsyncWorkflow,
                           party_key: str) -> None:
     """
 
+    :param data_dir: The directory to load interactions configuration from.
     :param certs_file: The filename of the certificate to be used to identify this MHS to a remote MHS.
     :param key_file: The filename of the private key for the certificate identified by certs_file.
     :param workflow: The workflow to be used to handle messages.
@@ -79,9 +77,12 @@ def start_tornado_servers(certs_file: str, key_file: str, workflow: sync_async_w
     """
     callbacks = {}
 
+    interactions_config_file = str(data_dir / "interactions" / "interactions.json")
+    config_manager = configuration_manager.ConfigurationManager(interactions_config_file)
+
     supplier_application = tornado.web.Application(
         [(r"/", client_request_handler.SynchronousHandler,
-          dict(workflow=workflow, callbacks=callbacks, async_timeout=ASYNC_TIMEOUT))])
+          dict(config_manager=config_manager, workflow=workflow, callbacks=callbacks, async_timeout=ASYNC_TIMEOUT))])
     supplier_server = tornado.httpserver.HTTPServer(supplier_application)
     supplier_server.listen(80)
 
@@ -106,9 +107,9 @@ def main():
     certs_file, key_file = load_certs(certs_dir)
     party_key = load_party_key(certs_dir)
 
-    workflow = initialise_workflow(data_dir, certs_dir, party_key)
+    workflow = initialise_workflow(certs_dir, party_key)
 
-    start_tornado_servers(certs_file, key_file, workflow, party_key)
+    start_tornado_servers(data_dir, certs_file, key_file, workflow, party_key)
 
 
 if __name__ == "__main__":
--- a/mhs/main.py
+++ b/mhs/main.py
@@ -1,6 +1,5 @@
 import pathlib
-import ssl
-from typing import Tuple, Dict
+from typing import Dict
 
 import tornado.httpserver
 import tornado.ioloop
@@ -8,7 +7,6 @@ import tornado.web
 
 import definitions
 import mhs.common.configuration.configuration_manager as configuration_manager
-import mhs.inbound.request.handler as async_request_handler
 import mhs.outbound.request.synchronous.handler as client_request_handler
 import utilities.config as config
 import utilities.file_utilities as file_utilities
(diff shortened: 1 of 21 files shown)


mhs/main.py
```diff
@@ -44,39 +43,37 @@ def load_party_key(data_dir: pathlib.Path) -> str:
...
-    :param data_dir: The directory to load interactions configuration from.
     :param certs_dir: The directory containing certificates/keys to be used to identify this MHS to a remote MHS.
     :param party_key: The party key to use to identify this MHS.
-    :return: The workflow that can be used to handle messages.
+    :return: The workflows that can be used to handle messages.
     """
-    interactions_config_file = str(data_dir / "interactions" / "interactions.json")
-    config_manager = configuration_manager.ConfigurationManager(interactions_config_file)
+    # transmission = outbound_transmission.OutboundTransmission(str(certs_dir))
+    # workflow = sync_async_workflow.SyncAsyncWorkflow(transmission, party_key)
```
Philip-Woods, line 40: Do we need these commented lines? And this function doesn't seem to be using the parameters either

mhs/main.py
```diff
@@ -44,39 +43,37 @@ def load_party_key(data_dir: pathlib.Path) -> str:
...
+    interactions_config_file = str(data_dir / "interactions" / "interactions.json")
+    config_manager = configuration_manager.ConfigurationManager(interactions_config_file)
+
     supplier_application = tornado.web.Application(
-        [(r"/(.*)", client_request_handler.SynchronousHandler,
-          dict(workflow=workflow, callbacks=callbacks, async_timeout=ASYNC_TIMEOUT))])
+        [(r"/", client_request_handler.SynchronousHandler,
+          dict(config_manager=config_manager, workflows=workflows))])
     supplier_server = tornado.httpserver.HTTPServer(supplier_application)
     supplier_server.listen(80)
```
Philip-Woods, line 60: Since this is intended to be the main file for the outbound service should we be removing references to the inbound one as well, I have done this as part of the splitting the application so just thinking it might make merging the two easier as it would just require the main file to be replaced

mhs/main.py
```diff
@@ -101,15 +98,15 @@ def main():
     certs_file, key_file = load_certs(certs_dir)
     party_key = load_party_key(certs_dir)
 
-    workflow = initialise_workflow(data_dir, certs_dir, party_key)
+    workflows = initialise_workflows(certs_dir, party_key)
 
-    start_tornado_servers(certs_file, key_file, workflow, party_key)
+    start_tornado_servers(data_dir, certs_file, key_file, workflows, party_key)
```
Philip-Woods: If the certs and party key are stored in the `data_dir` does it makes sense to pass those as well? this feels like overlapping information

mhs/mhs/common/workflow/common.py
```diff
@@ -12,6 +14,17 @@ def __init__(self, interaction_name):
...
+    async def handle_supplier_message(self, message_id: str, interaction_details: dict, payload: str) -> Tuple[int, str]:
+        """
+        Handle a message from the supplier system (or a message from an adaptor that the supplier system speaks to).
+
+        :param message_id: ID of the message to send
+        :param interaction_details: interaction details used to construct the message to send downstream
+        :param payload: payload to send downstream
+        :return: the HTTP status and body to return as a response
+        """
+        pass
```
Philip-Woods, line 20: Think it might be a bit more consistent if we make this method refer to handing the outbound message vs the supplier message

mhs/mhs/outbound/request/synchronous/handler.py
```diff
@@ -1,91 +1,108 @@
...
 import mhs.outbound.request.common as common
-from utilities import integration_adaptors_logger as log
+from mhs.common import workflow
+from mhs.common.configuration import configuration_manager
+from utilities import integration_adaptors_logger as log, message_utilities
 
 logger = log.IntegrationAdaptorsLogger('MHS_OUTBOUND_HANDLER')
 
 
 class SynchronousHandler(common.CommonOutbound, tornado.web.RequestHandler):
```
Philip-Woods: I see that the `CommonOutbound` is currently empty so it isn't creating a problem here but Dave expressed that we should attempt to avoid multiple inheritance

mhs/mhs/outbound/request/synchronous/handler.py
```diff
@@ -1,91 +1,108 @@
...
-    async def _pause_request(self, async_message_id):
-        """Pause the incoming request until an asynchronous response is received (or a timeout is hit).
-
-        :param async_message_id: The ID of the request that expects an asynchronous response.
-        :raises: An HTTPError if we timed out waiting for an asynchronous response.
-        """
+            interaction_details = self._get_interaction_details(interaction_id)
+        except common_workflow.UnknownInteractionError as e:
+            logger.warning('0007', 'Unknown {InteractionId} in request', {'InteractionId': interaction_id})
+            raise tornado.web.HTTPError(404, "Unknown interaction ID: %s", interaction_id) from e
```
Philip-Woods: Might be useful to return an actual error message as well as status as part of these error responses

mhs/mhs/outbound/request/synchronous/handler.py
```diff
@@ -1,91 +1,108 @@
...
-            logger.info('0003', 'Waiting for asynchronous response to message with {MessageId}',
-                        {'MessageId': async_message_id})
-            await self.async_response_received.wait(datetime.timedelta(seconds=self.async_timeout))
-        except TimeoutError:
-            raise tornado.web.HTTPError(log_message=f"Timed out waiting for a response to message {async_message_id}")
+            workflow = self.workflows[interaction_details['workflow']]
+        except KeyError as e:
+            logger.error('0008', "Weren't able to determine workflow for {InteractionId} . This likely is due to a "
+                                 "misconfiguration in interactions.json", {"InteractionId": interaction_id})
+            raise tornado.web.HTTPError(500, "Couldn't determine workflow to invoke for interaction ID: %s",
```
Philip-Woods: Same again here, a response message might be useful

mhs/mhs/outbound/request/synchronous/tests/test_handler.py
```diff
@@ -1,61 +1,144 @@
...
+        response = self.fetch("/", method="POST", headers={"Interaction-Id": INTERACTION_NAME}, body="")
 
-        self.assertEqual(response.code, 500)
+        self.assertEqual(response.code, 400)
+
+    @staticmethod
+    def awaitable(result):
+        future = asyncio.Future()
+        future.set_result(result)
+        return future
```
Philip-Woods: This seems to happen in a few places, wondering if it's worth moving it into the `utilities.test_utilities` at some point if it keeps cropping up?

mhs/mhs/outbound/request/synchronous/handler.py
```diff
@@ -1,91 +1,108 @@
...
+                                        interaction_id) from e
+
+        status, response = await workflow.handle_supplier_message(message_id, interaction_details, body)
+
+        self._write_response(status, response)
```
Philip-Woods, line 59: So this might be something worth double checking when Dave gets back but I recall him mentioning that we should be recording message information almost as soon as we receive it in the state store. Not sure if this wants to be done in the handler or when it enters a work flow though

mhs/mhs/outbound/request/synchronous/handler.py
```diff
@@ -1,91 +1,108 @@
...
+        self.config_manager = config_manager
 
-    async def post(self, interaction_name):
-        logger.info('0001', 'Client POST received. {Request}', {'Request': str(self.request)})
+    async def post(self):
+        message_id = self._extract_message_id()
+        self._extract_correlation_id()
 
-        message_id = self.get_query_argument("messageId", default=None)
+        logger.info('0006', 'Client POST received. {Request}', {'Request': str(self.request)})
```
Philip-Woods: Pretty minor but just thinking for consistency

```suggestion
        logger.info('0006', 'Outbound POST received. {Request}', {'Request': str(self.request)})
```

mhs/mhs/outbound/request/synchronous/handler.py
```diff
@@ -1,91 +1,108 @@
...
+        else:
+            log.correlation_id.set(correlation_id)
+            logger.info('0004', 'Found correlation id on incoming request.')
+
+    def _extract_interaction_id(self):
+        try:
+            interaction_id = self.request.headers['Interaction-Id']
+        except KeyError as e:
+            logger.warning('0005', 'Required Interaction-Id header not passed in request')
+            raise tornado.web.HTTPError(404, 'Required Interaction-Id header not found') from e
```
Philip-Woods: Don't know if I'm on the right track with this but it feels to me like this method shouldn't know/care about tornado errors. Also if this method doesn't have a tornado error it might make extracting a common module from these methods a bit simpler if we ever do implement the outbound async queue handler.

mhs/main.py
ismail-s: Reason I did this was because the next task, working on the workflow, would mean that we'd essentially bung some of this code back in again. I know, it's not amazing committing commented-out code, but it felt a bit simpler than just stripping the code out only to add it in again in a pretty similar fashion.

mhs/main.py
Philip-Woods: Given our discussion earlier about how the certs dir will be phased out in favor of secrets management I'm fine with this being left for the time being.

mhs/mhs/outbound/request/synchronous/handler.py
Philip-Woods: As discussed this will be done as soon as the message enters the workflow

Philip-Woods (approved):

PR #52, opened 2019-08-12
Feature/RT119-Split-Application

Philip-Woods:
This pull request splits the MHS application into 4 main modules:
* mhs_common: A library of common modules used by all MHS services. The queue adaptor, state adaptor/management, config and any workflow management will live here.
* Outbound service: This is an application which handles the outbound messages from the supplier to spine
* Inbound service: An application which handles inbound messages from spine (or more generally other MHSs) into the supplier
* Spine Route Lookup service: This will be extended to become a full service managing routing calls to SDS. This is a WIP to be completed with the sync-async workflow. 

A few additional things have been updated:
* The packer file has now been split in 2, one for inbound and one for outbound
* There is a top level script to build both inbound and outbound containers using the packer files
* There is a top level docker-compose.yml file to run each of the services and a local version of DynamoDB and RabbitMQ
* The pipeline has been updated accordingly to run each modules unit tests, The integration tests are currently being skipped since the sync-async workflow no longer works (This is intentional)


Jenkinsfile
```diff
@@ -8,6 +8,30 @@ pipeline {
     }
 
     stages {
+        stage('Build modules') {
+            steps{
+                dir('common'){
+                    sh label: 'Installing common dependencies', script: 'pipenv install --dev --deploy --ignore-pipfile'
+                }
```
ismail-s: Optional: extract common code into a function.

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+  dynamodb:
+    image: amazon/dynamodb-local
+    ports:
+      - "8000:8000"
+  rabbitmq:
+    build: ./common/local-env/rabbit/
+    ports:
+      - "15672:15672"
+      - "5672:5672"
+    hostname: "localhost"
```
ismail-s, line 25: This docker-compose file is the the root of this repo, but the documentation of how to use it is in `mhs/README.md`, not `README.md`. I can see why it's there as this docker-compose file currently only spins up MHS, but if this docker-compose file is intended to spin up a full local dev cluster, then:
1. SCR should also get spun up (maybe do this in some later task)
1. The documentation of using docker-compose should be in the readme at the root of the repo, possibly linking out to eg `mhs/README.MD`.

mhs/Pipfile
```diff
@@ -10,19 +10,11 @@ unittest-xml-reporting = "~=2.5"
 [packages]
 requests = "*"
 integration-adaptors-common = {editable = true,path = "./../common"}
-ldap3 = "*"
 tornado = "*"
-aioboto3 = "*"
 defusedxml = "~=0.6"
```
ismail-s, line 12: So is this Pipfile just used for the MHS integration tests? In that case, maybe all these deps should be in the dev-packages section.

mhs/README.md
```diff
@@ -60,10 +60,21 @@ When running the tests locally, you will need to set the MHS_ADDRESS and ASID in
...
-`pipenv run mhs` will run the `main.py` script (you can also run this directly). This will start up an MHS
-instance listening for 'client' requests on port 80 and asynchronous responses from Spine on port 443.
-Note that the following environment variables need to be set when running MHS:
-- `MHS_LOG_LEVEL` - log level threshold
+Since the MHS is split into multiple services the simplest way to run an instance locally is to use the docker-compose script
+found in the root directory. The following steps should be followed: 
+* Requirements: `Docker`, [`Packer`](https://www.packer.io/)
+* Run the `./build.sh` script found in the top level directory of this project. This will build the inbound and outbound  
+    containers ready for running
+* Run `docker-compose build`. This build all the additional containers
```
ismail-s:
```suggestion
* Run `docker-compose build`. This will build all the additional containers
```

mhs/README.md
```diff
@@ -60,10 +60,21 @@ When running the tests locally, you will need to set the MHS_ADDRESS and ASID in
...
+* Run `docker-compose up`. This will start the inbound and outbound services, along with an instance of RabbitMQ and 
+    DynamoDb
+* Depending on the OS the MHS is being run on, measures may have to be taken to allow public inbound traffic so that
+the async responses from Spine can access the MHS. For example on windows a inbound rule was required in windows
+firewall to allow inbound traffic through port 443. In the AWS test environment an inbound rule was added
+in the security groups for machine the MHS was deployed to
```
ismail-s: Did we actually have to do this, for the Opentest stuff IIRC I didn't have to do anything with security groups, simply because Opentest is a VPN connection. Unless you're talking about in terms of the container ports.

mhs/inbound/Pipfile
```diff
@@ -0,0 +1,26 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+coverage = "~=4.5"
+unittest-xml-reporting = "~=2.5"
+
+[packages]
+requests = "*"
```
ismail-s: Do we actually use requests in inbound? And same question for defusedxml.

mhs/inbound/request/tests/test_handler.py
```diff
@@ -8,7 +8,7 @@
 import utilities.message_utilities as message_utilities
 import utilities.xml_utilities as xml_utilities
 
-import mhs.inbound.request.handler as handler
+import request.handler as handler
```
ismail-s: I feel like it would be better to keep the current import namespace we've been using, to avoid any potential clashes/confusion. This is an example: `requests` is a popular package, and `request` here is our own code, it's not obvious which is which.

mhs/inbound/sonar-project.properties
```diff
@@ -0,0 +1,3 @@
+sonar.projectKey=mhs-reference-implementation
```
ismail-s: Did you check the Sonar reports, do they look correct? Reason why I'm asking is that before this PR, there were 3 `sonar-project.properties` files, all with different `sonar.projectKey`s.

mhs/mhs_common/definitions.py
```diff
@@ -0,0 +1,4 @@
+import os
+
+
+ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
```
ismail-s: Missing newline at end of file.

mhs/mhs_common/workflow/sync_async.py
```diff
@@ -18,8 +17,9 @@
 class SyncAsyncWorkflow(common_synchronous.CommonSynchronousWorkflow):
     """Handles the workflow for the sync-async messaging pattern."""
 
+    # TODO: This used to take an outbound transmission object
```
ismail-s: What's going on here?

mhs/outbound/Pipfile
```diff
@@ -0,0 +1,27 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+coverage = "~=4.5"
+unittest-xml-reporting = "~=2.5"
+v = {editable = true,version = "*"}
```
ismail-s: What is this line doing?

mhs/mhs_common/workflow/asynchronous_express.py
```diff
@@ -1,6 +1,6 @@
 """This module defines the asynchronous express workflow."""
 
-import mhs.common.workflow.common_asynchronous as common_asynchronous
+import workflow.common_asynchronous as common_asynchronous
```
ismail-s: Here you're doing `import workflow.blah`, in `intermediary_reliable.py` you're doing `import common.workflow.blah`

pipeline/packer/inbound.json
```diff
@@ -0,0 +1,83 @@
...
+        "WORKDIR {{user `app_dir`}}/{{user `inbound_dir`}}",
+        "ENTRYPOINT pipenv run start-inbound"
+      ]
+    }
+  ],
+  "provisioners": [
+    {
+      "type": "shell",
+      "inline": [
+        "mkdir -p {{user `app_dir`}}",
```
ismail-s:
```suggestion
```

Don't think we need this line, as `mkdir -p` can create a sequence of subdirectories. And same for the other packer file.

pipeline/packer/outbound.json
```diff
@@ -0,0 +1,83 @@
...
+        "WORKDIR {{user `app_dir`}}/{{user `outbound_dir`}}",
+        "ENTRYPOINT pipenv run start-outbound"
+      ]
+    }
+  ],
+  "provisioners": [
+    {
+      "type": "shell",
+      "inline": [
+        "mkdir -p {{user `app_dir`}}",
```
ismail-s:
```suggestion
```

mhs/spineroutelookup/Pipfile
```diff
@@ -0,0 +1,27 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
```
ismail-s, line 5: How is this going to work whilst we work on the async-express workflow, will we just import the files from here? Because then for the timebeing this folder would need to be included in the Packer outbound.json

mhs/spineroutelookup/Pipfile
```diff
@@ -0,0 +1,27 @@
...
+[dev-packages]
+coverage = "~=4.5"
+unittest-xml-reporting = "~=2.5"
+
+[packages]
+requests = "*"
```
ismail-s: Is requests being used in this (sub) package?

mhs/selenium_tests/test_scenarios/int_smoke_test.py
```diff
@@ -1,12 +1,13 @@
...
+from unittest import TestCase, skip
 
 from selenium_tests.page_objects import methods
 
 TEST_NHS_NUMBER = '9446245796'
 GP_SUMMARY_UPLOAD_INTERACTION = 'gp_summary_upload'
 
 
-class FunctionalTest(TestCase):
+@skip("Sync Async workflow not implemented yet")
```
ismail-s, line 10: Minor/optional: I did this in #51 as an `expectedFailure` (so that the request would still get made), though that doesn't allow you to say a reason, which `skip` does.

pipeline/terraform/test-environment/test-environment.tf
```diff
@@ -97,19 +145,25 @@ resource "aws_ecs_task_definition" "test-environment-scr-service-task" {
   ]
   )
   cpu = "128"
-  memory = "256"
+  memory = "128"
```
ismail-s, line 148: I'm guessing you had to do this to get everything to fit onto one box.

pipeline/terraform/test-environment/test-environment.tf
```diff
@@ -97,19 +145,25 @@ resource "aws_ecs_task_definition" "test-environment-scr-service-task" {
...
+  cluster = var.cluster_id
+  task_definition = aws_ecs_task_definition.test-environment-mhs-outbound-task.arn
+  desired_count = 1
+  launch_type = "EC2"
+}
 
-
-resource "aws_ecs_service" "test-mhs-environment-service" {
-  name = "${var.build_id}-service"
+resource "aws_ecs_service" "test-mhs-inbound-environment-service" {
```
ismail-s: Minor/optional: Can you have both this service and `test-outbound-mhs-environment-service` with more similarly formatted names.

mhs/mhs_common/Pipfile
```diff
@@ -0,0 +1,25 @@
...
+[dev-packages]
+coverage = "~=4.5"
+tornado = "*"
+unittest-xml-reporting = "~=2.5"
+
+[packages]
+mhs-common = {editable = true,path = "."}
+defusedxml = "~=0.6"
+integration-adaptors-common = {editable = true,path = "./../../common"}
+aioboto3 = "*"
```
ismail-s: How do the deps play between here and `setup.py`, does there have to be duplication (possibly useful to read [this](https://pipenv.readthedocs.io/en/latest/advanced/#pipfile-vs-setup-py))? Also, do you need `mhs-common` (ie this package) to be mentioned in this packages section?

mhs/outbound/main.py
```diff
@@ -1,22 +1,23 @@
...
+import configuration.configuration_manager as configuration_manager
+import workflow.sync_async as sync_async_workflow
+
+import request.synchronous.handler as client_request_handler
+import transmission.outbound_transmission as outbound_transmission
 import utilities.config as config
 import utilities.file_utilities as file_utilities
+import utilities.integration_adaptors_logger as log
+
+logger = log.IntegrationAdaptorsLogger('OUTBOUND_MAIN')
```
ismail-s, line 16: I think the logger component should take the name of the thing (component?) that is running eg MHS_OUTBOUND, MHS_INBOUND, SCR_ADAPTOR etc so that we can use more generic names in places like this and have the logs still saying which thing is the cause. This would also tie in with the common logs, showing what caused them.

Not necessarily something to do in this PR/task, but if not, not sure where else to do it, unless we create a separate task to do this.

pipeline/packer/outbound.json
```diff
@@ -0,0 +1,83 @@
...
+      },
+      {
+        "name": "publish",
+        "type": "docker-push",
+        "ecr_login": true,
+        "login_server": "{{user `docker_registry_url`}}"
+      },
+      {
+        "type": "docker-tag",
+        "repository": "local/outbound"
```
ismail-s: Might be better to tag as something like `local/mhs-outbound`

mhs/README.md
```diff
@@ -60,10 +60,21 @@ When running the tests locally, you will need to set the MHS_ADDRESS and ASID in
...
+* Note that when running the MHS the environment variable `MHS_LOG_LEVEL` is required to be set to one of: `NOTSET`,
+`INFO`, `WARNING`, `ERROR` or `CRITICAL`. Where `NOTSET` displays the most logs and `CRITICAL` displays the least.
+
 
 Any content POSTed to `/path` (for example) on port 80 will result in the request configuration for the `path` entry in
 `data/interactions.json` being loaded and the content sent as the body of the request to Spine. Adding entries to
```
ismail-s, line 65: The last line of this file doesn't make any sense, it references steps that aren't there.

ismail-s (changes requested):

docker-compose.yml
Philip-Woods: I've moved the running section to the top level `README.md`, and changed the wording a bit to describe starting a development cluster since as you say, it will also be used for the SCR adaptor.

mhs/Pipfile
Philip-Woods: This Pipfile is used just for the integration tests now yes, There was an issue with imports as well after merging in develop so I've moved the pipfile to the `selenium_tests` module as well since nothing else used it. 

As suggested I updated the packages to be dev packages as well

mhs/README.md
Philip-Woods: Think you have to enable the inbound rules for that specific instance even if there is a vpn connection, otherwise spine is blocked sending the async response. There is a rule on the build box to allow this traffic through

mhs/inbound/Pipfile
Philip-Woods: Did a quick search and no, it doesn't look like it so I've removed them and updated the pipfile.lock

mhs/mhs_common/workflow/sync_async.py
Philip-Woods: As discussed this workflow was reliant on receiving an outbound transmitter, since this was defined in the `Outbound` package it felt odd to be importing the outbound module into the common. This has now been updated so there is a transmission interface in the `integration_adaptors_common.comms` package, the `OutboundTransmission` class implements this and the `SyncAsyncWorkflow` now expects some implementation of this transmission interface

mhs/outbound/Pipfile
Philip-Woods: Honestly no idea, I've removed it and everything installs as expected and the tests pass, I'm going to guess it was a copy/paste mistake somehow

mhs/mhs_common/workflow/asynchronous_express.py
Philip-Woods: Should all be using the same paths now

mhs/spineroutelookup/Pipfile
Philip-Woods: For the time being I wasn't really planning on using any of the SDS stuff before it has it's own service, I figured we could probably do something similar to how it's currently done with fixed values from the interactions.json. I also don't think this work flow accounted for using anything from SDS

mhs/spineroutelookup/Pipfile
Philip-Woods: Doesn't look like it, and I don't think it will be in the future, will remove it

mhs/selenium_tests/test_scenarios/int_smoke_test.py
Philip-Woods: Yeah I noticed that, personally I think being able to provide a reason is useful but as this will be changed in the next sprint it not too fussed

pipeline/terraform/test-environment/test-environment.tf
Philip-Woods: Yep, total resources for the box are 1024 CPU and 985 memory, though while testing it, it seemed to need some spare

mhs/outbound/main.py
Philip-Woods: This seems like a good idea, sounds like something that would be changed in the config method of the logging. There is already a ticket for `logging` enhancements, maybe it makes sense to include it with that?

pipeline/packer/outbound.json
Philip-Woods: Updated this and inbound as well

mhs/README.md
Philip-Woods: Binned it

common/comms/transmission_adaptor.py
```diff
@@ -0,0 +1,9 @@
+import abc
+
+
+class TransmissionAdaptor(abc.ABC):
+
+    def make_request(self,  interaction_details, message):
+        """Given a set of interaction details and a message, the message is propogated to the appropriate
```
ismail-s:
```suggestion
        """Given a set of interaction details and a message, the message is propagated to the appropriate
```

common/comms/transmission_adaptor.py
```diff
@@ -0,0 +1,9 @@
+import abc
+
+
+class TransmissionAdaptor(abc.ABC):
+
+    def make_request(self,  interaction_details, message):
```
ismail-s: Shouldn't this be async def, or is that once @ZameerParvez has done his task (probably)? And shouldn't this method return some response? And this method probably should have a `abstractmethod` decorator.

mhs/common/mhs_common/messages/soap_envelope.py
```diff
@@ -1,6 +1,6 @@
 """This module defines the envelope used to wrap synchronous messages to be sent to a remote MHS."""
 
-import messages.envelope as envelope
+from mhs_common import messages as envelope
```
ismail-s: Is this correct?

mhs/common/setup.py
```diff
@@ -3,7 +3,7 @@
 import setuptools
 
 setup(
-    name='mhs_common',
+    name='common',
```
ismail-s: I think the previous name (`mhs_common`) was better. Or `mhs-common` would be more consistent with the other Pipfile usages.

mhs/mhs_common/Pipfile
Philip-Woods: From what I understand since we're running all the unit tests through a pip environment the packages need duplicate requirements since they're tested with pipenv and exported with the setup.py requirements. 

Also after removing `mhs-common` the tests all seem to run fine so I guess we didn't need the self-dependancy

mhs/selenium_tests/Pipfile
```diff
@@ -6,13 +6,13 @@ verify_ssl = true
...
 unittest-xml-reporting = "~=2.5"
-
-[packages]
 requests = "*"
-integration-adaptors-common = {editable = true,path = "./../common"}
+integration-adaptors-common = {editable = true,path = "./../../common"}
 tornado = "*"
 defusedxml = "~=0.6"
 
+[packages]
```
ismail-s: Do you even need this line if this section doesn't exist?

mhs/outbound/main.py
ismail-s: Can do.

ismail-s (dismissed):

ismail-s (approved):

PR #53, opened 2019-08-12
Updated integration test template to populate timestamps correctly

Philip-Woods:
The current template doesn't load the correct time stamps so this is an update to populate the template with the correct time stamp


mhs/selenium_tests/data/templates/xml_IN150016UK05.mustache
```diff
@@ -21,7 +22,7 @@
         <author1 typeCode="AUT">
             <AgentSystemSDS classCode="AGNT">
                 <agentSystemSDS classCode="DEV" determinerCode="INSTANCE">
-                    <id root="1.2.826.0.1285.0.2.0.107" extension="{{asid}}"/>
+                    <id root="1.2.826.0.1285.0.2.0.107" extension="9{{asid}}"/>
```
PhilSmithBJSS, line 25: Why the '9'?
The asid is input from the configuration, and shouldn't need amending...

Philip-Woods: @Nayira Any idea where this came from/if it's needed?

Philip-Woods:
Closed as it is expected to be updated in #55

PR #56, opened 2019-08-15
Feature/RT153: Inbound Message Handler

Philip-Woods:


common/utilities/integration_adaptors_logger.py
```diff
@@ -69,6 +69,9 @@ def _format_and_write(self, message, values, process_key_num, level):
             message += f' RequestId={message_id.get()}'
         if correlation_id.get():
             message += f' CorrelationId={correlation_id.get()}'
+        if inbound_message_id.get():
+            message += f'InboundMessageId={inbound_message_id.get()}'
+
```
Philip-Woods, line 74: The logging strategy I went with here (which I think is what we discussed) was that the message ref (which is the message id of the outbound message received by spine that causes an inbound response) will be referred to as the `MessageId` in the logs, as to align with the outbound message (which should make it easier to search in logs). Then the `MessageID` of the inbound response will be referred to as the `InboundMessageId`. Might be worth running this by Dave when he's back.

common/utilities/integration_adaptors_logger.py
```diff
@@ -69,6 +69,9 @@ def _format_and_write(self, message, values, process_key_num, level):
             message += f' RequestId={message_id.get()}'
         if correlation_id.get():
             message += f' CorrelationId={correlation_id.get()}'
+        if inbound_message_id.get():
+            message += f'InboundMessageId={inbound_message_id.get()}'
```
ismail-s:
```suggestion
            message += f' InboundMessageId={inbound_message_id.get()}'
```

ismail-s: Can one of the tests be updated to test this?

mhs/common/mhs_common/state/tests/test_work_description.py
```diff
@@ -22,7 +23,8 @@
         wd.VERSION_KEY: 0,
         wd.CREATED_TIMESTAMP: '11:59',
         wd.LATEST_TIMESTAMP: '12:00',
-        wd.STATUS: wd.MessageStatus.IN_OUTBOUND_WORKFLOW
+        wd.STATUS: wd.MessageStatus.IN_OUTBOUND_WORKFLOW,
+        wd.WORKFLOW: 'sync'
```
ismail-s: Minor: can extract common string as a constant.

mhs/common/mhs_common/workflow/__init__.py
```diff
@@ -7,6 +7,11 @@
 from mhs_common.workflow.intermediary_reliable import IntermediaryReliableWorkflow
 from mhs_common.workflow.synchronous import SynchronousWorkflow
 
+ASYNC_EXPRESS = 'async-express'
```
ismail-s, line 10: v. minor/optional: could combine these into an enum like:
```python
class WorkflowType(str, enum.Enum):
    ASYNC_EXPRESS = 'async-express'
    ...
```
then in `work_description.py`, could use the type `WorkflowType` instead of `str`. Though then that might mean a circular import, which would then mean moving the enum into it's own file.

mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -10,3 +11,6 @@ class AsynchronousExpressWorkflow(common_asynchronous.CommonAsynchronousWorkflow
     async def handle_outbound_message(self, message_id: str, interaction_details: dict, payload: str) \
             -> Tuple[int, str]:
         raise NotImplementedError()
+
+    async def handle_inbound_message(self, work_description: wd.WorkDescription, payload: str):
```
ismail-s, line 15: It will be interesting to see what happens in the inbound workflow task. The outbound and inbound sides feel a little like separate things with their own dependencies/interfaces, and atm we've got them both in one class. Which I don't think is necessarily going to be a problem, but we may end up having a bunch of optional dependencies and then having the outbound/inbound MHS passing in the dependencies they need.

EDIT: you've mentioned this yourself in https://github.com/nhsconnect/integration-adaptors/pull/57#discussion_r314388707.

ismail-s, line 15: I think this should return a `str` for the ack back to Spine.

mhs/common/mhs_common/workflow/common_asynchronous.py
```diff
@@ -1,6 +1,8 @@
 """This module defines the common base for all asynchronous workflows."""
 
 import mhs_common.workflow.common as common
+import abc
+from mhs_common.state import work_description as wd
```
ismail-s: Unused imports

mhs/inbound/main.py
```diff
@@ -74,7 +94,10 @@ def main():
     certs_file, key_file = load_certs(certs_dir)
     party_key = load_party_key(certs_dir)
 
-    start_inbound_server(certs_file, key_file, party_key)
+    workflows = initialise_workflows(certs_dir, party_key)
+    store = dynamo_persistence_adaptor.DynamoPersistenceAdaptor(table_name=config.get_config('STATE_TABLE_NAME'))
```
ismail-s, line 98: Whilst the Jenkins build is passing, the inbound MHS isn't starting up, because this environment variable isn't being set in the pipeline. I have done the outbound side in #57, I think you could do the inbound side in this PR.

It's because we don't yet have integration tests for the async express workflow that the Jenkins build pipeline is passing currently.

This comment is related to https://github.com/nhsconnect/integration-adaptors/pull/57#discussion_r314377278.

mhs/inbound/inbound/request/handler.py
```diff
@@ -1,46 +1,70 @@
...
-        self.callbacks = callbacks
+        self.workflows = workflows
         self.party_id = party_id
+        self.state_store = state_store
 
-    def post(self):
-        logging.debug("POST received: %s", self.request)
-        logging.debug("Body: %s", self.request.body)
+    async def post(self):
+        logger.info('001', 'Inbound POST recieved: {request}', {'request': self.request})
```
ismail-s:
```suggestion
        logger.info('001', 'Inbound POST received: {request}', {'request': self.request})
```

mhs/inbound/inbound/request/handler.py
```diff
@@ -1,46 +1,70 @@
...
 import mhs_common.messages.ebxml_ack_envelope as ebxml_ack_envelope
 import mhs_common.messages.ebxml_envelope as ebxml_envelope
 import mhs_common.messages.ebxml_request_envelope as ebxml_request_envelope
+from utilities import message_utilities
+import mhs_common.workflow as workflow
 
+logger = log.IntegrationAdaptorsLogger('INBOUND_HANDLER')
 
 
 class InboundHandler(tornado.web.RequestHandler):
```
ismail-s, line 20: Can you stick a timing decorator on this handler. I'll get the decorator to log using the logging component in my PR (unless you wish to do it)

mhs/inbound/inbound/request/handler.py
```diff
@@ -53,7 +77,48 @@ def _send_ack(self, parsed_message: ebxml_envelope.EbxmlEnvelope):
...
+        :return: the message id the inbound message is a response to
+        """
+        try:
+            message_id = message.message_dictionary[ebxml_envelope.RECEIVED_MESSAGE_ID]
+            log.message_id.set(message_id)
+            logger.info('0008', 'Found message id on inbound message.')
+            return message_id
+        except KeyError:
+            logger.info('0007', "No Message reference found ")
+            return None
```
ismail-s: If None is returned, then what happens next, what code errors out, do we get a decent-enough error?

mhs/inbound/inbound/request/handler.py
```diff
@@ -53,7 +77,48 @@ def _send_ack(self, parsed_message: ebxml_envelope.EbxmlEnvelope):
...
+        :param message:
+        :return: the message id the inbound message is a response to
+        """
+        try:
+            message_id = message.message_dictionary[ebxml_envelope.RECEIVED_MESSAGE_ID]
+            log.message_id.set(message_id)
+            logger.info('0008', 'Found message id on inbound message.')
+            return message_id
+        except KeyError:
+            logger.info('0007', "No Message reference found ")
```
ismail-s:
```suggestion
            logger.info('0007', "No Message reference found")
```

mhs/inbound/inbound/request/handler.py
```diff
@@ -1,46 +1,70 @@
...
 
         request_message = ebxml_request_envelope.EbxmlRequestEnvelope.from_string(self.request.headers,
                                                                                   self.request.body.decode())
-        ref_to_id = request_message.message_dictionary[ebxml_envelope.RECEIVED_MESSAGE_ID]
-        logging.debug("Message received is in reference to '%s'", ref_to_id)
-
-        if ref_to_id in self.callbacks:
+        ref_to_message_id = self.extract_ref_message(request_message)
+
+        logger.info('002', 'Message received has reference to {messageID}', {'messageID': ref_to_message_id})
```
ismail-s: Not against this being here, but it feels redundant.

mhs/inbound/inbound/request/handler.py
```diff
@@ -1,46 +1,70 @@
...
+
+        try:
+            work_description = await wd.get_work_description_from_store(self.state_store, ref_to_message_id)
+        except wd.EmptyWorkDescriptionError as e:
+            logger.warning('003', 'No work description found in state store with {messageId}',
+                           {'messageId': ref_to_message_id})
+            raise tornado.web.HTTPError(500, 'No work description in state store, unsolicited message '
+                                             'received from spine',
+                                        reason="Unknown message reference") from e
```
ismail-s, line 60: Ok, so right now this is an error, but later on, when we work on forward reliable, then IIUC this might no longer be an error scenario?

mhs/inbound/inbound/request/handler.py
```diff
@@ -1,46 +1,70 @@
...
+
+        message_workflow = self.workflows[work_description.workflow]
+        logger.info('004', 'Retrieved word description from state store, forwarding message to {workflow}',
```
ismail-s:
```suggestion
        logger.info('004', 'Retrieved work description from state store, forwarding message to {workflow}',
```

mhs/inbound/inbound/request/handler.py
```diff
@@ -53,7 +77,48 @@ def _send_ack(self, parsed_message: ebxml_envelope.EbxmlEnvelope):
         }
 
         ack_message = ebxml_ack_envelope.EbxmlAckEnvelope(ack_context)
-        _, _, serialized_message = ack_message.serialize()
-
-        self.set_header("Content-Type", "text/xml")
+        message_id, http_headers, serialized_message = ack_message.serialize()
+        print(http_headers)
```
ismail-s:
```suggestion
```

mhs/inbound/inbound/request/handler.py
```diff
@@ -53,7 +77,48 @@ def _send_ack(self, parsed_message: ebxml_envelope.EbxmlEnvelope):
...
+        for k, v in http_headers.items():
+            self.set_header(k, v)
+            
         self.write(serialized_message)
+
+    def _extract_correlation_id(self, message):
```
ismail-s, line 95: Don't see this method being called from anywhere.

mhs/inbound/inbound/request/handler.py
```diff
@@ -53,7 +77,48 @@ def _send_ack(self, parsed_message: ebxml_envelope.EbxmlEnvelope):
...
+        """
+        Extracts the message id of the inbound message, this is to be logg
+        :param message:
+        :return:
+        """
+        message_id = message.message_dictionary[ebxml_envelope.MESSAGE_ID]
+        if not message_id:
+            logger.info('008', "Didn't receive message id in inbound message")
+        else:
+            log.message_id.set(message_id)
```
ismail-s:
```suggestion
            log.inbound_message_id.set(message_id)
```

mhs/inbound/inbound/request/handler.py
```diff
@@ -53,7 +77,48 @@ def _send_ack(self, parsed_message: ebxml_envelope.EbxmlEnvelope):
...
+        correlation_id = message.message_dictionary[ebxml_envelope.CONVERSATION_ID]
+        if not correlation_id:
+            correlation_id = message_utilities.MessageUtilities.get_uuid()
+            log.correlation_id.set(correlation_id)
+            logger.info('006', "Didn't receive conversation id from inbound request , so have generated a new one.")
+        else:
+            log.correlation_id.set(correlation_id)
+            logger.info('007', 'Found correlation id on inbound request.')
+
+    def _extract_message_id(self, message):
```
ismail-s, line 100: This method is not being called at the moment.

mhs/inbound/inbound/request/tests/test_handler.py
```diff
@@ -26,37 +55,87 @@ class TestInboundHandler(tornado.testing.AsyncHTTPTestCase):
...
+    def test_receive_ack(self, mock_get_uuid, mock_get_timestamp):
+        mock_get_uuid.return_value = "5BB171D4-53B2-4986-90CF-428BE6D157F5"
+        mock_get_timestamp.return_value = "2012-03-15T06:51:08Z"
+        expected_ack_response = file_utilities.FileUtilities.get_file_string(
+            str(self.message_dir / EXPECTED_RESPONSE_FILE))
+        request_body = file_utilities.FileUtilities.get_file_string(str(self.message_dir / REQUEST_FILE))
+
+        ack_response = self.fetch("/", method="POST", body=request_body, headers=CONTENT_TYPE_HEADERS)
+
+        print(ack_response.headers)
```
ismail-s:
```suggestion
```

mhs/inbound/inbound/request/tests/test_handler.py
```diff
@@ -26,37 +55,87 @@ class TestInboundHandler(tornado.testing.AsyncHTTPTestCase):
     message_dir = pathlib.Path(current_dir) / MESSAGES_DIR
 
     def setUp(self):
-        self.callbacks = {}
+        self.state = unittest.mock.MagicMock()
+        self.state.get = unittest.mock.MagicMock(side_effect=state_return_values)
```
ismail-s: I think (but double-check this) that this can be simplified to:
```suggestion
        self.state.get.side_effect=state_return_values
```

mhs/inbound/inbound/request/tests/test_handler.py
```diff
@@ -26,37 +55,87 @@ class TestInboundHandler(tornado.testing.AsyncHTTPTestCase):
...
+        self.mocked_workflows = {
+            workflow.ASYNC_EXPRESS: mock_workflow
+        }
+
         super().setUp()
 
     def get_app(self):
         return tornado.web.Application([
-            (r".*", handler.InboundHandler, dict(callbacks=self.callbacks, party_id=FROM_PARTY_ID))
+            (r".*", handler.InboundHandler, dict(workflows=self.mocked_workflows, state_store=self.state, party_id=FROM_PARTY_ID))
```
ismail-s, line 74: Note that the path regex here isn't the same as in `main.py`, which is `/.*`, but both match the same stuff.

mhs/inbound/inbound/request/tests/test_handler.py
```diff
@@ -26,37 +55,87 @@ class TestInboundHandler(tornado.testing.AsyncHTTPTestCase):
...
+
+class TestInboundWorkflow(tornado.testing.AsyncHTTPSTestCase):
+    """A simple integration test for the async response endpoint."""
+
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    message_dir = pathlib.Path(current_dir) / MESSAGES_DIR
+
+    def setUp(self):
+        self.state = unittest.mock.MagicMock()
+        self.state.get = unittest.mock.MagicMock(side_effect=state_return_values)
```
ismail-s: https://github.com/nhsconnect/integration-adaptors/pull/56#discussion_r314631102:
```suggestion
        self.state.get.side_effect=state_return_values
```

mhs/inbound/inbound/request/tests/test_handler.py
```diff
@@ -26,37 +55,87 @@ class TestInboundHandler(tornado.testing.AsyncHTTPTestCase):
...
+        self.mocked_workflows[workflow.ASYNC_EXPRESS].handle_inbound_message.assert_called()
+
+        self.assertEqual(ack_response.code, 200)
+
+
+class TestInboundWorkflow(tornado.testing.AsyncHTTPSTestCase):
+    """A simple integration test for the async response endpoint."""
+
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    message_dir = pathlib.Path(current_dir) / MESSAGES_DIR
```
ismail-s, line 147: You might be able to do something like:
```suggestion
    message_dir = pathlib.Path(__file__) / '..' / MESSAGES_DIR
```
based on [this SO answer](https://stackoverflow.com/a/50098973)

mhs/inbound/inbound/request/tests/test_handler.py
```diff
@@ -26,37 +55,87 @@ class TestInboundHandler(tornado.testing.AsyncHTTPTestCase):
...
+
+    def get_app(self):
+        return tornado.web.Application([
+            (r".*", handler.InboundHandler,
+             dict(workflows=workflow.get_workflow_map(), state_store=self.state, party_id=FROM_PARTY_ID))
+        ])
+
+    @unittest.mock.patch.object(message_utilities.MessageUtilities, "get_timestamp")
+    @unittest.mock.patch.object(message_utilities.MessageUtilities, "get_uuid")
+    def test_receive_ack(self, mock_get_uuid, mock_get_timestamp):
```
ismail-s, line 162: What's different about this test vs the similar test above?

ismail-s (changes requested):

mhs/common/mhs_common/state/tests/test_work_description.py
Philip-Woods: Updated to use the strings from `workflow`

mhs/common/mhs_common/workflow/asynchronous_express.py
Philip-Woods: I don't think the ack message actually contains an informational tag. Looking at the ack template it just contains the message id's to associate the ack with the inbound message.

ismail-s: I think when I wrote the comment I was imagining that the workflow would construct the ack response and then pass it back to the handler to return to Spine. What you're doing is having the handler construct the ack response instead. Which I'm not so against if the workflows are all returning the same sort of ack back to Spine (in the outbound handler, this isn't the case, so I expect the workflow to determine the response to return back to the supplier system.

mhs/inbound/inbound/request/handler.py
Philip-Woods: At first I thought the work description would handle this but I've updated it to do a null check on the result of `extract_ref_message` and log appropriately.

mhs/inbound/inbound/request/handler.py
Philip-Woods: Binned it then if it's not useful

mhs/inbound/inbound/request/handler.py
Philip-Woods: I'm not sure about the specific workflow that will change this, but yes your understanding is correct, eventually this will be updated to support unsolicited inbound messages

ismail-s:
Also, double-check the code coverage of the handler, it looks like there's 2 error scenarios that aren't being tested.

mhs/inbound/inbound/request/tests/test_handler.py
Philip-Woods: Works as expected

mhs/inbound/inbound/request/tests/test_handler.py
Philip-Woods: I'll update it just for consistency

mhs/inbound/inbound/request/tests/test_handler.py
Philip-Woods: I can't remember exactly but the `get_workflow_map` was failing to instantiate due to the parent methods at some point, and I figured mocking the method was avoiding this method and currently there's no calling test methods to this workflow to check the instantiation doesn't fail. Perhaps not the best place to do be making this check though...

mhs/inbound/inbound/request/tests/test_handler.py
Philip-Woods: This failed on the pipeline actually so I'll leave it for the time being as it passed previously

mhs/common/mhs_common/workflow/asynchronous_express.py
Philip-Woods: Yeah I think the inbound handler will pretty much always return the same ack to spine, we're not responding with errors to spine since it doesn't process them, it simply retries. And as mentioned looking over the ack response there's nothing that jumps out that should be variable based on the workflow

mhs/inbound/inbound/request/handler.py
Philip-Woods: Actually this no longer applies, whilst adding tests I've found the message parser will throw an ebxml parsing exception if it can't find the particular tags, so I've been able to move the exception handling to the message parsing instead. And any `_extract_` methods are safe to assume that the value will be in that dictionary otherwise an error would have occurred earlier

mhs/inbound/inbound/request/tests/test_handler.py
ismail-s: So then, when the integration tests are written, would they be testing this enough to remove this test?

Philip-Woods: I would say so yes, or more likely when the workflow tests become further developed

mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -12,5 +12,5 @@ class AsynchronousExpressWorkflow(common_asynchronous.CommonAsynchronousWorkflow
             -> Tuple[int, str]:
         raise NotImplementedError()
 
-    async def handle_inbound_message(self, work_description: wd.WorkDescription, payload: str):
+    async def handle_inbound_message(self, work_description: wd.WorkDescription, payload: str) -> str:
```
ismail-s:
```suggestion
    async def handle_inbound_message(self, work_description: wd.WorkDescription, payload: str):
```

mhs/inbound/inbound/request/tests/test_handler.py
```diff
@@ -106,6 +107,32 @@ def test_correct_workflow(self):
...
+        self.mock_workflow.handle_inbound_message.side_effect = Exception("what a failure")
+        mock_get_uuid.return_value = "5BB171D4-53B2-4986-90CF-428BE6D157F5"
+        mock_get_timestamp.return_value = "2012-03-15T06:51:08Z"
+        request_body = file_utilities.FileUtilities.get_file_string(str(self.message_dir / REQUEST_FILE))
+
+        response = self.fetch("/", method="POST", body=request_body, headers=CONTENT_TYPE_HEADERS)
+
+        self.assertEqual(response.code, 500)
+        expected = b'<html><title>500: Exception in workflow</title><body>500: ' \
+                   b'Exception in workflow</body></html>'.decode('utf-8')
```
ismail-s: There's no point hardcoding a bytestring that you then decode when you can just do a normal string in the first place.

mhs/inbound/inbound/request/handler.py
```diff
@@ -33,14 +34,21 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
         self.party_id = party_id
         self.state_store = state_store
 
+    @time_request
     async def post(self):
-        logger.info('001', 'Inbound POST recieved: {request}', {'request': self.request})
+        logger.info('001', 'Inbound POST received: {request}', {'request': self.request})
+        request_message = None
```
ismail-s: I don't think you need this line

mhs/inbound/inbound/request/handler.py
```diff
@@ -33,14 +34,21 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
...
+        try:
+            request_message = ebxml_request_envelope.EbxmlRequestEnvelope.from_string(self.request.headers,
+                                                                                      self.request.body.decode())
+        except ebxml_envelope.EbXmlParsingError as e:
+            logger.error('020', f'Failed to parse response: {e}')
```
ismail-s: Rather than using a f-string, can use the key=value support you added to the logger

mhs/inbound/inbound/request/handler.py
```diff
@@ -33,14 +34,21 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
...
+            raise tornado.web.HTTPError(500, 'Error occurred during message parsing',
+                                        reason=f'Exception in workflow') from e
```
ismail-s: Saying _Exception in workflow_ is a little misleading here if someone's diagnosing the error.

mhs/inbound/inbound/request/handler.py
```diff
@@ -101,24 +109,17 @@ def _extract_message_id(self, message):
...
-        try:
-            message_id = message.message_dictionary[ebxml_envelope.RECEIVED_MESSAGE_ID]
-            log.message_id.set(message_id)
-            logger.info('0008', 'Found message id on inbound message.')
-            return message_id
-        except KeyError:
-            logger.info('0007', "No Message reference found ")
-            return None
+        message_id = message.message_dictionary[ebxml_envelope.RECEIVED_MESSAGE_ID]
+        log.message_id.set(message_id)
```
ismail-s, line 117: I'm not sure you're testing that any of these context variables are being set (see outbound handler tests for how I had to test this, I had to mock out the context variables as I think Tornado stops the event loop during the test, after you've called `self.fetch`.

ismail-s (commented):
Missing some sort of test for correlation/conversation id not being present (or at least, you've got code handling this which maybe can be removed).

mhs/inbound/inbound/request/handler.py
Philip-Woods: Pycharm complained at first saying request message might not have been assigned before it was used, doesn't seem to care now so I'll remove it

mhs/inbound/inbound/request/handler.py
Philip-Woods: Should now be updated to check that they're set, as discussed the cases where these variables have been removed under the assumption they should always be there, and if they aren't the message parser will error first

mhs/inbound/inbound/request/tests/test_handler.py
```diff
@@ -107,32 +109,37 @@ def test_correct_workflow(self):
...
+    def test_no_reference_to_id(self):
         request_body = file_utilities.FileUtilities.get_file_string(str(self.message_dir / NO_REF_FILE))
 
         ack_response = self.fetch("/", method="POST", body=request_body, headers=CONTENT_TYPE_HEADERS)
 
         self.assertEqual(ack_response.code, 500)
 
+    def test_logging_context_variables_are_set(self):
+        log.correlation_id = unittest.mock.MagicMock()
+        log.inbound_message_id = unittest.mock.MagicMock()
```
ismail-s: Modify these variables using the `patch` decorator, so that they get reset afterwards, so that there's no risk of any other tests being affected.

ismail-s (changes requested):

ismail-s (approved):

PR #57, opened 2019-08-15
RT-151 outbound workflow

ismail-s:
- Implement outbound side of async-express workflow
  - Note that connecting to the SDS component for routing is not part of this task
- Set content type header to `text/plain` when returning error responses

TODO list
- [x] Think more about error HTTP statuses returned in async-express outbound workflow and whether I need to change any of them
- [x] Time execution of key components
- [x] Combine #56 to store workflow type in state table
- [x] Maybe (probably) test that the AUDIT log message is output
- [x] Simplify `interactions.json`, determine what needs to be there, what the different parts all mean, make sure there's an interaction in it for async express


docker-compose.yml
```diff
@@ -13,6 +13,7 @@ services:
       - "80:80"
     environment:
       - MHS_LOG_LEVEL=NOTSET
+      - MHS_STATE_TABLE_NAME=mhs_state
```
Philip-Woods, line 16: Think this will want adding to the inbound service as well

common/utilities/timing.py
```diff
@@ -90,3 +91,6 @@ def method_wrapper(*args, **kwargs):
 
     return method_wrapper
 
+def get_time() -> str:
+    """Returns UTC time in the appropriate format """
+    return datetime.datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S.%f')
```
Philip-Woods: Assuming this wants to be the same as what we're logging so it will want the postfix `Z` to indicate UTC

mhs/README.md
```diff
@@ -23,12 +23,20 @@ pipenv install
...
 To prepare a development environment for this application, ensure you have [Pipenv](https://docs.pipenv.org/en/latest/)
-installed and on your path, then from this directory run:
+installed and on your path, then within each of the subfolders `common`, `inbound` and `outbound` in this directory, run:
 ```
 pipenv install --dev
 ```
 
-### Running Unit Tests
+## Running MHS
+MHS is made up of multiple components, and running them all separately can be tedious. Instead, use Docker Compose, see [here](..) for how to do this.
```
Philip-Woods: Should this have a link assigned to it? Currently just 404s

mhs/common/mhs_common/workflow/__init__.py
```diff
@@ -1,21 +1,26 @@
...
 from mhs_common.workflow.asynchronous_express import AsynchronousExpressWorkflow
 from mhs_common.workflow.asynchronous_reliable import AsynchronousReliableWorkflow
 from mhs_common.workflow.common import CommonWorkflow
 from mhs_common.workflow.intermediary_reliable import IntermediaryReliableWorkflow
 from mhs_common.workflow.synchronous import SynchronousWorkflow
 
 
-def get_workflow_map() -> Dict[str, CommonWorkflow]:
+def get_workflow_map(persistence_store: persistence_adaptor.PersistenceAdaptor,
+                     transmission: transmission_adaptor.TransmissionAdaptor,
```
Philip-Woods: Should this default to `None`? Otherwise the inbound service will have to pass a `TransmissionAdaptor`. I don't think inbound needs the persistence adaptor either since that's passed to the actual handler to determine workflow then the work flow just uses the work description.

mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -1,12 +1,72 @@
...
+            interaction_details[ebxml_envelope.FROM_PARTY_ID] = self.party_key
+            interaction_details[ebxml_envelope.CONVERSATION_ID] = correlation_id
+            _, http_headers, message = ebxml_request_envelope.EbxmlRequestEnvelope(interaction_details).serialize()
+        except Exception as e:
+            logger.warning('0002', 'Failed to serialise outbound message. {Exception}', {'Exception': e})
+            await wdo.set_status(work_description.MessageStatus.OUTBOUND_MESSAGE_PREPARATION_FAILED)
+            return 500, 'Error serialising outbound message'
+        else:
+            logger.info('0003', 'Message serialised successfully')
+            await wdo.set_status(work_description.MessageStatus.OUTBOUND_MESSAGE_PREPARED)
```
Philip-Woods: Might be neater to split this serialization into it's own method

mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -1,12 +1,72 @@
...
+
+        try:
+            logger.info('0004', 'About to make outbound request')
+            start_time = timing.get_time()
+            response = await self.transmission.make_request(interaction_details, message)
+            end_time = timing.get_time()
```
Philip-Woods, line 43: Should this section also be checking the SOAP (I think it's soap) header and making the appropriate workflow based retries? There doesn't seem to be any retry logic within this workflow

mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -1,12 +1,72 @@
...
+        if response.status_code == 202:
+            logger.audit('0007', 'Async-express workflow invoked. Message sent to Spine and acknowledgment received. '
+                                 '{RequestSentTime} {AcknowledgmentReceivedTime}',
+                         {'RequestSentTime': start_time, 'AcknowledgmentReceivedTime': end_time})
+            await wdo.set_status(work_description.MessageStatus.OUTBOUND_MESSAGE_ACKD)
+            return 202, ''
+        else:
+            logger.warning('0008', "Didn't get expected HTTP status 202 from Spine, got {HTTPStatus} instead",
+                           {'HTTPStatus': response.status_code})
+            await wdo.set_status(work_description.MessageStatus.OUTBOUND_MESSAGE_NACKD)
```
Philip-Woods: The ticket says we should be logging the ack/nack status of the response which makes me think there should be an audit log here as well?

mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -1,12 +1,72 @@
...
+        wdo = work_description.create_new_work_description(self.persistence_store, message_id,
+                                                           work_description.MessageStatus.OUTBOUND_MESSAGE_RECEIVED)
+        await wdo.publish()
+
+        try:
+            # Call message packer to serialise message
+            interaction_details[ebxml_envelope.MESSAGE_ID] = message_id
+            interaction_details[ebxml_request_envelope.MESSAGE] = payload
+            interaction_details[ebxml_envelope.FROM_PARTY_ID] = self.party_key
+            interaction_details[ebxml_envelope.CONVERSATION_ID] = correlation_id
```
Philip-Woods: Could this correlation id come from the logging context? Or would that be a bad practice to rely on?

mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -1,12 +1,72 @@
...
+            logger.audit('0007', 'Async-express workflow invoked. Message sent to Spine and acknowledgment received. '
+                                 '{RequestSentTime} {AcknowledgmentReceivedTime}',
+                         {'RequestSentTime': start_time, 'AcknowledgmentReceivedTime': end_time})
+            await wdo.set_status(work_description.MessageStatus.OUTBOUND_MESSAGE_ACKD)
+            return 202, ''
+        else:
+            logger.warning('0008', "Didn't get expected HTTP status 202 from Spine, got {HTTPStatus} instead",
+                           {'HTTPStatus': response.status_code})
+            await wdo.set_status(work_description.MessageStatus.OUTBOUND_MESSAGE_NACKD)
+            return 500, "Didn't get expected success response from Spine"
```
Philip-Woods, line 65: This exception handling is fairly verbose so it might be a bit more readable to break it up a bit, or at least move the general exception handling to it's own method just to cut this main method up a bit

mhs/common/mhs_common/workflow/tests/test_asynchronous_express.py
```diff
@@ -0,0 +1,153 @@
...
+        self.mock_ebxml_request_envelope = patcher.start()
+        self.addCleanup(patcher.stop)
+
+        self.workflow = async_express.AsynchronousExpressWorkflow(self.mock_persistence_store,
+                                                                  self.mock_transmission_adaptor, PARTY_KEY)
+
+    def tearDown(self):
+        self.mock_create_new_work_description.assert_called_once_with(self.mock_persistence_store, MESSAGE_ID,
+                                                                      MessageStatus.OUTBOUND_MESSAGE_RECEIVED)
+        self.mock_ebxml_request_envelope.assert_called_once()
```
Philip-Woods: Feels a bit odd using the tear down method for assertions, and if a test is ever added that doesn't  perform all the same actions the current tests do this tear down might not apply.

Philip-Woods (commented):
Got up to the tests, will take a look at the rest later

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -99,7 +100,11 @@ def _write_response(self, status: int, message: str) -> None:
         """
         logger.info('0010', 'Returning response with {HttpStatus}', {'HttpStatus': status})
         self.set_status(status)
-        self.set_header("Content-Type", "text/xml")
+        if 400 <= status:
+            content_type = "text/plain"
+        else:
+            content_type = "text/xml"
+        self.set_header("Content-Type", content_type)
```
Philip-Woods, line 111: Is this because positive responses/acknowledgements return in plain text and our tornado errors are html?

pipeline/terraform/test-environment/test-environment.tf
```diff
@@ -88,6 +88,10 @@ resource "aws_ecs_task_definition" "test-environment-mhs-outbound-task" {
         {
           name = "MHS_LOG_LEVEL"
           value = var.mhs_log_level
+        },
+        {
+          name = "MHS_STATE_TABLE_NAME",
+          value = var.mhs_state_table_name
         }
```
Philip-Woods, line 99: ~Optional: Could add this to the inbound service as well~

I'll add this as per in mine as per your comment about the pipeline failing to stand up correctly

mhs/common/mhs_common/state/work_description.py
```diff
@@ -75,7 +79,7 @@ def create_new_work_description(persistence_store: pa.PersistenceAdaptor,
         logger.error('007', 'Failed to build new work description, status should not be null')
         raise ValueError('Expected status to not be None')
 
-    timestamp = get_time()
+    timestamp = timing.get_time()
     work_description_map = {
         DATA_KEY: key,
         DATA: {
```
Philip-Woods, line 91: Does this map/method parameters want to be updated to include the workflow or were we thinking that would be in mine?

Philip-Woods (commented):
Just a couple things on the rest of it, if I'm right about needed the workflow retry stuff in here then naturally the tests will want extending to check results from the outbound transmission

docker-compose.yml
ismail-s: I think it would be better for you to do this, see https://github.com/nhsconnect/integration-adaptors/pull/56#discussion_r314386313

mhs/common/mhs_common/state/work_description.py
ismail-s: I was thinking that it would be in your PR, and then at merge time we'd get it in the outbound workflow.

mhs/common/mhs_common/workflow/asynchronous_express.py
ismail-s: I've commented on the [JIRA ticket](https://gpitbjss.atlassian.net/browse/RT-151) that we'll need to address this next sprint as I don't have enough information to progress this now (and I wasn't sure that it was part of the requirements for this sprint).

mhs/common/mhs_common/workflow/asynchronous_express.py
ismail-s: It felt like a bad practice to me, which is why I didn't do it.

mhs/outbound/outbound/request/synchronous/handler.py
ismail-s: This is a bit confusing atm, I'm starting to think that I maybe should change the Tornado errors to be `text/plain` so that then all errors are at least consistent in terms of content type. The success responses, I presumed they'd be xml when writing the outbound handler. The acks would be (or they'd be empty), the responses might be, I'm not sure about multipart (and I'm not sure if that would be sent down the async channel or as a HTTP response, but at that time, I think I can change the code accordingly if necessary.

mhs/common/mhs_common/state/work_description.py
Philip-Woods: OK this map is now updated in dev, so I think all you should have to do is update the workflow so when it creates a new work description pass in the current workflow

ismail-s:
I think I've addressed the review comments, as well as all of the outstanding TODO comments (minus handling changes when #56 gets merged).

In 97407af, I've added the interaction we're going to be using to test the async-express workflow. I've taken this interaction code from @Nayira (blame her for any issues with it, I'm completely innocent..) and slightly tweaked it. I've tested this locally by hacking the integration tests that @PhilSmithBJSS has been working on. Doing this has meant that I've identified some issues with the code that we haven't spotted with unittests, I'll either put those fixes in this PR or in a separate PR. But hackily fixing those issues has meant that I've got a HTTP 202 back from Spine and the MHS.

ghost (approved):
I've reviewed the latest commits and am happy with them.

PR #58, opened 2019-08-19
RT-151 fix outbound workflow

ismail-s:
This PR fixes 3 bugs:
- f03c83e Pass workflow type to a function, this bug was introduced in #57. A way of spotting this type of bug in the future would be to have integration tests (which we hopefully will soon) and to use something like http://mypy-lang.org/
- dce6d56 Return Requests Response object from outbound transmission component. This is a temporary fix that I should have done in #57. The outbound transmission component is being redone anyway, so this fix will go away pretty soon.
- 63b3816 Fix bug with double-deserialisation. Basically, `json.dumps` was getting called twice on the same data, so the first time the object would get turned into a string, and the second time it would get turned into a string-inside-a-string.

### DIFF
--- a/mhs/common/mhs_common/state/tests/test_work_description.py
+++ b/mhs/common/mhs_common/state/tests/test_work_description.py
@@ -1,5 +1,4 @@
 import copy
-import json
 import unittest
 from unittest.mock import MagicMock, patch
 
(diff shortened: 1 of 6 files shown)


ghost (approved):
Looks good to me.

PR #59, opened 2019-08-19
RT-154 inbound workflow

ismail-s:
- Implement the inbound workflow for the async-express workflow.
  - [x] Send the message ID along with the payload, not just the payload

I've tested this locally by making a request that goes through MHS outbound to Spine, then Spine makes a request to MHS inbound which puts the message on the queue.


mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -87,5 +90,18 @@ def _record_outbound_audit_log(self, end_time, start_time, acknowledgment):
...
-    async def handle_inbound_message(self, work_description: wd.WorkDescription, payload: str):
-        pass
+    @timing.time_function
+    async def handle_inbound_message(self, message_id: str, correlation_id: str, work_description: wd.WorkDescription,
+                                     payload: str):
+        logger.info('0009', 'Entered async express workflow to handle inbound message')
+        await work_description.set_status(wd.MessageStatus.INBOUND_RESPONSE_RECEIVED)
+        try:
+            await self.queue_adaptor.send_async(payload, properties={'message-id': message_id,
+                                                                     'correlation-id': correlation_id})
```
ismail-s, line 101: Here I'm passing the message id & correlation id as [additional parameters](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-application-properties). The AMQP spec does have message & correlation ids as part of the [standard properties](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-properties), but it says in the spec that the message id should be unique, otherwise messages may get discarded. And I'm not sure we can say that with certainty. Then I didn't want to put the correlation id as a standard parameter and have the message id as an additional property, so I just made them both additional properties.

@gareth-allan @dshipp-bjss any thoughts on this?

ghost: If we're using the message ID (rather than the conversation ID), my understanding is that it's intended to identify an individual message, so should be unique. They're also UUIDs, so uniqueness shouldn't be an issue from that point of view. Is the question whether a retried messages will contain a new message ID?

common/comms/proton_queue_adaptor.py
```diff
@@ -35,25 +37,28 @@ def __init__(self, **kwargs) -> None:
...
+        payload = self.__construct_message(message, properties=properties)
         await tornado.ioloop.IOLoop.current() \
-            .run_in_executor(executor=None, func=lambda: self.__send(self.__construct_message(message)))
+            .run_in_executor(executor=None, func=lambda: self.__send(payload))
 
-    def send_sync(self, message: str) -> None:
+    def send_sync(self, message: str, properties: Dict[str, Any] = None) -> None:
         logger.info('009', 'Sending message synchronously.')
-        self.__send(self.__construct_message(message))
+        self.__send(self.__construct_message(message, properties=properties))
```
ghost: Optional: It would be more consistent to extract the `__construct_message` call to a `payload` variable as you've done in `send_async`.

mhs/common/mhs_common/workflow/common.py
```diff
@@ -23,11 +23,14 @@ class CommonWorkflow(abc.ABC):
...
-    async def handle_inbound_message(self, work_description: wd.WorkDescription, payload: str):
+    async def handle_inbound_message(self, message_id: str, correlation_id: str, work_description: wd.WorkDescription,
+                                     payload: str):
         """
         Handles an inbound message from an external system (typically from spine)
-        :param work_description:
-        :param payload:
-        :return:
+
+        :param message_id: ID of the message
```
ghost: It looks like you actually expect this to be the ID of the message the inbound message is in reference to (which I think makes sense if that's the ID you're using to identify the response on the queue with). This could be made clearer in the parameter name/documentation.

mhs/common/mhs_common/workflow/tests/test_asynchronous_express.py
```diff
@@ -152,6 +160,36 @@ def setUp(self):
...
+    ############################
+
+    @async_test
+    async def test_handle_inbound_message(self):
+        self.setup_mock_work_description()
+        self.mock_queue_adaptor.send_async.return_value = test_utilities.awaitable(None)
+
+        await self.workflow.handle_inbound_message(MESSAGE_ID, CORRELATION_ID, self.mock_work_description, PAYLOAD)
+
+        self.mock_queue_adaptor.send_async.assert_called_once_with(PAYLOAD, properties={'message-id': MESSAGE_ID, 'correlation-id': CORRELATION_ID})
```
ghost: Pycharm tells me this line is too long.

ghost (dismissed):

ghost (approved):

mhs/common/mhs_common/workflow/asynchronous_express.py
ismail-s: Between @briandiggle, me and @gareth-allan, we've discussed this a bit. The conclusion was to keep this as-is, it's simple enough as it is, possibly we should rename `message-id` to something else, but I'd want this to be discussed/thought through further, outside of this PR, before making any changes to this.

PR #60, opened 2019-08-21
Feature/rt 152 outbound transmission

ZameerParvez:
Add retries for outbound transmission and move where https requests are made to common comms


mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -34,21 +52,53 @@ def make_request(self, interaction_details, message):
...
+        for error_type in self.errors_not_to_retry:
+            if isinstance(e, error_type):
+                return False
+
+        return True
 
     @staticmethod
-    def _build_headers(interaction_details):
+    def _build_headers(interaction_details: Dict[str, str]):
         headers = {'type': interaction_details['type'],
```
ismail-s: Headers shouldn't be coming from the interaction details, instead get the headers from the message packer (look at the workflow, it currently is getting headers and not passing them to the outbound transmission). And then remove the headers that are in `interactions.json`.

common/comms/common_https.py
```diff
@@ -0,0 +1,27 @@
+from typing import Dict
+
+from tornado import httpclient
+
+from utilities import integration_adaptors_logger as log
+
+logger = log.IntegrationAdaptorsLogger("COMMON_HTTPS")
+
+
+class CommonHttps:
```
ismail-s, line 10: Optional: do you need this function to be inside a class?

common/comms/common_https.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @staticmethod
+    async def make_request(url: str, method: str, headers: Dict[str, str], body: str, client_cert: str = None,
+                     client_key: str = None, ca_certs: str = None):
```
ismail-s: Minor: maybe stick a docstring on this (nothing fancy).

common/comms/common_https.py
```diff
@@ -0,0 +1,27 @@
...
+                    {"method": method, "headers": headers, "url": url, "body": body})
+        response = await httpclient.AsyncHTTPClient().fetch(url,
+                                                            method=method,
+                                                            body=body,
+                                                            headers=headers,
+                                                            client_cert=client_cert,
+                                                            client_key=client_key,
+                                                            ca_certs=ca_certs,
+                                                            validate_cert=True)
+        logger.info("0002", "Sent {method} request with {headers} to {url} and received status code {code} : {body}",
```
ismail-s: This log is misleading, it makes it sound like body is the response body when it's actually the request body.

common/comms/tests/test_common_https.py
```diff
@@ -0,0 +1,58 @@
...
+CLIENT_CERT = "client.cert"
+CLIENT_KEY = "client.key"
+CA_CERTS = "ca.certs"
+
+
+class TestCommonHttps(TestCase):
+
+    @async_test
+    async def test_make_request(self):
+        with patch.object(httpclient.AsyncHTTPClient(), "fetch") as mock_fetch:
```
ismail-s, line 22: As a general point, note that `patch` (and `patch.object` and others) can be used as decorators, which can avoid nesting of code. But I wouldn't bother changing this code now, using it as a context manager is fine, it's when you're patching multiple things that it helps more.

Actually, ignore what I've just said, I'm assuming you're doing this because of something to do with requiring a running event loop?

common/utilities/config.py
```diff
@@ -25,16 +25,19 @@ def setup_config(component_name: str):
...
+    if key in config:
         return config[key]
-    except KeyError as e:
-        # Can't use IntegrationAdaptorsLogger due to circular dependency
+    elif default is not None:
+        return default
+    else:
         logging.exception(f'Failed to get config ConfigName:"{key}" ProcessKey=CONFIG001')
-        raise e
+        raise KeyError
```
ismail-s, line 44: V. minor: missing newline at end of file.

common/utilities/config.py
```diff
@@ -25,16 +25,19 @@ def setup_config(component_name: str):
...
     :param key: key to lookup
+    :param default: default value to return if none is found
     :return: the config variable
     """
-    try:
+
+    if key in config:
         return config[key]
-    except KeyError as e:
-        # Can't use IntegrationAdaptorsLogger due to circular dependency
```
ismail-s, line 38: Can you keep this comment in? Or if you want to get fancy, you could import the logger within this function, but that might be getting too fancy.

common/utilities/config.py
```diff
@@ -25,16 +25,19 @@ def setup_config(component_name: str):
             config[k[len(prefix):]] = v
 
 
-def get_config(key: str) -> str:
+def get_config(key: str, default: str = None) -> str:
```
ismail-s, line 28: V. minor/optional: a slightly better way of doing this is to have a private variable (ie one that starts with an underscore) that is a unique object, and use that here to distinguish when `default` isn't set. ie something like:
```python
_blah = object()
def get_config(key: str, default: str = _blah) -> str:
    if default == _blah:
        ...
    else:
        ....
```

mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -1,16 +1,16 @@
 """This module defines the asynchronous express workflow."""
 from typing import Tuple
 
-import requests
```
ismail-s, line 4: Can you also `pipenv uninstall requests` from wherever it is used, and also remove it from the `setup.py` file that it is in (I think it's in the `mhs_common` `setup.py` file).

mhs/outbound/main.py
```diff
@@ -70,11 +69,17 @@ def main():
 
     data_dir = pathlib.Path(definitions.ROOT_DIR) / "data"
     certs_dir = data_dir / "certs"
+    client_cert = "client.cert"
+    client_key = "client.key"
+    ca_certs = "client.pem"
+    max_retries = int(config.get_config('OUTBOUND_TRANSMISSION_MAX_RETRIES', default="3"))
```
ismail-s, line 75: Can you document this config in `mhs/README.md`.

mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -1,31 +1,49 @@
...
 
-    def make_request(self, interaction_details, message):
+    async def make_request(self, interaction_details: Dict[str, str], message: str) -> httpclient.HTTPResponse:
         """Make a request for the specified interaction, containing the provided message. Raises an exception if a
         non-success HTTP status code is returned by the server.
 
         :param interaction_details: A dictionary containing details of the interaction this message is for.
         :param message: The message body to send.
-        :return: The content of the response returned by the remote end.
+        :return: The tornado HTTPResponse object that represents the reponse of the object
```
ismail-s, line 47:
```suggestion
        :return: The tornado HTTPResponse object that represents the response of the object
```

mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -34,21 +52,53 @@ def make_request(self, interaction_details, message):
...
-                                  verify=self._verify)
-
-        # Raise an error if a 4xx or 5xx HTTP status was returned.
-        response.raise_for_status()
-
-        return response
+        retries_remaining = self._max_retries
+        while retries_remaining > 0:
+            try:
+                logger.info("0001", "About to send message with {headers} to {url} : {message}",
```
ismail-s, line 55: If this is being logged in `CommonHttps`, then I don't think we need to log here as well.

mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -57,10 +107,10 @@ def _build_headers(interaction_details):
...
 
     @staticmethod
-    def _get_request_method(interaction_details):
-        request_method = requests.post
+    def _get_request_method(interaction_details: Dict[str, str]):
+        request_method = "POST"
 
         if interaction_details['request_type'] == "GET":
-            request_method = requests.get
+            request_method = "GET"
```
ismail-s: So from speaking to Colin, we know that this will always be a POST

mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -1,31 +1,49 @@
...
-        client_key_path = str(Path(certs_dir) / "client.key")
-        self._cert = (client_cert_path, client_key_path)
-        self._verify = str(Path(certs_dir) / "client.pem")
+        self._client_cert = str(Path(certs_dir) / client_cert)
+        self._client_key = str(Path(certs_dir) / client_key)
+        self._ca_certs = str(Path(certs_dir) / ca_certs)
+        self._max_retries = max_retries
 
-    def make_request(self, interaction_details, message):
+    async def make_request(self, interaction_details: Dict[str, str], message: str) -> httpclient.HTTPResponse:
```
ismail-s: In combination with other comments, this should end up being changed to something like:
```suggestion
    async def make_request(self, url: str, message: str) -> httpclient.HTTPResponse:
```

mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -34,21 +52,53 @@ def make_request(self, interaction_details, message):
...
+            except Exception as e:
+                if self._is_retriable(e):
+                    retries_remaining -= 1
+                    logger.warning("0003",
+                                   "A retriable error was encountered {exception} {retries_remaining} {max_retries}",
+                                   {"exception": e,
+                                    "retries_remaining": retries_remaining,
+                                    "max_retries": self._max_retries
+                                    })
+                else:
```
ismail-s: Minor/optional; if this else is flipped, then the else can be removed.

mhs/outbound/outbound/transmission/tests/test_outbound_transmission.py
```diff
@@ -38,30 +53,85 @@ class TestOutboundTransmission(TestCase):
...
+            self.assertIs(actual_response, sentinel.result, "Expected content should be returned.")
+
+    @async_test
+    async def test_make_request_max_retries(self):
+        with patch.object(httpclient.AsyncHTTPClient(), "fetch") as mock_fetch:
+            mock_fetch.side_effect = httpclient.HTTPClientError(code=599)
+
+            with self.assertRaises(outbound_transmission.MaxRetriesExceeded):
+                await self.transmission.make_request(INTERACTION_DETAILS, "")
+                self.assertEqual(mock_fetch.call_count, MAX_RETRIES)
```
ismail-s: If you want this line to be called, I think you need it to be dedented once.

ismail-s (commented):
Also, Jenkins build is failing. Might just need to be manually run again.

mhs/outbound/outbound/transmission/outbound_transmission.py
ZameerParvez: I've hard coded this as POST now

mhs/outbound/outbound/transmission/outbound_transmission.py
ZameerParvez: I've made this change and updated the tests that are affected too. Now the Url and Headers are parameters for the make_request method on the outbound transmission object

common/comms/common_https.py
ZameerParvez: I did this to be consistent with some of the other packages/classes in common

common/comms/common_https.py
ZameerParvez: I've changed it, but I don't really think it's much better

mhs/outbound/outbound/transmission/tests/test_outbound_transmission.py
ZameerParvez: I haven't made this change, but will it not run because the error is thrown and handled by the assertRaises?

ZameerParvez: I've made this change now too

common/comms/common_https.py
```diff
@@ -0,0 +1,38 @@
...
+        response = await httpclient.AsyncHTTPClient().fetch(url,
+                                                            method=method,
+                                                            body=body,
+                                                            headers=headers,
+                                                            client_cert=client_cert,
+                                                            client_key=client_key,
+                                                            ca_certs=ca_certs,
+                                                            validate_cert=True)
+        logger.info("0002",
+                    "Sent {method} request with {headers} to {url} with body: {body}, and received status code {code}",
```
ismail-s:
```suggestion
                    "Sent {method} request with {headers} to {url} with {body}, and received status code {code}",
```
The logger component does the _body:blah_ stuff for you, so you don't need to do that in the log strings yourself.

mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -37,20 +37,17 @@ def __init__(self, certs_dir: str, client_cert: str, client_key: str, ca_certs:
...
         self._max_retries = max_retries
 
-    async def make_request(self, interaction_details: Dict[str, str], message: str) -> httpclient.HTTPResponse:
+    async def make_request(self, url: str, headers: Dict[str, str], message: str) -> httpclient.HTTPResponse:
         """Make a request for the specified interaction, containing the provided message. Raises an exception if a
         non-success HTTP status code is returned by the server.
 
-        :param interaction_details: A dictionary containing details of the interaction this message is for.
+        :param url: A string containing the url to send the request to.
+        :param headers: A dictionary containing details of the interaction this message is for.
```
ismail-s: This part of the docstring is wrong, the headers are literally just the HTTP headers

mhs/outbound/outbound/transmission/outbound_transmission.py
ismail-s: You've missed out removing the headers from `interactions.json`.

common/utilities/config.py
ismail-s: Comment outstanding.

mhs/common/mhs_common/workflow/asynchronous_express.py
ismail-s: Comment outstanding.

mhs/outbound/main.py
ismail-s: Comment outstanding.

ismail-s (commented):
The following comments are outstanding:
- https://github.com/nhsconnect/integration-adaptors/pull/60#discussion_r316271044
- https://github.com/nhsconnect/integration-adaptors/pull/60#discussion_r316267695
- https://github.com/nhsconnect/integration-adaptors/pull/60#discussion_r316264451

mhs/common/Pipfile
```diff
@@ -12,7 +12,6 @@ unittest-xml-reporting = "~=2.5"
 defusedxml = "~=0.6"
 integration-adaptors-common = {editable = true,path = "./../../common"}
 aioboto3 = "*"
-requests = "*"
```
ismail-s, line 15: `mhs/outbound/Pipfile` still has a requests dependency.

ismail-s (approved):

PR #61, opened 2019-08-22
RT-77: Integration tests are able to get full logs from AWS CloudWatch

anstm:
To get full logs from AWS Cloud Watch some changes were carried out:
- Full logs could be obtained using AWS Cloud Watch pagination machanism;
- Generators have been used to prevent memory leaks in case of large logs.


mhs/selenium_tests/page_objects/logs_retriever.py
```diff
@@ -0,0 +1,84 @@
...
+                                                          logStreamNamePrefix=log_prefix,
+                                                          nextToken=next_token)
+        next_token = log_streams["nextToken"]
+        streams = log_streams["logStreams"]
+        print(next_token)
+
+        for stream in streams:
+            log_stream_name = stream["logStreamName"]
+            log_events = aws_client.get_log_events(logGroupName=log_group_name, logStreamName=log_stream_name)["events"]
+            yield from [event['message'] for event in log_events]
```
ismail-s, line 81: Try doing:
```suggestion
            yield from (event['message'] for event in log_events)
```
This might reduce memory usage (if it works).

mhs/common/Pipfile
```diff
@@ -7,6 +7,7 @@ verify_ssl = true
 coverage = "~=4.5"
 tornado = "*"
 unittest-xml-reporting = "~=2.5"
+boto3 = "*"
```
ismail-s, line 10: Not sure this needs to be here.

mhs/selenium_tests/data/test.txt
```diff
@@ -0,0 +1,6 @@
+2019-07-25T09:37:38.123456 - INFO: 200 POST / (192.168.128.11) 21.94ms
```
ismail-s, line 1: Does this file need to be in this PR?

mhs/selenium_tests/page_objects/logs_retriever.py
```diff
@@ -0,0 +1,84 @@
...
+    if os.environ.get("MHS_ADDRESS", "localhost") == "localhost":
+        return _get_logs_from_file(component)
+    return _get_logs_from_aws_cloud_watch(component)
+
+
+def get_logs_as_list(component: Component) -> List[AnyStr]:
+    return list(get_logs(component))
+
+
+def _get_logs_from_file(component: Component) -> Generator:
```
ismail-s, line 37: Minor/Optional: could be more specific as to what types the generator returns, or maybe just do `Iterator[str]` (I'm basing this on what [the docs say](https://docs.python.org/3.7/library/typing.html#typing.Generator))

ismail-s (commented):
Not a thorough review as @Nayira will be doing a proper review.

Also, Jenkins build is failing.

PR #62, opened 2019-08-23
RT-156 use local dynamodb

ismail-s:
- Various changes to the local Docker compose dev cluster:
  - Configure the local DynamoDB with a state table
  - Configure MHS to connect to the local DynamoDB
  - Fix issue with MHS connection to the inbound RabbitMQ queue


ghost (approved):

PhilSmithBJSS (approved):

ismail-s:
Note that I've tested this locally as per the docs in the README.md file at the root of the repo, using the async-express workflow.

PR #63, opened 2019-08-27
RT-156 Add env vars for setting inbound queue login details

ismail-s:
- Add `MHS_INBOUND_QUEUE_USERNAME` and `MHS_INBOUND_QUEUE_PASSWORD` environment variables for setting the username and password
  - I've tested this locally with the async-express workflow and have checked the logs to see where the username/password gets logged. There was only one log line that included it: `inbound_1   | [2019-08-27T10:22:22.361533Z] connecting to Url('amqp://guest:guest@rabbitmq:5672/inbound')... pid=6 LogLevel=DEBUG`. That is a DEBUG log, so shouldn't be being logged in production, so I think this is working correctly.
- Remove the `CommonOutbound` file. I don't see any point in keeping this file, as we are currently expecting to only support a sync outbound request handler. If we need this file later, we can always add it again.

### DIFF
--- a/common/comms/proton_queue_adaptor.py
+++ b/common/comms/proton_queue_adaptor.py
@@ -31,10 +31,14 @@ class ProtonQueueAdaptor(comms.queue_adaptor.QueueAdaptor):
         Construct a Proton implementation of a :class:`QueueAdaptor <comms.queue_adaptor.QueueAdaptor>`.
         The kwargs provided should contain the following information:
           * host: The host of the Message Queue to be interacted with.
+          * username: The username to use to connect to the Message Queue.
+          * password The password to user to connect to the Message Queue.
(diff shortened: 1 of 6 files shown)


common/comms/proton_queue_adaptor.py
```diff
@@ -31,10 +31,14 @@ def __init__(self, **kwargs) -> None:
         Construct a Proton implementation of a :class:`QueueAdaptor <comms.queue_adaptor.QueueAdaptor>`.
         The kwargs provided should contain the following information:
           * host: The host of the Message Queue to be interacted with.
+          * username: The username to use to connect to the Message Queue.
+          * password The password to user to connect to the Message Queue.
```
Philip-Woods: Minor: typo in use/user

Philip-Woods (approved):

PR #64, opened 2019-08-27
Feature/RT-148: Sync-Async workflow (Inbound)

Philip-Woods:
This PR adds the inbound workflow for the sync-async messaging pattern, in particular this workflow updates the persistence store as needed, and pushes the inbound message data/payload onto the sync-async state store. The appropriate changes have been made to the  docker/docker-compose file. The outbound main file has also been updated to reflect the new signature of the workflow map generator. This has also been tested manually against the local dynamo db instance and it behaves as expected.


mhs/common/mhs_common/workflow/__init__.py
```diff
@@ -29,5 +33,6 @@ def get_workflow_map(party_key: str = None, persistence_store: persistence_adapt
         ASYNC_EXPRESS: AsynchronousExpressWorkflow(party_key, persistence_store, transmission, queue_adaptor),
         ASYNC_RELIABLE: AsynchronousReliableWorkflow(),
         FORWARD_RELIABLE: IntermediaryReliableWorkflow(),
+        SYNC_ASYNC: SyncAsyncWorkflow(party_key, transmission = transmission, sync_async_store=persistence_store),
```
ismail-s: Shouldn't this be:
```suggestion
        SYNC_ASYNC: SyncAsyncWorkflow(party_key, transmission = transmission, sync_async_store=sync_async_store),
```

mhs/common/mhs_common/workflow/__init__.py
```diff
@@ -10,14 +10,18 @@
...
 ASYNC_EXPRESS = 'async-express'
 ASYNC_RELIABLE = 'async-reliable'
 FORWARD_RELIABLE = 'forward-reliable'
 SYNC = 'sync'
+SYNC_ASYNC = 'sync-async'
 
 
-def get_workflow_map(party_key: str = None, persistence_store: persistence_adaptor.PersistenceAdaptor = None,
+def get_workflow_map(party_key: str = None,
+                     persistence_store: persistence_adaptor.PersistenceAdaptor = None,
```
ismail-s: Think this can be renamed to something more specific.

mhs/inbound/main.py
```diff
@@ -62,6 +66,7 @@ def start_inbound_server(certs_file: str, key_file: str, party_key: str,
                          ) -> None:
```
ismail-s, line 76: The type hint of the line above this looks wrong, it looks like it's saying that `persistence_store` is a module instead of a class instance.

mhs/common/mhs_common/workflow/__init__.py
ismail-s: This may potentially need to change when working on the outbound workflow, but that depends on what approach is taken in that task.

mhs/common/mhs_common/workflow/sync_async.py
```diff
@@ -10,27 +10,31 @@
...
-    def __init__(self, transmission: ta.TransmissionAdaptor, party_id: str):
-
+    def __init__(self,
+                 party_id: str,
+                 transmission: ta.TransmissionAdaptor = None,
+                 sync_async_store: pa.PersistenceAdaptor = None):
         """Create a new SyncAsyncWorkflow that uses the specified dependencies to load config, build a message and
         send it.
-
         :param transmission: The component that can be used to send messages.
```
ismail-s, line 34: Even when working on the outbound workflow, I don't think this transmission adaptor will be needed here, as the sync-async workflow itself doesn't make outbound requests, the other async workflows do that.

mhs/common/mhs_common/workflow/tests/test_sync_async.py
```diff
@@ -56,3 +63,40 @@ def test_send_message(self):
 
         self.mock_transmission.make_request.assert_called_with(sentinel.interaction_details, sentinel.message)
         self.assertIs(sentinel.response, actual_response)
+
+
+class TestSyncAsyncWorkFlowInbound(TestCase):
```
ismail-s:
```suggestion
class TestSyncAsyncWorkflowInbound(TestCase):
```

mhs/common/mhs_common/workflow/tests/test_sync_async.py
```diff
@@ -56,3 +63,40 @@ def test_send_message(self):
...
+
+    @test_utilities.async_test
+    async def test_inbound_workflow_happy_path(self):
+        persistence = MagicMock()
+        work_description = MagicMock()
+        work_description.set_status.return_value = test_utilities.awaitable(True)
+        persistence.add.return_value = test_utilities.awaitable(True)
+        
+        self.workflow = sync_async.SyncAsyncWorkflow(PARTY_ID, transmission=None, sync_async_store=persistence)
```
ismail-s: Common setup of `self.workflow` can be moved into a `setUp` method.

mhs/common/mhs_common/workflow/tests/test_sync_async.py
```diff
@@ -1,19 +1,26 @@
...
 import mhs_common.messages.ebxml_request_envelope as ebxml_request_envelope
 from mhs_common.workflow import sync_async
+from mhs_common.state import work_description as wd
+from utilities import test_utilities
 
 PARTY_ID = "PARTY-ID"
 MOCK_UUID = "5BB171D4-53B2-4986-90CF-428BE6D157F5"
 
 
+def throws_exception(x, y):
```
ismail-s: Optional: unless you expect this function to be used elsewhere, I think it's better to be moved into the one test that it's used in. It feels a bit random it being here and accepting 2 random params.

ismail-s (changes requested):

mhs/common/mhs_common/workflow/__init__.py
Philip-Woods: `persistence store`? I've updated it to `work_description_store` to reflect what it actually contains, I've updated the inbound handler and main files to make it consistent

Philip-Woods: I've also updated the `queue_adaptor` parameter because was the same as the module name we are importing: `from comms import queue_adaptor`

mhs/common/mhs_common/workflow/sync_async.py
Philip-Woods: I considered removing the references to the transmission but since the current sync-async outbound relies on it I think I will leave it as to avoid removing code/tests that might be needed/useful for the outbound ticket

mhs/common/mhs_common/workflow/tests/test_sync_async.py
Philip-Woods: I've done as you suggested and also tried to name it a bit more descriptively

ismail-s (dismissed):
Looks fine, though I'm not sure about retry logic, if you need that when putting the message onto the sync-async table. Before just adding this, check whether AWS boto3 does this already (I think it might do, I feel like I've seen logs mentioning retries). If it does, then we'd need to think about what's best to do in terms of where retries should be being done, whether we should make those retries configurable.

Philip-Woods:
This has now been updated to include retry logic for adding messages to the sync-async store (I did a quick check on the boto3 library as suggested and couldn't find any immediate information on default retries for get/set items). The updates briefly included retry logic for the `work_description` store updates but I think since these update types will be done as part of every workflow it should be done in a more generic way if it's a feature we need/want.

mhs/inbound/main.py
```diff
@@ -32,7 +32,13 @@ def initialise_workflows() -> Dict[str, workflow.CommonWorkflow]:
     sync_async_store = dynamo_persistence_adaptor.DynamoPersistenceAdaptor(
         table_name=config.get_config('SYNC_ASYNC_STATE_TABLE_NAME'))
 
-    return workflow.get_workflow_map(inbound_async_queue=queue_adaptor, sync_async_store=sync_async_store)
+    sync_async_retries = int(config.get_config('SYNC_ASYNC_STORE_MAX_RETRIES', default='3'))
+    sync_async_delay = int(config.get_config('SYNC_ASYNC_STORE_RETRY_DELAY', default='100'))
```
ismail-s, line 38: Document these in `mhs/README.md`

mhs/common/mhs_common/workflow/sync_async.py
```diff
@@ -47,18 +52,30 @@ def __init__(self,
...
+        while True:
+            try:
+                await self.sync_async_store.add(key, data)
+                logger.info('003', 'Successfully updated state store')
+                break
+            except Exception as e:
+                logger.warning('021', 'Exception raised while adding to sync-async store {exception} {retry}',
+                               {'exception': e, 'retry': retry})
+                if retry == 0:
+                    logger.error('022', 'Final retry has been attempted for adding message to sync async store')
```
ismail-s: So in the other retry places, this was logged as WARNING and not ERROR. I don't want this comment to hold up this PR, but I think we need to be clear in the future as to the difference between the two.

mhs/common/mhs_common/workflow/sync_async.py
```diff
@@ -47,18 +52,30 @@ def __init__(self,
...
+                    raise e
```
ismail-s: In my PR, I was using the retry exception used in the outbound transmission, which I'd moved to the common package. Though I haven't merged my PR yet (Jenkins build taking forever).

mhs/common/mhs_common/workflow/sync_async.py
```diff
@@ -47,18 +52,30 @@ def __init__(self,
...
+                retry -= 1
```
ismail-s, line 75: Because this line is before the if statement, there's an out-by-one error, where if eg the number of retries is set to 1, then 2 attempts are made. Actually, you could argue that this is correct, but in any case, I think it's different to the retrying in the outbound transmission and the one I was doing in the inbound async-express workflow, and I think they should all be consistent.

ismail-s (dismissed):

mhs/README.md
```diff
@@ -44,6 +44,8 @@ MHS takes a number of environment variables when it is run. These are:
 * `MHS_INBOUND_QUEUE_PASSWORD` (inbound only) The password to use when connecting to the amqp inbound queue.
 * `MHS_INBOUND_QUEUE_MAX_RETRIES` (inbound only) The max number of times to retry putting a message onto the amqp inbound queue. Defaults to 3.
 * `MHS_INBOUND_QUEUE_RETRY_DELAY` (inbound only) The delay in milliseconds between retrying putting a message onto the amqp inbound queue. Defaults to 100ms.
+* `SYNC_ASYNC_STORE_MAX_RETRIES'` (inbound only) The max number of retries when attempting to add a message to the sync-async store.
+* `SYNC_ASYNC_STORE_RETRY_DELAY` (inbound only) The delay in milliseconds between retrying placing a message on the sysnc-async store.
```
ismail-s: Can mention default values as well.

ismail-s (dismissed):

ismail-s (approved):

PR #65, opened 2019-08-28
RT-152: Add a delay between outbound transmission's retries

ghost:
This PR adds a configurable delay between the retries made by the outbound transmission component as discussed in the comments of [RT-152](https://gpitbjss.atlassian.net/browse/RT-152).  A default delay of 100ms is used.

I've verified that with this change I am still able to successfully make a request to Opentest.

### DIFF
--- a/mhs/outbound/outbound/transmission/outbound_transmission.py
+++ b/mhs/outbound/outbound/transmission/outbound_transmission.py
@@ -1,13 +1,13 @@
 """This module defines the outbound transmission component."""
 
+import asyncio
 from pathlib import Path
 from ssl import SSLError
 from typing import Dict
 
+from comms.common_https import CommonHttps
 from mhs_common.transmission import transmission_adaptor
 from tornado import httpclient
-
-from comms.common_https import CommonHttps
 from utilities import integration_adaptors_logger as log
 
 logger = log.IntegrationAdaptorsLogger("OUTBOUND_TRANSMISSION")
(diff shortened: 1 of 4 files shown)


mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -65,6 +68,10 @@ def __init__(self, certs_dir: str, client_cert: str, client_key: str, ca_certs:
                                 "max_retries": self._max_retries
                                 })
 
+                logger.info("0005", "Waiting for {retry_delay} milliseconds before next request attempt.",
+                            {"retry_delay": self._retry_delay})
+                await asyncio.sleep(self._retry_delay / 1000)
```
ismail-s, line 73: Optional: possibly better to use `tornado.gen.sleep`.

ismail-s (approved):

mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -25,18 +25,21 @@ class OutboundTransmission(transmission_adaptor.TransmissionAdaptor):
...
-    def __init__(self, certs_dir: str, client_cert: str, client_key: str, ca_certs: str, max_retries: int):
+    def __init__(self, certs_dir: str, client_cert: str, client_key: str, ca_certs: str, max_retries: int,
+                 retry_delay: int):
         """Create a new OutboundTransmission that loads certificates from the specified directory.
         :param certs_dir: A string containing the path to the directory to load certificates from.
         :param client_cert: A string containing the name of the client certificate file in the certs directory.
         :param client_key: A string containing the name of the client private key file in the certs directory.
         :param ca_certs: A string containing the name of the certificate authority certificate file in the certs directory.
         :param max_retries: An integer with the value of the max number times to retry sending the request.
+        :param retry_delay: An integer representing the delay (in milliseconds) to use between retry attempts.
```
Philip-Woods, line 36: The number of parameters here seems to be getting a bit long to me, I think it's probably fine for the time being but I think if there's ever another parameter added or we get time for a refactor it's probably worth splitting these into their own certs/retry data objects

Philip-Woods (approved):
Looks good, just one comment about a possible future change to make it a bit neater but it's not really a priority currently

mhs/outbound/outbound/transmission/outbound_transmission.py
ghost: As discussed offline, I can't see any particular advantage to using the tornado function over asyncio's built-in version, so I'll keep this as it is.

PR #68, opened 2019-08-28
RT-123: Create New Spine Route Lookup Service

ghost:
This PR adds a Spine Route Lookup service container to the MHS. This includes:

- A new packer file (`pipeline/packer/spineroutelookup.json`) to create the docker image, which is built in the package stage of the Jenkins build and as part of `build.sh`
- Adding new task and service definitions to the Terraform file (`pipeline/terraform/test-environment/test-environment.tf` to ensure an instance of the docker container is stood up (and doing the same in the local `docker-compose.yml`)
- Adding the necessary configuration variables to the terraform and docker-compose files to point configure the new service and point the MHS at it
- Added three Tornado request handlers to the spine route lookup service, to allow routing, reliability and a combined list to be returned (see `mhs/spineroutelookup/request/`)
- Made the spine route lookup service set up the existing routing & reliability components (`mhs/spineroutelookup/main.py`)
- Adding a `RoutingAndReliability` component to the MHS common package (`mhs/common/mhs_common/routing/routing_reliability.py`. This functions as a facade for the spine route lookup service, making HTTP requests to it when it is called
- Made the asynchronous express workflow use the new routing and reliability component to obtain the URL (and party key) for the interaction a request is being made for

## Known Issues:
- The endpoint URL returned from SDS for our sample async express interaction is `https://192.168.128.11/reliablemessaging/queryrequest`. Since this uses the IP address, rather than a hostname, the certificates that are presented by the Opentest instance of Spine fail verification when we attempt to connect, as they identify the subject using its hostname. I have confirmed with the Opentest support team that this is an error they plan to correct, but their timeframe for the fix is in the order of weeks so I have worked around this by (temporarily!) disabling certificate validation for the outbound transmission component only (see ab92612)
- In 02cda0c I added logic that expects there to always be a single endpoint URL returned from SDS. This was agreed with @colin-bateson and Richard Dobson.


mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -21,29 +27,36 @@ class AsynchronousExpressWorkflow(common_asynchronous.CommonAsynchronousWorkflow
 
     def __init__(self, party_key: str = None, persistence_store: persistence_adaptor.PersistenceAdaptor = None,
                  transmission: transmission_adaptor.TransmissionAdaptor = None,
-                 queue_adaptor: queue_adaptor.QueueAdaptor = None):
+                 queue_adaptor: queue_adaptor.QueueAdaptor = None,
+                 routing_reliability: routing_reliability.RoutingAndReliability = None):
```
Philip-Woods: Might want to be careful with this as `routing_reliability` is also the name of the module as well as the variable

mhs/common/mhs_common/workflow/tests/test_asynchronous_express.py
```diff
@@ -13,7 +13,8 @@
 from mhs_common.state import work_description
 from mhs_common.state.work_description import MessageStatus
 
-PARTY_KEY = 'party-key'
+FROM_PARTY_KEY = 'to-party-key'
+TO_PARTY_KEY = 'from-party-key'
```
Philip-Woods: Are these the wrong way round?

mhs/outbound/main.py
```diff
@@ -31,16 +32,20 @@ def load_party_key(data_dir: pathlib.Path) -> str:
...
     :param party_key: The party key to use to identify this MHS.
     :param persistence_store: The persistence adaptor for the state database
+    :param routing_reliability: The routing and reliability component to use to request routing/reliability details
+    from.
     :return: The workflows that can be used to handle messages.
     """
 
-    return workflow.get_workflow_map(party_key, persistence_store, transmission)
+    return workflow.get_workflow_map(party_key=party_key, persistence_store=persistence_store,
+                                     transmission=transmission, routing_reliability=routing_reliability)
```
Philip-Woods: Again here this is using the module name for a variable

mhs/spineroutelookup/request/tests/test_routing_reliability_handler.py
```diff
@@ -0,0 +1,36 @@
...
+        self.routing.get_end_point.return_value = test_utilities.awaitable(END_POINT_DETAILS)
+        self.routing.get_reliability.return_value = test_utilities.awaitable(RELIABILITY_DETAILS)
+
+        url = tornado.httputil.url_concat("/", {"org-code": ORG_CODE, "service-id": SERVICE_ID})
+        response = self.fetch(url, method="GET")
+
+        self.assertEqual(response.code, 200)
+        self.assertEqual(COMBINED_DETAILS, json.loads(response.body))
+        self.routing.get_end_point.assert_called_with(ORG_CODE, SERVICE_ID)
+        self.routing.get_reliability.assert_called_with(ORG_CODE, SERVICE_ID)
```
Philip-Woods: Would there be any value in testing exception cases in the routing module, and maybe as a broader question should the endpoints have wrappers for exceptions that arise during calls to the routing module?

pipeline/packer/spineroutelookup.json
```diff
@@ -0,0 +1,77 @@
...
+    "tag": "{{env `BUILD_TAG`}}",
+    "docker_registry_url": "{{env `DOCKER_REGISTRY_URL`}}"
+  },
+  "builders": [
+    {
+      "type": "docker",
+      "image": "python:3-slim",
+      "commit": true,
+      "changes": [
+        "EXPOSE 443",
```
Philip-Woods: Should this be port 80? The `main` file for spine route lookup exposes port 80 at least

mhs/spineroutelookup/request/reliability_handler.py
```diff
@@ -0,0 +1,30 @@
...
+        org_code = self.get_query_argument("org-code")
+        service_id = self.get_query_argument("service-id")
+
+        logger.info("001", "Looking up reliability information. {org_code}, {service_id}",
+                    {"org_code": org_code, "service_id": service_id})
+        reliability_info = await self.routing.get_reliability(org_code, service_id)
+        logger.info("002", "Obtained reliability information. {reliability_information}",
+                    {"reliability_information": reliability_info})
+
+        self.write(reliability_info)
```
Philip-Woods, line 30: Think I mentioned it before but I'm wondering if it's worth catching any errors here and generating easy to understand error responses for the outbound service?

mhs/spineroutelookup/request/routing_handler.py
```diff
@@ -0,0 +1,29 @@
+import tornado.web
+from utilities import timing, integration_adaptors_logger as log
+
+from lookup import routing_reliability
+
+logger = log.IntegrationAdaptorsLogger('SPINE_ROUTE_LOOKUP_ROUTING_REQUEST_HANDLER')
```
Philip-Woods: Seems to me that having `SPINE_ROUTE_LOOKUP_` prefixing these handler logs might be a bit verbose?

pipeline/terraform/test-environment/test-environment.tf
```diff
@@ -96,6 +96,14 @@ resource "aws_ecs_task_definition" "test-environment-mhs-outbound-task" {
...
           name = "MHS_STATE_TABLE_NAME",
           value = var.mhs_state_table_name
+        },
+        {
+          name = "MHS_SPINE_ROUTE_LOOKUP_URL",
+          value = var.mhs_spine_route_lookup_url
+        },
+        {
+          name = "MHS_SPINE_ORG_CODE",
+          value = var.mhs_spine_org_code
```
Philip-Woods, line 134: I guess this is a question about the uncertainty I had whilst doing SDS, but do we have an org code for spine? Seemed like the the environment docs suggested there was a fixed party key that should be used to retrieve the information

pipeline/terraform/test-environment/test-environment.tf
```diff
@@ -117,6 +125,67 @@ resource "aws_ecs_task_definition" "test-environment-mhs-outbound-task" {
...
+  container_definitions = jsonencode(
+  [
+    {
+      name = "spineroutelookup"
+      image = "${var.ecr_address}:spineroutelookup-${var.build_id}"
+      essential = true
+      logConfiguration = {
+        logDriver = "awslogs"
+        options = {
+          awslogs-group = "/ecs/spineroutelookup-test-environment"
```
Philip-Woods, line 173: Has this log group been created in aws? I can't seem to spot it anywhere?

pipeline/terraform/test-environment/test-environment.tf
```diff
@@ -117,6 +125,67 @@ resource "aws_ecs_task_definition" "test-environment-mhs-outbound-task" {
...
+          containerPort = 80
+          hostPort = var.spineroutelookup_service_port
+          protocol = "tcp"
+        }
+      ]
+    }
+  ]
+  )
+  cpu = "128"
+  memory = "256"
```
Philip-Woods, line 210: Can't seem to find the logs for this container, but I think with 4 containers at 256 memory I think that's the limit of the ecs instance, I'd maybe just double check that all 4 services are starting correctly

mhs/common/mhs_common/workflow/tests/test_asynchronous_express.py
ghost: Oops, they are. Thanks.

mhs/common/mhs_common/workflow/asynchronous_express.py
ghost: Will change.

mhs/outbound/main.py
ghost: Will change.

pipeline/terraform/test-environment/test-environment.tf
ghost: No, good spot. I've created it now (and added details to our README).

pipeline/packer/spineroutelookup.json
ghost: Yes, it should, thanks.

pipeline/terraform/test-environment/test-environment.tf
ghost: We confirmed with the Spine team that an MHS should have the Spine ODS code as a configuration item and then can use this for SDS lookups.

mhs/spineroutelookup/request/routing_handler.py
ghost: Will update them to `SRL_*`.

mhs/spineroutelookup/request/tests/test_routing_reliability_handler.py
ghost: As discussed offline, Tornado does a pretty good job of handling errors raised from within these handlers, so I didn't want to overcomplicate things. I'll add cases to the test classes to ensure an HTTP 500 is returned when the routing component raises exceptions, and that a 400 is returned if the required query parameters are not provided.

ghost:
@Philip-Woods - I believe this is ready for re-review. As well as addressing your review comments I've done the following:
- Disabled certificate verification for outbound transmission's requests
- Obtained the CPA ID for messages from SDS
- Added logic to handle the fact that the endpoint URLs returned by SDS are a list (errors if 0 returned, warns and uses the first if >1)
- Extracted the routing lookup to the common workflow, and the reliability one to the common async workflow

mhs/README.md
```diff
@@ -50,7 +50,17 @@ MHS takes a number of environment variables when it is run. These are:
...
 * `MHS_RESYNC_INTERVAL` The time in between polls of the sync-async store, the interval is in seconds and defaults to 1
-
+* `MHS_SPINE_ROUTE_LOOKUP_URL` (outbound only) The URL of the Spine route lookup service. E.g `https://example.com`. This URL should not contain path or query parameter parts.
+* `MHS_SPINE_ORG_CODE` (outbound only) The organisation code for the Spine instance that your MHS is communicating with. E.g `YES`
+* `MHS_SDS_URL` (Spine Route Lookup service only) The URL to communicate with SDS on. e.g. `ldaps://example.com`
+* `MHS_DISABLE_SDS_TLS` (Spine Route Lookup service only) A flag that can be set to disable TLS for SDS connections.
+*Must* be set to exactly `True` for TLS to be disabled.
+* `MHS_SDS_CACHE_IMPLEMENTATION` (Spine Route Lookup service only) Identifies the caching implementation to use when
+caching values retrieved from SDS. This value is optional. Possible values are:
+    * `DICTIONARY_CACHE` - An in-memory cache using a Python dictionary. Not suitable for production use. If no
```
Philip-Woods, line 65: What's the purpose of this `DICTIONARY_CACHE` env variable?

mhs/common/mhs_common/routing/routing_reliability.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        if org_code is None:
+            org_code = self.spine_org_code
+            logger.info("0007", "No org code provided when obtaining endpoint details. Using {spine_org_code}",
+                        {"spine_org_code": org_code})
+
+        url = self._build_request_url(ROUTING_PATH, org_code, service_id)
+
+        try:
+
```
Philip-Woods, line 43: Optional: Might want to keep them separate for the different log lines but these two try blocks seem to do pretty much the same thing, might be neater to pull them out into a method

mhs/common/mhs_common/workflow/common_asynchronous.py
```diff
@@ -1,8 +1,33 @@
...
+MHS_PERSIST_DURATION = "nhsMHSPersistDuration"
+MHS_DUPLICATE_ELIMINATION = "nhsMHSDuplicateElimination"
+MHS_ACK_REQUESTED = "nhsMHSAckRequested"
+
 
 class CommonAsynchronousWorkflow(common.CommonWorkflow):
     """Common functionality across all asynchronous workflows."""
-    pass
+
+    async def _lookup_reliability_details(self, interaction_details: Dict) -> Dict:
```
Philip-Woods, line 21: Just to double check but is reliability exclusive to async workflows? Is there no retry information for sync messages?

mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -49,9 +49,14 @@ def __init__(self, certs_dir: str, client_cert: str, client_key: str, ca_certs:
...
                             {"headers": headers, "url": url, "message": message})
+                # ******************************************************************************************************
+                # TLS CERTIFICATE VALIDATION HAS BEEN TEMPORARILY DISABLED! This is required because Opentest's SDS
+                # instance currently returns endpoints as IP addresses. This MUST be changed before this code is used in
+                # a production environment
+                # ******************************************************************************************************
                 response = await CommonHttps.make_request(url=url, method=request_method, headers=headers, body=message,
                                                           client_cert=self._client_cert, client_key=self._client_key,
-                                                          ca_certs=self._ca_certs)
+                                                          ca_certs=self._ca_certs, validate_cert=False)
```
Philip-Woods, line 59: Just wondering if it's worth placing a warning in the `README` about this TLS disable. I guess depending on time the other option might be to add an env variable dictates the certs validation, it could default to True so in a normal system the check is run by default but in our system we can set it to false

mhs/spineroutelookup/lookup/routing_reliability.py
```diff
@@ -11,7 +11,7 @@
     'nhsMHSAckRequested'
 ]
 
-ROUTING_KEYS = ['nhsMhsFQDN', 'nhsMHSEndPoint']
+ROUTING_KEYS = ['nhsMhsFQDN', 'nhsMHSEndPoint', 'nhsMHSPartyKey', 'nhsMhsCPAId']
```
Philip-Woods, line 14: I noticed as part of Colin's PR that the sync message wrapper is using the destination ASID as well, just wondering if that's definitely present for spine and if so if it could be added here

mhs/spineroutelookup/request/tests/test_routing_handler.py
```diff
@@ -0,0 +1,31 @@
...
+import tornado.web
+from utilities import test_utilities
+
+from request import routing_handler
+from request.tests import test_request_handler
+
+END_POINT_DETAILS = {"end_point": "http://www.example.com"}
+
+
+class TestRoutingRequestHandler(test_request_handler.TestRequestHandler, tornado.testing.AsyncHTTPTestCase):
```
Philip-Woods: Just a general comment but I guess normally we were attempting to avoid multiple inheritance and in theory we should hold test code to the same standard as prod but there's no real risk of any multiple inheritance side effects so I'm not too bothered about this being altered really

mhs/outbound/outbound/transmission/outbound_transmission.py
ghost: I wasn't looking to make this too complicated, since it's really a workaround until Opentest is fixed, and I don't think it's necessarily something we need to support after that.

I'll add a note to the README to make it explicitly clear that the current version of the MHS isn't doing certificate validation.

mhs/common/mhs_common/workflow/common_asynchronous.py
ghost: I've confirmed with @colin-bateson, no reliability information is needed for sync workflows.

mhs/README.md
ghost: The idea was that each implementation of the SDS cache interface will register itself in a dictionary, and then the desired implementation can be loaded from there based on a config key. It's a pattern I'd discussed with Dave previously, which I thought made sense to use here since we plan to implement a second implementation of the SDS cache interface imminently

See https://github.com/nhsconnect/integration-adaptors/pull/68/commits/76825f249aa37cba6694653ced12235647cef94c for the implementation.

mhs/common/mhs_common/routing/routing_reliability.py
ghost: As you say, I was trying to keep the separate log keys for these two cases, and as the actual methods aren't very complex I'd be inclined to keep them as they are, unless you feel strongly about reducing the duplication.

mhs/spineroutelookup/request/tests/test_routing_handler.py
ghost: It's a fair point, and there's not a huge benefit to doing this, so I'll remove the inheritance.

mhs/README.md
Philip-Woods: Sorry I just noticed that `DICTIONARY_CACHE` is indented, and thus being one of the expected values of `MHS_SDS_CACHE_IMPLEMENTATION`, that makes perfect sense now 👍

mhs/common/mhs_common/routing/routing_reliability.py
Philip-Woods: I'm fine with that, having separate log messages makes sense and the code is only duplicated once so I'm fine with it

Philip-Woods (approved):

mhs/spineroutelookup/lookup/routing_reliability.py
ghost: Agreed offline with @colin-bateson & @Philip-Woods that this would be addressed as part of the sync workflow task, if required.

PR #70, opened 2019-08-29
Feature/rt 157 common logging package enhancements

ZameerParvez:
The traceback is formatted as an array of strings, which are the lines that are usually shown when an error is thrown. It is also added to the log whenever a log is taken when an error is being handled.


ZameerParvez:
The traceback for an error is formatted as an array of strings, which are the lines that are usually shown when an error is thrown

common/utilities/integration_adaptors_logger.py
```diff
@@ -36,7 +37,7 @@ def formatTime(self, record, datefmt):
     handler = logging.StreamHandler(sys.stdout)
 
     formatter = _CustomFormatter(
-        fmt='[%(asctime)sZ] %(message)s pid=%(process)d LogLevel=%(levelname)s ',
+        fmt='[%(asctime)sZ] %(message)s pid=%(process)d LogLevel=%(levelname)s LoggerName=%(name)s',
```
Philip-Woods, line 40: Can I ask where this `LoggerName` came from? It seems to be a repeat information of the `ProcessKey` which holds the same value except it's concatenated with a process key number.

Edit: Just seen Ismail's comment on the actual ticket regarding distinguishing logs between our code and other libraries, still feels a little annoying that there will be repeated information in each of our own logs but I can't think of another way to do it off the top of my head

common/utilities/integration_adaptors_logger.py
```diff
@@ -102,6 +102,15 @@ def _format_values_in_map(self, dict_values: dict) -> dict:
         Replaces the values in the map with key=value so that the key in a string can be replaced with the correct
         log format, also surrounds the value with quotes if it contains spaces and removes spaces from the key
         """
+
+        for key in dict_values.keys():
+            if str(key).lower() == "exception":
```
Philip-Woods, line 110: I feel like in theory we shouldn't ever really be using something that isn't a `string` for the key anyway, but this check is fine as it is.

common/utilities/integration_adaptors_logger.py
```diff
@@ -102,6 +102,15 @@ def _format_values_in_map(self, dict_values: dict) -> dict:
...
+                exc_inf = sys.exc_info()
+                if all(exc_inf):
+                    (etype, value, trace) = exc_inf
+                    dict_values[key] = f'{value}{traceback.format_exception(etype, value, trace)}'
+                    break
```
Philip-Woods: Optional: I'd maybe consider moving some of this to an aptly named method that returns the formatted string, it's doesn't feel immediately obvious that this is actually retrieving the last exception from the stack - or maybe just add a line to the docstring explaining whats happening to exceptions.

Philip-Woods (commented):
Just one question on the `LoggerName` and a possible improvement to the bit that formats the exception but I'm happy to approve once the `LoggerName` stuff is confirmed

common/utilities/integration_adaptors_logger.py
Philip-Woods, line 110: I'd maybe add to one of the higher level docs (Maybe the `__init__`) that exceptions should be keyed as `{exception}` so that someone doesn't have to dig down to this line to find out that exceptions won't print correctly if they aren't keyed appropriately

Philip-Woods (approved):

PR #73, opened 2019-09-02
Fix/Int-Tests

Philip-Woods:
Adds a missing SYNC_ASYNC_STATE_TABLE_NAME variable to the terraform and jenkins script.

This also splits the `status` of a work description into two types (inbound and outbound) and adds retry logic for updating the work description status as a result of a race condition causing the integration tests to fail

### DIFF
--- a/mhs/common/mhs_common/state/tests/test_work_description.py
+++ b/mhs/common/mhs_common/state/tests/test_work_description.py
@@ -14,7 +14,8 @@ input_data = {
         wd.CREATED_TIMESTAMP: '11:59',
         wd.LATEST_TIMESTAMP: '12:00',
         wd.VERSION_KEY: 1,
-        wd.STATUS: wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED,
+        wd.INBOUND_STATUS: None,
+        wd.OUTBOUND_STATUS: wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED,
         wd.WORKFLOW: workflow.SYNC
     }
 }
@@ -25,7 +26,8 @@ old_data = {
         wd.VERSION_KEY: 0,
         wd.CREATED_TIMESTAMP: '11:59',
         wd.LATEST_TIMESTAMP: '12:00',
-        wd.STATUS: wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED,
+        wd.INBOUND_STATUS: wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED,
+        wd.OUTBOUND_STATUS: None,
         wd.WORKFLOW: workflow.SYNC
     }
 }
@@ -37,7 +39,7 @@ class TestWorkDescription(unittest.TestCase):
         persistence = MagicMock()
         work_description = wd.WorkDescription(persistence, input_data)
 
-        self.assertEqual(work_description.status, wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED)
+        self.assertEqual(work_description.outbound_status, wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED)
         self.assertEqual(work_description.version, 1)
         self.assertEqual(work_description.created_timestamp, '11:59')
         self.assertEqual(work_description.last_modified_timestamp, '12:00')
@@ -77,7 +79,7 @@ class TestWorkDescription(unittest.TestCase):
             wd.DATA: {
                 wd.VERSION_KEY: 3,
                 wd.LATEST_TIMESTAMP: '11:00',
-                wd.STATUS: wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED
+                wd.OUTBOUND_STATUS: wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED
             }
         })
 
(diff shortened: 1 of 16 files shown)


mhs/common/mhs_common/state/tests/test_work_description.py
```diff
@@ -98,7 +100,7 @@ def test_constructor(self):
             wd.DATA: {
                 wd.VERSION_KEY: 1,
                 wd.LATEST_TIMESTAMP: '11:00',
-                wd.STATUS: wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED
+                wd.INBOUND_STATUS: wd.MessageStatus.OUTBOUND_MESSAGE_PREPARED,
```
ghost: I'm not sure it affects this test case, but does it make sense for the inbound status to be `OUTBOUND_MESSAGE_PREPARED`?

mhs/common/mhs_common/state/work_description.py
```diff
@@ -45,6 +45,24 @@ class EmptyWorkDescriptionError(RuntimeError):
...
+                                     status: MessageStatus,
+                                     retries: int):
+    attempts = 0
+    while attempts < retries + 1:
+        try:
+            await wdo.update()
+            await update_status_method(status)
+            break
+        except OutOfDateVersionError as e:
+            logger.error('0021', 'Failed attempt to update state store')
```
ghost: Perhaps a warning level log here, indicating that we'll be retrying (and we're on x of y retries), and then an error log in the `if attempts == retries` block?

mhs/common/mhs_common/state/work_description.py
```diff
@@ -45,6 +45,24 @@ class EmptyWorkDescriptionError(RuntimeError):
     pass
 
 
+async def update_status_with_retries(wdo: WorkDescription,
```
ghost, line 48: I think we need tests for this method in `test_work_description.py`, to ensure we make the correct number of retries, etc.

mhs/common/mhs_common/state/work_description.py
```diff
@@ -81,9 +100,6 @@ def create_new_work_description(persistence_store: pa.PersistenceAdaptor,
     if not key:
         logger.error('005', 'Failed to build new work description, key should not be null or empty')
         raise ValueError('Expected key to not be None or empty')
-    if status is None:
```
ghost, line 84: Could we not make this check whether inbound _and_ outbound status are null? Presumably we need at least one of them?

mhs/common/mhs_common/state/work_description.py
```diff
@@ -155,13 +173,33 @@ def __init__(self, persistence_store: pa.PersistenceAdaptor, store_data: dict):
         logger.info('016', 'Successfully updated work description to state store for {key}', {'key': self.message_key})
         return old_data
 
-    async def set_status(self, new_status: MessageStatus):
+    async def update(self):
```
ghost, line 181: Are you missing test cases for this method?

mhs/common/mhs_common/state/work_description.py
Philip-Woods: Added the test back in as well

mhs/common/mhs_common/state/work_description.py
Philip-Woods: Updated to include the test cases for the logic in this method, then I also included one for the new `_deserialize_data` method as well

mhs/common/mhs_common/state/tests/test_work_description.py
```diff
@@ -127,15 +129,85 @@ def test_constructor(self):
...
+        work_description = wd.WorkDescription(persistence, old_data)
+
+        await work_description.update()
+
+        self.assertEqual(work_description.version, 1)
+        self.assertEqual(work_description.outbound_status, wd.MessageStatus.OUTBOUND_MESSAGE_ACKD)
+        self.assertEqual(work_description.last_modified_timestamp, '13:00')
+
+    @async_test
+    async def test_update_status(self):
```
ghost: This has the same name as the above method (which I think may result in one of these test cases not running!):
```suggestion
    async def test_update_status_no_data(self):
```

mhs/common/mhs_common/state/tests/test_work_description.py
```diff
@@ -127,15 +129,85 @@ def test_constructor(self):
...
+                wd.INBOUND_STATUS: None,
+                wd.OUTBOUND_STATUS: wd.MessageStatus.OUTBOUND_MESSAGE_ACKD,
+                wd.WORKFLOW: workflow.SYNC
+            }
+        }
+
+        work_description = wd.WorkDescription(MagicMock(), old_data)
+        work_description._deserialize_data(new_data)
+
+        self.assertEqual(work_description.version, 1)
```
ghost, line 207: For the sake of a few more assert statements I might be inclined to check that all of the expected fields have been set here.

mhs/common/mhs_common/state/tests/test_work_description.py
```diff
@@ -236,3 +289,49 @@ def test_create_wd_null_parameters(self):
...
+        self.assertEqual(wdo.set_outbound_status.call_count, 1)
+
+    @test_utilities.async_test
+    async def test_update_correct_number_retries(self):
+        async def throws_out_of_date_exception(param):
+            raise wd.OutOfDateVersionError
+
+        wdo = MagicMock()
+        wdo.update.return_value = test_utilities.awaitable(True)
+        wdo.set_outbound_status.side_effect = throws_out_of_date_exception
```
ghost: Not something you have to change now, but I think you could have just specified `wd.OutOfDateVersionError` as the side-effect here, and it would be raised.

ghost (dismissed):
Happy for this to go in with the suggested changes.

mhs/common/mhs_common/state/tests/test_work_description.py
Philip-Woods: Thought this had to be set so that the call something that can be awaited but I've just tested it and it seems it doesn't matter  so changed it to your suggestion

ghost (approved):

PR #74, opened 2019-09-03
Feature/rt 147 sync async outbound workflow

Philip-Woods:
This PR includes the code for the outbound `sync-async` workflow, most importantly the approach is simply to invoke the appropriate asynchronous workflow, then if an acknowledgement is returned a resynchroniser is invoked to retrieve the async response from the async-async store. This has been tested locally in a modified integration test and returns a success response. 

There is one notable change to the other async workflows, the `handle_outbound_message` now takes an additional `WorkDescription` type parameter, this is so that the workflow of the work description can be set to `sync-async` in the sync-async workflow. It is expected that, as previously was done, the work description would be created by if a work description instance is not passed in


docker-compose.yml
```diff
@@ -25,8 +25,11 @@ services:
       - MHS_LOG_LEVEL=NOTSET
       - MHS_STATE_TABLE_NAME=mhs_state
       - MHS_DYNAMODB_ENDPOINT_URL=http://dynamodb:8000
+      - MHS_SYNC_ASYNC_STATE_TABLE_NAME=sync_async_state
```
ghost, line 28: Do you not need to add the equivalents of these to the task definitions in the terraform file?

mhs/common/mhs_common/state/work_description.py
```diff
@@ -155,6 +161,18 @@ def __init__(self, persistence_store: pa.PersistenceAdaptor, store_data: dict):
         logger.info('016', 'Successfully updated work description to state store for {key}', {'key': self.message_key})
         return old_data
 
+    async def update(self):
+        """
+        Retrieves the copy of the data from the store with the self.message_key key, this is an overwrite so
+        should be used to care not to remove local data
```
ghost: This line doesn't make sense to me.

mhs/common/mhs_common/workflow/sync_async.py
```diff
@@ -23,29 +21,80 @@ class SyncAsyncWorkflow(common_synchronous.CommonSynchronousWorkflow):
...
+
+        logger.info('0001', 'Entered sync-async workflow to handle outbound message')
+        wdo = wd.create_new_work_description(self.work_description_store, message_id,
+                                             wd.MessageStatus.OUTBOUND_MESSAGE_RECEIVED,
+                                             workflow.SYNC_ASYNC)
+
+        status_code, response = await async_workflow.handle_outbound_message(message_id, correlation_id,
+                                                                             interaction_details, payload, wdo)
+        if not (status_code == 202):
+            logger.info('0002', 'No ACK received ')
```
ghost: This should probably be a warning level log, shouldn't it?

mhs/common/mhs_common/workflow/sync_async.py
```diff
@@ -23,29 +21,80 @@ class SyncAsyncWorkflow(common_synchronous.CommonSynchronousWorkflow):
...
+            return status_code, response, wdo
+
+        status_code, response = await self._retrieve_async_response(message_id, wdo)
+        return status_code, response, wdo
+
+    async def _retrieve_async_response(self, message_id, wdo: wd.WorkDescription):
+        logger.info('0005', 'Attempting to retrieve the async response from the async store')
+        try:
+            response = await self.resynchroniser.pause_request(message_id)
+            log.correlation_id.set(response[CORRELATION_ID])
```
ghost: I'm not sure I understand why we'd change the correlation ID here?

mhs/common/mhs_common/workflow/sync_async.py
```diff
@@ -23,29 +21,80 @@ class SyncAsyncWorkflow(common_synchronous.CommonSynchronousWorkflow):
...
+            logger.info('0003', 'Retrieved async response from sync-async store, set correlation ID')
+            await self._update_state_store_success_retriaval(wdo)
+            return 200, response[MESSAGE_DATA]
+        except sync_async_resynchroniser.SyncAsyncResponseException as e:
+            logger.error('0004', 'No async response placed on async store within timeout for {messageId}',
+                         {'messageId': message_id})
+            return 500, "No async response received from sync-async store"
+
+    async def _update_state_store_success_retriaval(self, wdo):
```
ghost: Typo in the method name:
```suggestion
    async def _update_state_store_success_retrieval(self, wdo):
```

mhs/common/mhs_common/workflow/sync_async.py
```diff
@@ -23,29 +21,80 @@ class SyncAsyncWorkflow(common_synchronous.CommonSynchronousWorkflow):
...
+        attempts = 0
+        while attempts < 4:
```
ghost: It looks like `_add_to_sync_async_store` below uses `self.sync_async_store_max_retries` to determine how many times it should attempt to write to the sync/async store. Could you do the same here? Or at least use a configurable value in the same way?

mhs/common/mhs_common/workflow/sync_async_resynchroniser.py
```diff
@@ -0,0 +1,42 @@
...
+    async def pause_request(self, message_id: str) -> dict:
+        time_waited = 0
+        logger.info('001', 'Beginning async retrieval from sync-async store')
+        while time_waited < self.retry_timeout:
+            item = await self.sync_async_store.get(message_id)
+            if item is not None:
+                logger.info('002', 'Message found in sync-async store, ending polling')
+                return item
+            logger.info('003', 'Failed to find async response after {time}s', {'time': time_waited})
+            await tornado.gen.sleep(self.retry_interval)
```
ghost: Does this need to be the tornado-specific sleep method?

mhs/common/mhs_common/workflow/sync_async_resynchroniser.py
```diff
@@ -0,0 +1,42 @@
...
+
+            time_waited += self.retry_interval
```
ghost: I'm not sure how big an issue this really is, but you're assuming here that the rest of this method (including the `self.sync_async_store.get` call) took no time at all. Is that a deliberate simplification? If not, you could base this on timestamps, or perhaps just a number of retries (that might simplify things)?

mhs/outbound/main.py
```diff
@@ -31,16 +31,27 @@ def load_party_key(data_dir: pathlib.Path) -> str:
 
 
 def initialise_workflows(transmission: outbound_transmission.OutboundTransmission, party_key: str,
-                         persistence_store: persistence_adaptor.PersistenceAdaptor) \
+                         persistence_store: persistence_adaptor.PersistenceAdaptor,
```
ghost: For consistency with your change below:
```suggestion
                         work_description_store: persistence_adaptor.PersistenceAdaptor,
```
(Remember to update the docstring parameter name too!)

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -47,18 +46,41 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
                                         reason=f'Unknown interaction ID: {interaction_id}')
 
         try:
-            workflow = self.workflows[interaction_details['workflow']]
+            wf = self.workflows[interaction_details['workflow']]
```
ghost, line 49: Not sure there's much benefit to this.

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -47,18 +46,41 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
...
+            status, response, wdo = await async_workflow.handle_sync_async_outbound_message(message_id, correlation_id,
+                                                                                            interaction_details, body,
+                                                                                            wf)
+            await self.return_sync_async_response(status, response, wdo)
+        else:
+            status, response = await wf.handle_outbound_message(message_id, correlation_id, interaction_details, body,
+                                                                None)
+            self._write_response(status, response)
+
+    def _is_sync_async(self, inteaction_details):
```
ghost: Typo in param name:
```suggestion
    def _is_sync_async(self, interaction_details):
```

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -47,18 +46,41 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
...
             logger.error('0008', "Weren't able to determine workflow for {InteractionId} . This likely is due to a "
                                  "misconfiguration in interactions.json", {"InteractionId": interaction_id})
             raise tornado.web.HTTPError(500,
                                         f"Couldn't determine workflow to invoke for interaction ID: {interaction_id}",
                                         reason=f"Couldn't determine workflow to invoke for interaction ID: "
-                                               f"{interaction_id}") from e
+                                        f"{interaction_id}") from e
+
+        if self._is_sync_async(interaction_details):
+            async_workflow: workflow.SyncAsyncWorkflow = self.workflows[workflow.SYNC_ASYNC]
```
ghost: Clearer variable name?:
```suggestion
            sync_async_workflow: workflow.SyncAsyncWorkflow = self.workflows[workflow.SYNC_ASYNC]
```

mhs/outbound/outbound/request/synchronous/tests/test_handler.py
```diff
@@ -47,7 +47,7 @@ def test_post_message(self, mock_message_id, mock_correlation_id, mock_get_uuid)
 
         self.config_manager.get_interaction_details.assert_called_with(INTERACTION_NAME)
         self.workflow.handle_outbound_message.assert_called_with(MOCK_UUID, MOCK_UUID_2, INTERACTION_DETAILS,
-                                                                 REQUEST_BODY)
```
ghost, line 50: You've added several bits of logic to this handler, so I'd expect to see additional test cases for them.

ghost (changes requested):

docker-compose.yml
Philip-Woods: I do, and this is one of the reasons the pipeline was failing previously, but these variables should have been added as a part of the last (`fix/int-tests`) PR.

Edit: On a second look the `MHS_RESYNC...` variables are new so I've added these to the TF/jenkins script

mhs/common/mhs_common/state/work_description.py
Philip-Woods: Again should have been addressed in #73

mhs/common/mhs_common/workflow/sync_async.py
Philip-Woods: Yeah good point, I guess my thought was it should match whatever the inbound is set to but in reality they should both be the same anyway so I'll remove it

mhs/common/mhs_common/workflow/sync_async.py
Philip-Woods: So this should now be solved as a part of #73, there's a configurable retry value for any any interactions with the state store

mhs/common/mhs_common/workflow/sync_async_resynchroniser.py
Philip-Woods: updated to used the `asyncio` sleep method

mhs/common/mhs_common/workflow/sync_async_resynchroniser.py
Philip-Woods: Yeah I discussed this with Ismail and we decided this way was probably OK, but on second thoughts I think in other places in the system we've used number of retries so I've changed it to be consistent with that

mhs/outbound/outbound/request/synchronous/handler.py
Philip-Woods: There's an imported module called `workflow`, felt wrong to be overwriting it

mhs/outbound/outbound/request/synchronous/handler.py
Philip-Woods: Actually I've just deleted this method since `interaction_details.get('sync-async')` provides the same functionality without the extra method

mhs/outbound/outbound/request/synchronous/tests/test_handler.py
Philip-Woods: Added test cases to ensure the state is being updated appropriately based on whether or not the response was sent to the supplier system, and it's a little bit of repetition but I've added tests for checking the sync-async workflow is invoked correctly depending on the interactions

mhs/common/mhs_common/workflow/sync_async.py
```diff
@@ -37,9 +37,10 @@ def __init__(self,
         self.sync_async_store = sync_async_store
         self.work_description_store = work_description_store
         self.resynchroniser = resynchroniser
-        self.sync_async_store_max_retries = sync_async_store_max_retries
+        self.sync_async_store_max_retries = persistence_store_max_retries
         self.sync_async_store_retry_delay = sync_async_store_retry_delay / 1000 if sync_async_store_retry_delay \
             else None
+        self.persistence_store_retries = persistence_store_max_retries
```
ghost, line 42: A fairly minor point, bout you could probably just have a single member variable to store `persistence_store_max_retries` now, couldn't you?

mhs/common/mhs_common/workflow/sync_async_resynchroniser.py
```diff
@@ -14,29 +14,29 @@ class SyncAsyncResponseException(Exception):
...
 
     def __init__(self, sync_async_store: persistence_adaptor.PersistenceAdaptor,
-                 retry_timeout: int,
+                 max_retries: int,
                  retry_interval: float
                  ):
         """
         :param sync_async_store: The store where the sync-async messages are placed from the inbound service
-        :param retry_timeout: The total time to poll the sync-async store for a async response in seconds
+        :param max_retries: The total time to poll the sync-async store for a async response in seconds
```
ghost: This docstring still talks about this being the total time.

mhs/common/mhs_common/workflow/sync_async_resynchroniser.py
```diff
@@ -14,29 +14,29 @@ class SyncAsyncResponseException(Exception):
...
-        while time_waited < self.retry_timeout:
+        while retries < self.max_retries:
             item = await self.sync_async_store.get(message_id)
             if item is not None:
                 logger.info('002', 'Message found in sync-async store, ending polling')
                 return item
-            logger.info('003', 'Failed to find async response after {time}s', {'time': time_waited})
-            await tornado.gen.sleep(self.retry_interval)
+            logger.warning('003', f'Failed to find async response after {retries} of {self.max_retries}')
+            await asyncio.sleep(self.retry_interval)
```
ghost: Won't this perform an unnecessary sleep after the last retry?

mhs/common/mhs_common/workflow/sync_async_resynchroniser.py
```diff
@@ -14,29 +14,29 @@ class SyncAsyncResponseException(Exception):
...
 
-            time_waited += self.retry_interval
-        logger.error('004', 'Resync timer exceeded, waited {time}s', {'time': time_waited})
+            retries += 1
+        logger.error('004', 'Resync timer exceeded, waited {time}s', {'time': retries})
```
ghost: This log message still references time, rather than a number of retries.

mhs/common/mhs_common/workflow/tests/test_resynchroniser.py
```diff
@@ -23,7 +23,7 @@ class TestSyncAsyncReSynchroniser(TestCase):
         result = await resynchroniser.pause_request('Message')
         self.assertTrue(result)
 
-    @patch('tornado.gen.sleep')
+    @patch('asyncio.sleep')
     @test_utilities.async_test
     async def test_resync_no_response(self, sleep_mock):
         store = MagicMock()
```
ghost, line 29: As discussed in my other comment, if you're hitting the max number of retries, this method should verify that `store.get` was called 20 times, but `sleep_mock` should only be called 19.

mhs/outbound/outbound/request/synchronous/tests/test_handler.py
```diff
@@ -164,3 +168,54 @@ def test_post_with_no_body(self):
...
+        expected_response = "Hello world!"
+        wdo = unittest.mock.MagicMock()
+        wdo.set_outbound_status.return_value = test_utilities.awaitable(True)
+        result = test_utilities.awaitable((200, expected_response, wdo))
+
+        self.sync_async_workflow.handle_sync_async_outbound_message.return_value = result
+        self.config_manager.get_interaction_details.return_value = {'sync-async': True, 'workflow': WORKFLOW_NAME}
+
+        self.fetch("/", method="POST", headers={"Interaction-Id": INTERACTION_NAME}, body=REQUEST_BODY)
+        self.sync_async_workflow.handle_sync_async_outbound_message.assert_called_once()
```
ghost, line 208: I think it would probably be worth also verifying that the expected response is returned from the handler here (and in `test_sync_async_workflow_not_invoked` below), as that's the other part of its responsibilities.

mhs/common/mhs_common/workflow/tests/test_sync_async.py
```diff
@@ -1,66 +1,78 @@
...
+
+        await self.workflow.handle_sync_async_outbound_message('id123', 'cor123', {},
+                                                               'payload',
+                                                               async_workflow
+                                                               )
+        wd_mock.assert_called_with(self.work_description_store,
+                                   'id123',
+                                   workflow.SYNC_ASYNC,
+                                   outbound_status=wd.MessageStatus.OUTBOUND_MESSAGE_RECEIVED)
+        async_workflow.handle_outbound_message.assert_called_once()
```
ghost, line 45: We should check the return values here, and also verify that the async workflow is called with the expected arguments.

mhs/common/mhs_common/workflow/tests/test_sync_async.py
```diff
@@ -1,66 +1,78 @@
...
+                                                                                       'payload',
+                                                                                       async_workflow
+                                                                                       )
+        self.assertEqual(status, 500)
+        self.assertEqual("Failed to reach spine", response)
 
-        self.assertFalse(is_async)
-        self.assertIs(sentinel.message, actual_response)
+    @test_utilities.async_test
+    async def test_resync_failure(self):
```
ghost, line 73: I think it might be having an equivalent of this test case to verify that `_retrieve_async_response` updates the WDO with the expected state when the resynchroniser returns a message successfully (or the appropriate assertion could be added to the happy path test case above).

Philip-Woods: Added it to the happy path test, this also made me aware that I wasn't checking the status code correctly either so I corrected that in the test as well

mhs/common/mhs_common/workflow/sync_async_resynchroniser.py
```diff
@@ -0,0 +1,44 @@
...
+            if item is not None:
+                logger.info('002', 'Message found in sync-async store, ending polling')
+                return item
+            logger.warning('003', f'Failed to find async response after {retries} of {self.max_retries}')
+            retries += 1
+
+            if not (retries == self.max_retries):
+                await asyncio.sleep(self.retry_interval)
+
+        logger.error('004', 'Resync retries exceeded, attempted {retry}', {'retry': retries})
```
ghost: Might be a little clearer:
```suggestion
        logger.error('004', 'Resync retries exceeded, attempted {retries}', {'retries': retries})
```

ghost (dismissed):
Looks good. Just one (non-essential) suggestion to make a log message a bit clearer.

ghost (approved):

PR #75, opened 2019-09-03
RT-159: Spine SOAP:Fault handling

anstm:
Now system is able to handle SOAP:Fault(s) which are coming from Spine in context of outbound service's Asynchronous Express workflow.


mhs/common/mhs_common/messages/soap_fault.py
```diff
@@ -0,0 +1,80 @@
...
+ERR_CODE_CONTEXT = 'codeContext'
+
+ERR_FIELDS = [ERR_CODE, ERR_SEVERITY, ERR_LOCATION, ERR_DESCRIPTION, ERR_CODE_CONTEXT]
+
+
+def _extract_tag_text(elem: ElementTree.ElementTree, path: AnyStr) -> AnyStr:
+    try:
+        return elem.find(path, NS).text
+    except AttributeError as e:
+        # Log error here
```
Philip-Woods: Looks like a unfinished `TODO`

mhs/common/mhs_common/messages/tests/test_messages/soapfault_request.xml
```diff
@@ -0,0 +1,20 @@
...
+      <detail>
+        <nasp:errorList xmlns:nasp="http://national.carerecords.nhs.uk/schema/">
+          <nasp:error>
+            <nasp:codeContext>urn:nhs:names:error:tms</nasp:codeContext>
+            <nasp:errorCode>200</nasp:errorCode>
+            <nasp:severity>Error</nasp:severity>
+            <nasp:location>Not Supported</nasp:location>
+            <nasp:description>System failure to process message - default</nasp:description>
+          </nasp:error>
+        </nasp:errorList>
```
Philip-Woods: Do any of these tests cover multiple soap errors? Might be worth adding a multi-error test case just to ensure multiple errors are being handled correctly

mhs/common/mhs_common/workflow/asynchronous_express.py
```diff
@@ -58,7 +68,7 @@ def __init__(self, party_key: str = None, persistence_store: persistence_adaptor
             self._record_outbound_audit_log(timing.get_time(), start_time,
                                             wd.MessageStatus.OUTBOUND_MESSAGE_NACKD)
             await wdo.set_status(wd.MessageStatus.OUTBOUND_MESSAGE_NACKD)
-            return 500, 'Error received from Spine'
+            return AsynchronousExpressWorkflow.handle_spine_fault(e, logger)
```
Philip-Woods: Optional: I would maybe keep the `500` response here and have `handle_spine_fault` just return the text value, just seems a little strange to me that what is effectively string parsing has knowledge about the appropriate HTTP error code as well. 

Though I just noticed in the error parsing you're checking for a 500 anyway so maybe it doesn't matter too much, I'll leave it as optional/ a general comment since I'm not sure on the best approach here

mhs/common/mhs_common/workflow/common.py
```diff
@@ -34,3 +40,31 @@ class CommonWorkflow(abc.ABC):
         :param payload: payload to handle
         """
         pass
+
+    @staticmethod
+    def handle_spine_fault(error: httpclient.HTTPClientError,
```
Philip-Woods: Minor/optional: Seems more appropriate to name this more generally as some form of outbound fault since I guess these errors could also arise when sending to a non-spine MHS, we don't currently support this though so it's up to you.

mhs/common/mhs_common/workflow/common.py
```diff
@@ -34,3 +40,31 @@ class CommonWorkflow(abc.ABC):
...
+            assert 'Content-Type' in response_headers
+            if response_headers['Content-Type'] == 'text/xml':
+                parsed_body = ElementTree.fromstring(response_body)
+
+                if SOAPFault.is_soap_fault(parsed_body):
+                    fault: SOAPFault = SOAPFault.from_parsed(response_headers, parsed_body)
+
+                    for error in fault.error_list:
+                        err_text = ', '.join([f'{k}={v}' for k, v in error.items()])
+                        logger.error('0010', f'SOAP Fault returned from Spine: {err_text}')
```
Philip-Woods: This number of indented levels usually jumps out to me as something that would be appropriate for being broken into further methods which maybe applies here.

mhs/common/mhs_common/workflow/common.py
```diff
@@ -34,3 +40,31 @@ class CommonWorkflow(abc.ABC):
...
+
+        return 500, client_message
```
Philip-Woods: This `client_message` seems to be fixed right? The ticket mentions that "[The fault]....is logged and presented back to the client", so I'm wondering if this `client_message` should contain the soap faults rather than a generic message?

mhs/common/mhs_common/workflow/common.py
```diff
@@ -34,3 +40,31 @@ class CommonWorkflow(abc.ABC):
...
+        :return: Response to external client represented as HTTP status code and body
+        """
+        response_code = error.response.code
+        response_headers = error.response.headers
+        response_body = error.response.body
+        client_message = 'Error(s) returned from Spine. Contact system administrator.'
+
+        if response_code == 500:
+            assert 'Content-Type' in response_headers
+            if response_headers['Content-Type'] == 'text/xml':
```
Philip-Woods: Will we ever cover the condition where it's not `text/xml`? Just wondering if it's always this or if we need to be looking at other header types as well

mhs/common/mhs_common/messages/tests/test_soapfault_request.py
```diff
@@ -0,0 +1,19 @@
+import os
+from xml.etree import ElementTree
+from unittest import TestCase
+
+from mhs_common.messages.soap_fault import SOAPFault, NS
+
+
+class TestSOAPFaultRequest(TestCase):
+    def test_is_soap_fault(self):
+        with open(os.path.join('test_messages', 'soapfault_request.xml')) as f:
```
Philip-Woods: Couple things: There's a method in the `file_utilities` module that we've been using for file reads, and generally throughout the codebase we've been using the `definitions.ROOT_DIR` variable as the top level reference point for file paths

mhs/common/mhs_common/messages/tests/test_soapfault_request.py
```diff
@@ -0,0 +1,19 @@
...
+            message = f.read()
+
+        self.assertTrue(SOAPFault.is_soap_fault(ElementTree.fromstring(message)))
+
+        with open(os.path.join('test_messages', 'ebxml_header.xml')) as f:
+            message = f.read()
+
+        self.assertFalse(SOAPFault.is_soap_fault(ElementTree.fromstring(message)))
+        self.assertFalse(SOAPFault.is_soap_fault(None))
```
Philip-Woods: Minor/Optional: There's several arrange and assert steps here so it might be worth breaking these into subtests so it's a bit more atomic, but I'm not hugely fussed as there's no shared resources

mhs/common/mhs_common/workflow/common.py
Philip-Woods: Also are there any tests/does there need to be any tests around this method and the extra classmethods in `SOAPFault`? Only test I can spot is for `is_soap_fault`

mhs/common/mhs_common/workflow/common.py
Philip-Woods: Also because you're passing the logger in here there's a possibility that you're going to log two distinct log message types both with the same `{ProcessTag}0010` process keys, I would suggest one of:
- using a very random process key
- allowing the calling workflow to pass the process number into this method,
- or just having this module define it's own logging object. 

I wondering if maybe defining a new logging object is the best, unless there was a specific reason you wanted to pass it in to this method?

mhs/common/mhs_common/workflow/common.py
```diff
@@ -34,3 +40,33 @@ class CommonWorkflow(abc.ABC):
...
+                parsed_body = ElementTree.fromstring(response_body)
+                default_fields =  {'errorType': 'ebxml_error'}
+
+                if SOAPFault.is_soap_fault(parsed_body):
+                    fault: SOAPFault = SOAPFault.from_parsed(response_headers, parsed_body)
+
+                    for error in fault.error_list:
+                        all_fields = {**error, **default_fields}
+                        err_text = ', '.join([f'{k}={v}' for k, v in all_fields.items()])
```
Philip-Woods: Optional: The logger actually manages this formatting for you, if you check elsewhere you'll be to see how it's used, it might be neater to use that

Philip-Woods:
Also there's a couple unit tests failing on Jenkins as well

mhs/common/mhs_common/messages/tests/test_messages/soapfault_request.xml
anstm: Tests added.

mhs/common/mhs_common/workflow/asynchronous_express.py
anstm: Changed.

mhs/common/mhs_common/workflow/common.py
anstm: Name changed, tests added.

mhs/common/mhs_common/messages/tests/test_soapfault_request.py
anstm: Now tests use `file_utilities`

mhs/common/mhs_common/messages/tests/test_soapfault_request.py
anstm: Done.

mhs/common/mhs_common/workflow/common.py
anstm: Changed.

mhs/common/mhs_common/workflow/common.py
anstm: Changed.

mhs/common/mhs_common/errors/soap_handler.py
```diff
@@ -0,0 +1,40 @@
...
+    :param headers: NHS response headers
+    :param body: NHS response body
+    :return: Response to external client represented as HTTP status code and body
+    """
+    if code != 500:
+        logger.warning('0001', 'Not HTTP 500 response. {Code} {Body}', {'Code': code, 'Body': body})
+        return code, body
+
+    assert 'Content-Type' in headers, 'No Content-Type header in Spine response, response cannot be handled!'
+    assert headers['Content-Type'] == 'text/xml', 'Unexpected Content-Type {}!'.format(headers['Content-Type'])
```
Philip-Woods: I'd maybe suggest that `assert` might not be appropriate here, the [guidelines](http://google.github.io/styleguide/pyguide.html#244-decision) suggest assert should be used to check internal correctness and not values passed in as parameters. As this is a suggestion from the guidelines I won't hold up the PR based on this.

mhs/common/mhs_common/errors/tests/test_soap_handler.py
```diff
@@ -0,0 +1,41 @@
...
+        with self.assertRaises(AssertionError):
+            handle_soap_error(500, {'Content-Type': 'text/xml'}, '<a><b></b></a>')
+
+    def test_invalid_xml(self):
+        with self.assertRaises(Exception):
+            handle_soap_error(500, {'Content-Type': 'text/xml'}, '<a><b><b></a>')
+
+    def test_single_error(self):
+        message = FileUtilities.get_file_string(Path(self.message_dir) / 'soapfault_response_single_error.xml' )
+        self.assertTrue(' ' in handle_soap_error(500, {'Content-Type': 'text/xml'}, message)[1])
```
Philip-Woods: What's this checking? Seems to just be asserting there's a space rather than the error content? I'll consider this optional since we have the `test_multiple_errors` that covers a similar area

mhs/common/mhs_common/errors/__init__.py
```diff
@@ -0,0 +1,3 @@
+ERROR_RESPONSE_DEFAULTS = {
+    'errorType': 'ebxml_error'
+}
```
Philip-Woods: New line here

mhs/common/mhs_common/errors/soap_handler.py
```diff
@@ -0,0 +1,40 @@
...
+
+logger = IntegrationAdaptorsLogger('SOAP_ERROR_HANDLER')
+
+
+def handle_soap_error(code: int, headers: Dict, body: AnyStr) -> Tuple[int, AnyStr]:
+    """
+    Analyzes response from NHS which works in as web service mode
+    and returns result of interpretation to external client
+
+    :param code: NHS response code
```
Philip-Woods: Think this is a HTTP response code, not something specific to the NHS, unless I've misunderstood?

Philip-Woods (dismissed):
A few notes/questions on it but nothing I feel like should hold up a merge so should be good. 
Jenkins looks good apart from the integration tests, might have to get @gareth-allan to allow a merge for this while the integration tests are still being fixed

mhs/common/mhs_common/errors/tests/test_soap_handler.py
anstm: Changed.

mhs/common/mhs_common/errors/__init__.py
anstm: Done.

mhs/common/mhs_common/errors/soap_handler.py
anstm: Changed to Raise.

mhs/common/mhs_common/errors/soap_handler.py
anstm: Changed.

Philip-Woods (dismissed):

Philip-Woods (approved):

PR #76, opened 2019-09-09
Test/RT-149: Sync-Async Integration test

Philip-Woods:
The primary goal for this PR was to add an integration test for the sync-async wrapper around the async express message pattern. This has been added with the functionality to  wrap any given message in the integration tests with the sync-async header.

As a side effect I had to update the outbound handler to take a `sync-async` HTTP header to allow a supplier to decide if an asynchronous message should be processed as sync-async. A minor refactor of the outbound handler has been included for neatness


integration-tests/integration_tests/helpers/interactions.py
```diff
@@ -42,6 +43,11 @@ def call_mhs(mhs_command, hl7payload, message_id, pass_message_id, correlation_i
     if pass_correlation_id:
         headers['Correlation-Id'] = correlation_id
 
+    if sync_async:
+        headers['sync-async'] = 'true'
```
ismail-s: V. optional/v. minor: can do this as a single line if you want: `headers['sync-async'] = 'true' if sync_async else 'false'`

integration-tests/integration_tests/test_scenarios/int_async_express_tests.py
```diff
@@ -65,3 +67,24 @@ def test_async_express_inbound_patient(self):
         patient_number = methods.get_section(inbound_response, 'extension', 'id', 'patient')
         self.assertEqual(patient_number, '9689177923',
                          "Async Express inbound response test failed")
+
+
+class TestSyncAsyncWrapper(TestCase):
+
+    def test_async_express_sync_async_wrap(self):
+        # the response is just an acknowledgement from spine...
```
ismail-s: I'm guessing that you mean that the response from Spine for this interaction is an acknowledgement.

mhs/common/data/interactions/interactions.json
```diff
@@ -12,7 +12,7 @@
     "ack_requested": false,
     "ack_soap_actor": "urn:oasis:names:tc:ebxml-msg:actor:toPartyMSH",
     "sync_reply": true,
-    "sync-async": false
+    "sync_async": true
```
ismail-s, line 15: Just double-checking this, but for this interaction, is setting this to true a hack or what we'd actually want it to be? If it's a hack for the tests for the timebeing (until more interactions are added), then can you stick a comment in the task for adding more interactions that this should be switched back?

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -74,10 +53,28 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
...
+
     def write_error(self, status_code: int, **kwargs: Any):
         self.set_header('Content-Type', 'text/plain')
         self.finish(f'{status_code}: {self._reason}')
 
+    def _extract_sync_async_header(self):
+        sync_async_header = self.request.headers.get('sync-async', None)
+        if not sync_async_header:
+            logger.warning('0031', 'Failed to parse sync-async header from message')
+            raise tornado.web.HTTPError(400, 'Sync-Async header missing', reason='Sync-Async header missing')
```
ismail-s, line 64: Optional: I think it might be better to have this as optional (though I think you could argue either way). My only reason is that for the Sync workflow, you'd always have to pass this header, even though it wouldn't be used and could never be set to true. And actually, the same argument would apply to the async workflows for interactions that don't support sync-async.

integration-tests/integration_tests/test_scenarios/int_async_express_tests.py
```diff
@@ -65,3 +67,28 @@ def test_async_express_inbound_patient(self):
...
+                                                                        sync_async=True)
+        self.assertTrue(methods.check_status_code(outbound_response, 200),
+                        "Async Express outbound test failed")
+
+        root = ET.ElementTree(ET.fromstring(outbound_response.text)).getroot()
+
+        self.assertTrue(self.does_contain_success_code(root))
+
+    def does_contain_success_code(self, root):
+        for child in root.iter():
```
ismail-s: Couldn't you use the find methods mentioned [here](https://docs.python.org/3.7/library/xml.etree.elementtree.html#finding-interesting-elements)?

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -111,6 +108,57 @@ def _extract_interaction_id(self):
...
+            return False
+
+    async def _invoke_sync_async(self, message_id, correlation_id, interaction_details, body, async_workflow):
+        sync_async_workflow: workflow.SyncAsyncWorkflow = self.workflows[workflow.SYNC_ASYNC]
+        status, response, wdo = await sync_async_workflow.handle_sync_async_outbound_message(message_id,
+                                                                                             correlation_id,
+                                                                                             interaction_details,
+                                                                                             body,
+                                                                                             async_workflow)
+        await self.return_sync_async_response(status, response, wdo)
```
ismail-s, line 142: Optional: would be clearer if this method was moved down here, as it's only used here.

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -111,6 +108,57 @@ def _extract_interaction_id(self):
...
+                                        f"{interaction_id}") from e
+        return wf
+
+    def _extract_sync_async_from_interaction_details(self, interaction_details):
+        is_sync_async = interaction_details.get('sync_async')
+        if is_sync_async is None:
+            logger.error('0032', 'Failed to retrieve sync-async flag from interactions.json')
+            raise tornado.web.HTTPError(500, f'Failed to parse sync-async flag from interactions.json file',
+                                        reason='Failed to parse sync-async flag from interactions file for this '
+                                               'interaction, contact your administrator')
```
ismail-s: The people who see this error message should not be any end-users (by end-users I mean non-tech people), so I don't think we need to get too thingy with these error responses. As long as it's clear in the error response that the error is inside MHS and inside the `interactions.json` file (and I think this can be made clearer).

mhs/outbound/outbound/request/synchronous/tests/test_handler.py
```diff
@@ -215,10 +230,83 @@ def test_sync_async_workflow_not_invoked(self):
...
+
+        self.assertEqual(response.code, 400)
+        self.assertIn("Message header requested sync-async wrap for a message patternthat does not support sync-async",
+                      response.body.decode())
+        self.sync_async_workflow.handle_sync_async_outbound_message.assert_not_called()
+        self.workflow.handle_outbound_message.assert_not_called()
+
+    def test_not_sync_async_when_interactions_is_true_and_header_is_false(self):
+        self.setup_workflows()
+        self.config_manager.get_interaction_details.return_value = {'sync_async': 'true', 'workflow': WORKFLOW_NAME}
```
ismail-s: In some places, you're returning sync_async as `'true'`, in other places as `True`. Should be consistent. In the actual file you're using a boolean, so I think in these tests a boolean should also be being used, if that's what it's meant to be.

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -111,6 +108,57 @@ def _extract_interaction_id(self):
...
+            return True
+        elif (not interaction_config) and sync_async_header:
+            logger.error('0033', 'Message header requested sync-async wrap for un-supported sync-async')
+            raise tornado.web.HTTPError(400, f'Message header requested sync-async wrap for un-supported sync-async',
+                                        reason='Message header requested sync-async wrap for a message pattern'
+                                               'that does not support sync-async')
+        elif interaction_config and (not sync_async_header):
+            return False
+        else:
+            return False
```
ismail-s, line 133: Looking at the code coverage report, this line isn't hit. Also, the last 2 cases are the same and can be combined.

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -111,6 +108,57 @@ def _extract_interaction_id(self):
...
+            raise tornado.web.HTTPError(500, f'Failed to parse sync-async flag from interactions.json file',
+                                        reason='Failed to parse sync-async flag from interactions file for this '
+                                               'interaction, contact your administrator')
+        return is_sync_async
+
+    def _should_invoke_sync_async_workflow(self, interaction_config, sync_async_header):
+        if interaction_config and sync_async_header:
+            return True
+        elif (not interaction_config) and sync_async_header:
+            logger.error('0033', 'Message header requested sync-async wrap for un-supported sync-async')
```
ismail-s, line 128: I feel like this should be WARNING instead of ERROR. Not 100% certain though, can always defer back to @dshipp-bjss for his thoughts.

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -125,5 +173,15 @@ def _write_response(self, status: int, message: str) -> None:
         self.set_header("Content-Type", content_type)
         self.write(message)
 
+    def _retrieve_interaction_details(self, interaction_id):
+        interaction_details = self._get_interaction_details(interaction_id)
+
+        if interaction_details is None:
+            logger.error('0007', 'Unknown {InteractionId} in request', {'InteractionId': interaction_id})
```
ismail-s, line 178: I feel like this should be WARNING instead of ERROR. Same as other comment.

integration-tests/integration_tests/test_scenarios/int_async_express_tests.py
Philip-Woods: That was just a carry over from the copy/paste of the async test, it's not relevant now so I'll delete it

mhs/common/data/interactions/interactions.json
Philip-Woods: As I understand this is actually what it will be, interactions will be listed as whether or not they can be wrapped as sync async then the supplier will have the choice in the header

integration-tests/integration_tests/test_scenarios/int_async_express_tests.py
Philip-Woods: I initially tried `find` and was experimenting whilst waiting for review but it's returning no value. I _think_ it's because you have to provide the correct namespaces as a part of the search but I haven't managed to figure out the right input. I'll continue to experiment if I get time but I don't want to drain hours on this one check

mhs/outbound/outbound/request/synchronous/handler.py
Philip-Woods: As discussed this is staying as an error, and I've updated a few other log lines to be consistent with the idea that an error should be logged if handling of the message stops/faults and can't continue.

integration-tests/integration_tests/test_scenarios/int_async_express_tests.py
Philip-Woods: I continued to experiment with `find` and had no luck with the namespaces, but I found that using  `root.iter('{urn:hl7-org:v3}queryResponseCode')` will return the desired element which I think is  a reasonable compromise

mhs/outbound/outbound/transmission/outbound_transmission.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, certs_dir: str, client_cert: str, client_key: str, ca_certs:
                                 "max_retries": self._max_retries
                                 })
                 if retries_remaining <= 0:
-                    logger.warning("0004",
+                    logger.error("0004",
```
ismail-s, line 70: In `mhs/common/mhs_common/workflow/asynchronous_express.py` and `mhs/common/mhs_common/workflow/sync_async.py` there's another of these retries that can be changed to log at ERROR level instead of WARNING level.

integration-tests/integration_tests/test_scenarios/int_async_express_tests.py
```diff
@@ -82,13 +81,5 @@ def test_async_express_sync_async_wrap(self):
...
-        self.assertTrue(self.does_contain_success_code(root))
-
-    def does_contain_success_code(self, root):
-        for child in root.iter():
-            if child.tag == '{urn:hl7-org:v3}queryResponseCode':
-                self.assertEqual(child.attrib['code'], 'OK')
-                return True
-
-        return False
+        element = list(root.iter('{urn:hl7-org:v3}queryResponseCode'))[0]
```
ismail-s: I tried fiddling with this, and eventually figured it out (with some [Googling](https://stackoverflow.com/a/55049369)):
```suggestion
        element = root.find('.//hl7:queryResponseCode', namespaces={'hl7': 'urn:hl7-org:v3'})
```
Have to use XPath as without it, it's a non-recursive find.

ismail-s (dismissed):

ismail-s (approved):

PR #77, opened 2019-09-09
RT-179 AWS exemplar deployment

ismail-s:
- Create some Terraform scripts that do an AWS exemplar deployment of the MHS
  - These Terraform scripts rely on there already existing a supplier VPC and an Opentest VPC ()
  - Update the existing Terraform scripts used to run the integration tests to just deploy SCR
- Configure MHS outbound to:
  - Optionally use a HTTP proxy to route outbound transmission requests via (not requests to the MHS Spine route lookup service)
  - Always use [`curl_httpclient`](https://www.tornadoweb.org/en/branch6.0/httpclient.html). This is required to be able to do HTTP proxying. It also is faster ([according to the Tornado docs]((https://www.tornadoweb.org/en/branch6.0/httpclient.html))).
    - However, this means that we need a dependency on [`pycurl`](https://pypi.org/project/pycurl/). This shouldn't be an issue, except that pycurl/pycurl#569 means that there are no binary builds for Windows, and we can't use the previous release that does have binary builds as we're using Python 3.7 and pycurl only goes upto Python 3.6 (I don't think this is as much an issue as it sounds, as they basically do have Python 3.7 support according to pycurl/pycurl#549), and we do need Python 3.7 (eg we use [contextvars](https://docs.python.org/3.7/library/contextvars.html) which is only in Python 3.7). There is a workaround for Windows which is to use a binary wheel which their CI has built (I've put this in the code as a temporary, Windows-only hack). And even on Windows, our recommended approach of using Docker-compose still works.

# Notes
- No autoscaling has been done in this PR. Instead, the number of MHS tasks to create is specified as a variable when running Terraform.
- No health check has been added to the ECS task definition or as docker health checks. This is because ECS is already monitoring if the Docker entrypoint command fails. And the load balancers are already doing health checks, so I wasn't sure that we should be adding more health checks at another layer.
- I have raised the following JIRA techdebts/tasks/bugs off the back of RT-179:
  - [Support Pycurl on Windows for Python 64bit](https://gpitbjss.atlassian.net/browse/RT-210) (the workaround I've added to the code is for Python 32bit)
  - [Lock down (more) IAM role for Jenkins worker](https://gpitbjss.atlassian.net/browse/RT-208)
  - [Add HTTPS to outbound and route load balancers](https://gpitbjss.atlassian.net/browse/RT-209)
  - [MHS inbound load balancer logs not populating in S3](https://gpitbjss.atlassian.net/browse/RT-214)


Jenkinsfile
```diff
@@ -46,25 +48,100 @@ pipeline {
...
+                dir('pipeline/terraform/integration-test-environment') {
+                    sh label: 'Initialising Terraform', script: """
+                            terraform init \
+                            -backend-config="bucket=${TF_STATE_BUCKET}" \
+                            -backend-config="region=${TF_STATE_BUCKET_REGION}" \
+                            -backend-config="dynamodb_table=${TF_MHS_LOCK_TABLE_NAME}" \
+                            -input=false
+                        """
+                    sh label: 'Destroy any existing MHS services', script: """
+                            terraform destroy -auto-approve \
```
ismail-s: This command was added as, before it was here, ECS tasks from the previous build would take a while to get destroyed (ECS service would start the tasks for the new build and then get rid of the previous tasks over ~10 minutes), which resulted in the tasks from the previous build handling some requests. However, this command is taking ~6.5 minutes to run ie it slows the build down. We (me and @gareth-allan) believe this is to do with https://github.com/terraform-providers/terraform-provider-aws/issues/3414 ie we think this is a known issue. Nevertheless, with this command, the builds seem to be working correctly.

pipeline/terraform/integration-test-environment/lbs.tf
```diff
@@ -0,0 +1,112 @@
...
+
+resource "aws_lb_target_group" "outbound_alb_target_group" {
+  port = 80
+  protocol = "HTTP"
+  target_type = "ip"
+  vpc_id = aws_vpc.mhs_vpc.id
+
+  health_check {
+    path = "/"
+    matcher = "405"
```
ismail-s: The application load balancers do these health checks. Currently, they're doing them on the same endpoint that we make requests to MHS to. It would be cleaner if there were separate healthcheck endpoints that just respond with HTTP 200 without doing any further processing. I think we should maybe have a separate task/techdebt to do this.

@dshipp-bjss what do you think?

mhs/common/mhs_common/request/healthcheck_handler.py
```diff
@@ -0,0 +1,7 @@
+import tornado.web
+
+
+class HealthcheckHandler(tornado.web.RequestHandler):
```
dshipp-bjss, line 4: Needs comments to explain purpose

pipeline/terraform/mhs-environment/security-groups.tf
```diff
@@ -224,21 +290,34 @@ resource "aws_security_group" "sds_cache_security_group" {
...
+
+# MHS outbound load balancer security group
 resource "aws_security_group" "alb_outbound_security_group" {
   name = "Outbound ALB Security Group"
   description = "The security group used to control traffic for the outbound MHS Application Load Balancer."
   vpc_id = aws_vpc.mhs_vpc.id
 
+  # Allow inbound traffic from the supplier VPC. We don't make any
+  # assumptions here about the internal structure of the supplier VPC,
+  # instead just allowing inbound requests from the whole VPC.
```
dshipp-bjss, line 309: Worth adding a final sentence to this comment such as "A supplier could restrict this rule further by limiting access, for example to a specific Security Group."

dshipp-bjss (changes requested):

dshipp-bjss (approved):

dshipp-bjss (approved):

dshipp-bjss (approved):

PR #78, opened 2019-09-09
RT-158: Handle ebXML errors

anstm:
System is able to handle ebXML error coming from Spine with non 500 HTTP status (thus without HTTPClient exception). System checks response for specific ebXML format and logs all the errors returned.

PR #84, opened 2019-09-12
Feature/rt 172 sync workflow

Philip-Woods:
This PR adds functionality for synchronous workflow based messages through the MHS, the integration test for the chosen interaction has also been enabled. Notable changes to things outside of the workflow itself include additional parameters being retrieved from the spine route lookup service and workflows now take an additional parameter: the source ASID (know as `from_asid` in the code). This ASID is provided by a supplier via http header and passed to the sync workflow for populating the SOAP header


integration-tests/integration_tests/helpers/build_message.py
```diff
@@ -39,6 +40,7 @@ def build_message(template, asid, patient_nhs_number, payload):
         UUID: uuid,
         TIMESTAMP: timestamp,
         ASID: asid,
+        TO_ASID: 'S20001A000168',
```
ghost: This doesn't look like an ASID to me (it looks more like a CPA ID). Is this the correct value?

integration-tests/integration_tests/helpers/interactions.py
```diff
@@ -44,6 +46,7 @@ def call_mhs(mhs_command, hl7payload, message_id, pass_message_id, correlation_i
         headers['Correlation-Id'] = correlation_id
 
     headers['sync-async'] = 'true' if sync_async else 'false'
+    headers['from_asid'] = f'{from_asid}'
```
ghost: For consistency with other header names:
```suggestion
    headers['from-asid'] = f'{from_asid}'
```

integration-tests/integration_tests/test_scenarios/int_smoke_tests.py
```diff
@@ -25,18 +27,18 @@ def test_mhs_async_reliable(self):
                                                                    'REPC_IN150016UK05',
                                                                    '9446245796',
                                                                    'Asynchronous Reliable test')
-        self.assertTrue(methods.check_response(mhs_response,'requestSuccessDetail'),
+        self.assertTrue(methods.check_response(mhs_response, 'requestSuccessDetail'),
                         "Asynchronous Reliable smoke test failed")
 
     # Message Pattern Type: Synchronous
     # Interaction: PDS Retrieval Query (QUPA_IN040000UK32)
-    @skip('waiting for RT-15 - Implement MHS Synchronous Message Pattern')
```
ghost, line 33: I appreciate there's a separate ticket for implementing the integration tests, but just a heads-up that for that we'll need to have the equivalent of `int_async_express_tests.py` for synchronous requests (i.e. more than just this single happy path test case).

mhs/common/data/interactions/interactions.json
```diff
@@ -21,5 +21,15 @@
   "get_nhs_number": {
     "url": "https://msg.opentest.hscic.gov.uk/smsp/pds",
     "async_response_expected": false
+  },
+  "QUPA_IN040000UK32": {
+    "workflow": "sync",
+    "service": "urn:nhs:names:services:pdsquery",
+    "action": "QUPA_IN040000UK32",
+    "duplicate_elimination": false,
```
ghost: Are `duplicate_elimination`, `ack_requested`, `ack_soap_actor` or `sync_reply` actually required for this workflow? Aren't they ebXML attributes?

mhs/common/mhs_common/errors/soap_handler.py
```diff
@@ -24,7 +24,7 @@ def handle_soap_error(code: int, headers: Dict, body: AnyStr) -> Tuple[int, AnyS
     if 'Content-Type' not in headers:
         raise ValueError('No Content-Type header in Spine response, response cannot be handled!')
 
-    if headers['Content-Type'] == 'text/xml':
+    if headers['Content-Type'] != 'text/xml':
```
ghost: Are we missing test cases here? I'd have expected this case to be tested for (and the issue here to have broken the happy path test case).

mhs/common/mhs_common/state/work_description.py
```diff
@@ -21,11 +21,15 @@ class MessageStatus(str, enum.Enum):
...
     OUTBOUND_SYNC_ASYNC_MESSAGE_SUCCESSFULLY_RESPONDED = 'OUTBOUND_SYNC_ASYNC_MESSAGE_SUCCESSFULLY_RESPONDED'
+    OUTBOUND_MESSAGE_RESPONSE_RECEIVED = 'OUTBOUND_MESSAGE_RESPONSE_RECEIVED'
     INBOUND_RESPONSE_RECEIVED = 'INBOUND_RESPONSE_RECEIVED'
     INBOUND_RESPONSE_SUCCESSFULLY_PROCESSED = 'INBOUND_RESPONSE_SUCCESSFULLY_PROCESSED'
     INBOUND_RESPONSE_FAILED = 'INBOUND_RESPONSE_FAILED'
     INBOUND_SYNC_ASYNC_MESSAGE_STORED = 'INBOUND_SYNC_ASYNC_MESSAGE_STORED'
     INBOUND_SYNC_ASYNC_MESSAGE_FAILED_TO_BE_STORED = 'INBOUND_SYNC_ASYNC_MESSAGE_FAILED_TO_BE_STORED'
+    SYNC_RESPONSE_SUCCESSFUL = "SYNC_RESPONSE_SUCCESSFUL"
+    SYNC_RESPONSE_FAILED = "SYNC_RESPONSE_FAILED"
+
```
ghost, line 31: Unnecessary newline?

```suggestion
```

mhs/common/mhs_common/workflow/common.py
```diff
@@ -71,6 +74,24 @@ def __init__(self, routing: routing_reliability.RoutingAndReliability = None):
                            {'Exception': e})
             raise e
 
+    async def _lookup_to_asid_details(self, interaction_details: Dict):
```
ghost: This seems like unnecessary duplication. Couldn't you just add the call to `_extract_asid` into `_lookup_endpoint_details`? It would probably be worth refactoring it slightly to return an object/dictionary, rather than extending the tuple returned.

mhs/common/mhs_common/workflow/tests/test_common.py
```diff
@@ -78,3 +80,39 @@ def setUp(self):
...
+            common.CommonWorkflow._extract_asid(endpoint_details)
+
+    @async_test
+    async def test_extract_asid_multiple_returned(self):
+        expected_asid = "asid 1"
+        endpoint_details = {MHS_TO_ASID: [expected_asid, "asid 2"]}
+
+        actual_asid = common.CommonWorkflow._extract_asid(endpoint_details)
+
+        self.assertEqual(expected_asid, actual_asid)
```
ghost: Missing newline at the end of this file.

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -132,30 +134,47 @@ def _should_invoke_sync_async_workflow(self, interaction_config, sync_async_head
...
-                                                                  correlation_id,
-                                                                  interaction_details,
-                                                                  body,
-                                                                  None)
-        self._write_response(status, response)
+            if wdo:
+                await wdo.set_outbound_status(failure)
+
+    async def invoke_default_workflow(self, from_asid, message_id, correlation_id, interaction_details, body, wf):
+        if isinstance(wf, workflow.SynchronousWorkflow):
```
ghost: I don't like that this is doing different things for the sync workflow and the others. At the very least, you could just wrap the `write_response_with_store_updates` in the check for a sync workflow, and keep the call to `handle_outbound_message` the same for all workflows.

Ideally, we'd push the logic on whether/which states to write once a response is successful into the workflows. Perhaps define an abstract method on the base class that each subclass can provide an implementation of?

mhs/outbound/outbound/request/synchronous/tests/test_handler.py
```diff
@@ -309,3 +311,49 @@ def setup_workflows(self):
 
         self.sync_async_workflow.handle_sync_async_outbound_message.return_value = result
         self.workflow.handle_outbound_message.return_value = test_utilities.awaitable((200, "Success"))
+
+
+class TestSynchronousHandlerSyncMessage(tornado.testing.AsyncHTTPTestCase):
```
ghost, line 320: Does this need to be a separate test class?

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -132,30 +134,47 @@ def _should_invoke_sync_async_workflow(self, interaction_config, sync_async_head
...
+                                                     wd.MessageStatus.OUTBOUND_SYNC_ASYNC_MESSAGE_SUCCESSFULLY_RESPONDED,
+                                                     wd.MessageStatus.OUTBOUND_SYNC_ASYNC_MESSAGE_FAILED_TO_RESPOND)
 
-    async def return_sync_async_response(self, status: int, response: str, wdo: wd.WorkDescription):
+    async def write_response_with_store_updates(self, status: int, response: str, wdo: wd.WorkDescription,
+                                                success, failure):
         try:
             self._write_response(status, response)
-            await wdo.set_outbound_status(wd.MessageStatus.OUTBOUND_SYNC_ASYNC_MESSAGE_SUCCESSFULLY_RESPONDED)
+            if wdo:
```
ghost, line 151: You need some test cases to validate that this method handles not being passed a `wdo`.

ghost (changes requested):

mhs/common/mhs_common/workflow/synchronous.py
```diff
@@ -1,17 +1,116 @@
...
+                 transmission: transmission_adaptor.TransmissionAdaptor = None,
+                 persistence_store_max_retries: int = None,
+                 routing: routing_reliability.RoutingAndReliability = None):
+        self.party_key = party_key
+        self.wd_store = work_description_store
+        self.transmission = transmission
+        self.persistence_store_retries = persistence_store_max_retries
+        super().__init__(routing)
+
+    async def handle_outbound_message(self,
```
ghost, line 34: To meet acceptance criteria number 4:
```suggestion
    @timing.time_function
    async def handle_outbound_message(self,
```

mhs/common/mhs_common/workflow/tests/test_synchronous.py
```diff
@@ -0,0 +1,236 @@
...
+            await self.wf.handle_outbound_message(from_asid="202020", message_id="123",
+                                                  correlation_id="qwe",
+                                                  interaction_details={},
+                                                  payload="nice message", work_description_object=None)
+        except Exception:
+            # Don't care for exceptions, just want to check the store is set correctly first
+            pass
+
+        wd_mock.assert_called_with(self.wd_store, "123", workflow.SYNC,
+                                   work_description.MessageStatus.OUTBOUND_MESSAGE_RECEIVED)
```
ghost, line 51: Maybe worth asserting that `wdo.publish()` was called too?

mhs/common/mhs_common/workflow/synchronous.py
```diff
@@ -1,17 +1,116 @@
...
+                                             )
+        await wdo.publish()
+        if not from_asid:
+            return 400, '`from_asid` header field required for sync messages', None
+
+        try:
+            url, to_asid = await self._lookup_to_asid_details(interaction_details)
+        except Exception:
+            logger.error('009', 'Failed to retrieve details from spine route lookup')
+            await wdo.set_outbound_status(wd.MessageStatus.OUTBOUND_MESSAGE_PREPARATION_FAILED)
```
ghost: The async express workflow uses `OUTBOUND_MESSAGE_TRANSMISSION_FAILED` when it fails to look up endpoint details. We should be consistent here.

mhs/common/mhs_common/workflow/tests/test_synchronous.py
```diff
@@ -0,0 +1,236 @@
...
+    @async_test
+    async def test_send_message_failure(self, wd_mock):
+        wdo = mock.MagicMock()
+        wdo.publish.return_value = test_utilities.awaitable(None)
+        wd_mock.return_value = wdo
+        self.wf._lookup_to_asid_details = mock.MagicMock()
+        self.wf._lookup_to_asid_details.return_value = test_utilities.awaitable(("yes", "313123"))
+        self.wf._prepare_outbound_message = mock.MagicMock()
+        self.wf._prepare_outbound_message.return_value = test_utilities.awaitable(("123", {"qwe": "qwe"}, "message"))
+        wdo.set_outbound_status.return_value = test_utilities.awaitable(None)
```
ghost, line 114: Don't you need to make `self.transmission.make_request.return_value` raise an exception here?

mhs/common/mhs_common/workflow/tests/test_synchronous.py
```diff
@@ -0,0 +1,236 @@
...
+        self.transmission.make_request.return_value = test_utilities.awaitable(response)
+
+        error, text, work_description_response = await self.wf.handle_outbound_message(from_asid="202020",
+                                                                                       message_id="123",
+                                                                                       correlation_id="qwe",
+                                                                                       interaction_details={},
+                                                                                       payload="nice message",
+                                                                                       work_description_object=None)
+
+        wdo.set_outbound_status.assert_called_with(work_description.MessageStatus.OUTBOUND_MESSAGE_RESPONSE_RECEIVED)
```
ghost: Is this the correct status when an error is returned from Spine? The story says it's for when a 200 HTTP status is returned. I think this should be `OUTBOUND_MESSAGE_TRANSMISSION_FAILED`.

I think this test is passing because you need your transmission mock to be returning an `HTTPClientError` with the response code and body set, so we fall into the block that calls the SOAP error handling utility. As it is, you're treating this as a successful response, but with an HTTP error code. I don't think the Tornado HTTP client should do that.

mhs/common/mhs_common/workflow/tests/test_synchronous.py
```diff
@@ -0,0 +1,236 @@
...
+    @async_test
+    async def test_no_inbound(self):
+        with self.assertRaises(NotImplementedError):
+            await self.wf.handle_inbound_message("1", "2", mock.MagicMock(), "payload")
+
+    @async_test
+    async def test_prepare_message(self):
+        id, headers, message = await self.wf._prepare_outbound_message("message_id", "to_asid", "from_asid", "mesasge",
+                                                                       {'service': 'service', 'action': 'action'})
+        self.assertEqual(id, "message_id")
```
ghost, line 234: Again, probably worth checking the other properties of the message are set as expected too.

integration-tests/integration_tests/helpers/build_message.py
Philip-Woods: As discussed I've placed the Accredited system lookup UID into the dictionary with the other lookup details.

mhs/common/data/interactions/interactions.json
Philip-Woods: I've delete them for now, either myself or Colin will have to update it further (depending on who's PR goes in first) as his has some extra logic to remove the `action` element as well

mhs/common/mhs_common/errors/soap_handler.py
Philip-Woods: So this is a bug that is to be (fully) fixed/tested in Andrey's branch, I just needed the error to be printed correctly for debugging, do you want me to delete this part for this PR and let Andrey handle it in his PR?

mhs/outbound/outbound/request/synchronous/tests/test_handler.py
Philip-Woods: Possibly not, I set it as a different class because the magic mock for the 'default' workflow I needed it to inherit from sync workflow to do that `isinstance` check, I didn't really want to enforce that type on the other tests so I figured it should be a separate class

mhs/common/mhs_common/workflow/tests/test_synchronous.py
Philip-Woods: I think this passes because there's no async response on `make_request`, hence it errors anyway, but I'll add the exception anyway for completeness

integration-tests/integration_tests/helpers/methods.py
```diff
@@ -18,7 +18,7 @@ def get_asid():
         asid_file = str(Path(ROOT_DIR) / "integration_tests/data/certs/asid.txt")
         asid = FileUtilities.get_file_string(asid_file)
     except:
-        asid = None
+        asid = 918999199084
```
ghost: Why do we need to hardcode an ASID here?

mhs/spineroutelookup/lookup/sds_client.py
```diff
@@ -60,6 +61,9 @@ def __init__(self, sds_connection: ldap3.Connection, timeout: int = 3):
         response = accredited_system_lookup[0]
         party_key = response['attributes'][MHS_PARTY_KEY]
 
+        print(f"SDS_RESPONSE: {response}")
```
ghost:
```suggestion
```

mhs/spineroutelookup/lookup/tests/test_sds_client.py
```diff
@@ -58,18 +59,20 @@ class TestSDSClient(TestCase):
         client = mocks.mocked_sds_client()
 
         result = await client._accredited_system_lookup(ODS_CODE, INTERACTION_ID)
+
         self.assertIsNotNone(result)
         self.assertIsNotNone(result[0]['attributes']['nhsMHSPartyKey'] == 'A91461-9199084')
```
ghost: I think it's worth adding a check of the `uniqueIdentifier` value here:
```suggestion
         self.assertEqual(result[0]['attributes']['nhsMHSPartyKey'], 'A91461-9199084')
         self.assertEqual(result[0]['attributes']['uniqueIdentifier'], 'A91461-9199084')
```

mhs/common/mhs_common/workflow/common.py
```diff
@@ -63,9 +72,17 @@ def __init__(self, routing: routing_reliability.RoutingAndReliability = None):
...
-            return url, to_party_key, cpa_id
+            to_asid = self._extract_asid(endpoint_details)
+            details = {self.ENDPOINT_SERVICE_ID: service_id,
+                       self.ENDPOINT_URL: url,
+                       self.ENDPOINT_PARTY_KEY: to_party_key,
+                       self.ENDPOINT_CPA_ID: cpa_id,
+                       self.ENDPOINT_TO_ASID: to_asid
+                       }
+            logger.info('0002',
+                        'Retrieved endpoint details for {service_id}. {url}, {party_key}, {cpa_id} {to_asid}',
```
ghost: Not something you have to change now, but I think the logger does pretty well at logging dictionary contents, so it might mean fewer future changes to do:
```suggestion
                        'Retrieved endpoint details for {service_id}. {details}, details)',
```

mhs/common/mhs_common/workflow/tests/test_common.py
ghost: It's still missing, isn't it?

mhs/common/mhs_common/errors/soap_handler.py
ghost: Yeah, it probably makes sense to handle the fix as part of that branch (and making sure @ismail-s is aware of it, I suppose?).

mhs/common/mhs_common/workflow/tests/test_common.py
Philip-Woods: 🤦‍♂

integration-tests/integration_tests/helpers/methods.py
Philip-Woods: Accidentally committed it for my local integration tests

mhs/spineroutelookup/lookup/tests/test_sds_client.py
Philip-Woods: Had to update the mock server files whilst I was at it as well since I realised my asid was in there

mhs/common/mhs_common/workflow/common.py
```diff
@@ -63,9 +72,17 @@ def __init__(self, routing: routing_reliability.RoutingAndReliability = None):
...
-                        {'service_id': service_id, 'url': url, 'to_party_key': to_party_key, 'cpa_id': cpa_id})
-            return url, to_party_key, cpa_id
+            to_asid = self._extract_asid(endpoint_details)
+            details = {self.ENDPOINT_SERVICE_ID: service_id,
+                       self.ENDPOINT_URL: url,
+                       self.ENDPOINT_PARTY_KEY: to_party_key,
+                       self.ENDPOINT_CPA_ID: cpa_id,
+                       self.ENDPOINT_TO_ASID: to_asid
+                       }
+            logger.critical('0002',
```
ghost:
```suggestion
            logger.info('0002',
```

mhs/common/mhs_common/errors/soap_handler.py
ismail-s: Fix is already in #78.

mhs/common/mhs_common/workflow/asynchronous_reliable.py
```diff
@@ -8,10 +8,17 @@
 class AsynchronousReliableWorkflow(common_asynchronous.CommonAsynchronousWorkflow):
     """Handles the workflow for the asynchronous reliable messaging pattern."""
 
-    async def handle_outbound_message(self, message_id: str, correlation_id: str, interaction_details: dict,
+    async def handle_outbound_message(self, from_asid: Optional[str],
+                                      message_id: str, correlation_id: str, interaction_details: dict,
                                       payload: str) -> Tuple[int, str]:
```
ghost: Think this type hint needs to be updated:
```suggestion
                                       payload: str) -> Tuple[int, str, Optional[wd.WorkDescription]]:
```

mhs/common/mhs_common/workflow/common.py
```diff
@@ -11,25 +11,34 @@
...
     def __init__(self, routing: routing_reliability.RoutingAndReliability = None):
         self.routing_reliability = routing
 
     @abc.abstractmethod
-    async def handle_outbound_message(self, message_id: str, correlation_id: str, interaction_details: dict,
+    async def handle_outbound_message(self, from_asid: Optional[str],
+                                      message_id: str, correlation_id: str, interaction_details: dict,
                                       payload: str,
                                       work_description_object: Optional[wd.WorkDescription]
                                       ) -> Tuple[int, str]:
```
ghost: Again, do we need?:
```suggestion
                                       ) -> Tuple[int, str, Optional[wd.WorkDescription]]:
```

ghost (approved):

PR #87, opened 2019-09-16
RT-171 secrets config

ismail-s:
- Add more logging to config component
- Add secrets config component
- Use secrets config component instead of config component throughout MHS wherever secret config is required
- [x] Run integration tests locally


common/utilities/config.py
```diff
@@ -38,10 +38,13 @@ def get_config(key: str, default: Optional[str] = _config_default) -> str:
...
+        logging.info(f'Obtained config ConfigName:"{key}" ConfigValue:"{config[key]}" ProcessKey=CONFIG001')
         return config[key]
     elif default is not _config_default:
+        logging.info(f'Failed to get config ConfigName:"{key}". Returning DefaultValue:"{default}". '
+                     f'ProcessKey=CONFIG002')
         return default
     else:
-        # Can't use IntegrationAdaptorsLogger due to circular dependency
-        logging.exception(f'Failed to get config ConfigName:"{key}" ProcessKey=CONFIG001')
+        logging.error(f'Failed to get config ConfigName:"{key}" ProcessKey=CONFIG003')
```
Philip-Woods: Optional since it's only got 3 log statements for the file, but might be a bit more consistent with the other files/neater if you set the log tag `CONFIG` as a variable, maybe the same for secrets but not too important

common/utilities/secrets.py
```diff
@@ -0,0 +1,51 @@
...
+This module holds the secret config used by an application. To use this module, first
+call `setup_secrets_config` to populate `secret_config`. Then, just get any required config
+using `get_secret_config` (or directly use `secret_config` for more complex use cases).
+"""
+import logging
+import os
+from typing import Dict, Optional
+
+secret_config: Dict[str, str] = {}
+
```
Philip-Woods, line 13: So something that Joe raised while he was having a run through our project was that this config stuff seemed a bit 'magic', and (at least in other languages) probably should be something that should be wrapped in a class with static elements (though obviously python doesn't have a natural static attributes in classes). In particular if you were to import the methods and not the module the element retrieval fails. I don't feel too strongly on changing the implementation but I figured it's something to be wary of.

Philip-Woods (dismissed):
Looks good, the only other question I have is that am I right in thinking this is effectively a copy of the default config except we're using/looking for specific naming convention to identify secret values? There's no other unique behavior?

common/utilities/tests/test_secrets.py
```diff
@@ -0,0 +1,81 @@
...
+        mock_environ["PREFIX_SECRET_TEST"] = "123"
+
+        secrets.setup_secret_config("PREFIX")
+
+        self.assertEqual("111", secrets.get_secret_config("LOG_LEVEL", default="111"))
+
+    def test_get_secret_config_default_none(self, mock_environ):
+        mock_environ["PREFIX_SECRET_TEST"] = "123"
+
+        secrets.setup_secret_config("PREFIX")
```
Philip-Woods: Minor as well: don't think we need these lines?

common/utilities/secrets.py
ismail-s: Ok, I've thought about this a bit.

When implementing this, the approach was to go with the simplest thing that gets the job done. So just filter environment variables by a prefix, have a setup function that populates a a dict, then that dict has all the config in it that everything else can reference (I thought about just using `os.environ` but that would add a little duplication everywhere). Then the function to get a config value was added, as a helper really (that does a bit of logging).

Could this be done as a class? Yes. But I don't think it would be any simpler (you'd probably want a singleton, but then would this be dependency-injected everywhere or just imported (ie dependency-injection in another way), in which case there's a little more complexity.

And tbh, I think modules are a bit like a singleton class. In Python, modules get cached after being imported. You can reload a module, but that's advanced wizardry that no one uses unless they really need to.

If this was a situation where multiple parts of the codebase were modifying the `config`/`secret_config` variables, then I'd find this iffy (though looking at the Python standard library, that's the way some of it's API actually works). But the only place that is doing so is this module itself. And the only place that calls the `setup_config` functions are the `main.py` files.

So I'd rather keep it as-is than change it, but if we were doing more fancy stuff, then it might need to be changed.

ismail-s:
@Philip-Woods I think I've addressed the review comments.

Regarding https://github.com/nhsconnect/integration-adaptors/pull/87#pullrequestreview-288719585, the difference between `config.py` and `secrets.py` is:
- slight difference in logging `secrets.py` doesn't log values, `config.py` does log config values
- naming convention of environment variables
- keeping the two things a bit separate from each other

Looking at the code now, what I could have done instead was add another function to `config.py` called `get_secret_config`. But I've already written this code. That, and when @dshipp-bjss was discussing it, IIRC he was talking about it being a separate thing, so I just went with that (ie blame Dave, not me..).

mhs/outbound/main.py
```diff
@@ -92,15 +94,15 @@ def main():
     ca_certs = "client.pem"
 
     certs_dir.mkdir(parents=True, exist_ok=True)
-    (certs_dir / client_cert).write_text(config.get_config('CLIENT_CERT'))
-    (certs_dir / client_key).write_text(config.get_config('CLIENT_KEY'))
-    (certs_dir / ca_certs).write_text(config.get_config('CA_CERTS'))
+    (certs_dir / client_cert).write_text(secrets.get_secret_config('CLIENT_CERT'))
```
ghost: We do this process of extracting the certificate environment variables and writing them to file in a few places now. Is it worth extracting this to a utility?

mhs/README.md
```diff
@@ -33,10 +33,10 @@ MHS takes a number of environment variables when it is run. These are:
 the most logs and `CRITICAL` displays the least. Note: Setting this value to one of the more detailed 'standard' Python
 log levels (such as `DEBUG` or `NOTSET`) may result in the libraries used by this application logging details that
 contain sensitive information such as the content of messages being sent.
-* `MHS_PARTY_KEY` The party key associated with your MHS.
-* `MHS_CLIENT_CERT` (outbound & Spine Route Lookup service only) Your endpoint certificate
-* `MHS_CLIENT_KEY` Your endpoint private key
-* `MHS_CA_CERTS` Should include the following in this order: endpoint certificate, endpoint issuing subCA certificate, root CA Certificate.
+* `MHS_SECRET_PARTY_KEY` The party key associated with your MHS.
+* `MHS_SECRET_CLIENT_CERT` (outbound & Spine Route Lookup service only) Your endpoint certificate
+* `MHS_SECRET_CLIENT_KEY` Your endpoint private key
```
ghost, line 38: Probably worth highlighting that this is `(outbound & Spine Route Lookup service only)` too.

ismail-s:
@gareth-allan I've addressed your comments and rerun the integration tests locally.

mhs/README.md
ghost: Given we only use the client key with the client certificate, shouldn't we also have?:
```suggestion
* `MHS_SECRET_CLIENT_KEY` (outbound & Spine Route Lookup service only) Your endpoint private key
```

ismail-s: AFAICT `MHS_CLIENT_KEY` is used in outbound, inbound and spine route lookup:
- https://github.com/nhsconnect/integration-adaptors/blob/f6c5e5cf63d3e9c7eca11d5e048af5f463d6d720/mhs/inbound/main.py#L61
- https://github.com/nhsconnect/integration-adaptors/blob/f6c5e5cf63d3e9c7eca11d5e048af5f463d6d720/mhs/outbound/main.py#L96
- https://github.com/nhsconnect/integration-adaptors/blob/f6c5e5cf63d3e9c7eca11d5e048af5f463d6d720/mhs/spineroutelookup/main.py#L41

ismail-s: Discussed offline with @gareth-allan, realised that we probably should be passing `CLIENT_CERT` to MHS inbound, and at the same time be able to remove the endpoint certificate from `CA_CERTS`. I've made this change in 158d2bb.

mhs/inbound/main.py
```diff
@@ -60,15 +60,17 @@ def load_party_key(data_dir: pathlib.Path) -> str:
...
                          workflows: Dict[str, workflow.CommonWorkflow],
                          persistence_store: persistence_adaptor.PersistenceAdaptor
                          ) -> None:
     """
 
     :param persistence_store: persistence store adaptor for message information
-    :param certs_file: The filename of the certificate to be used to identify this MHS to a remote MHS.
-    :param key_file: The filename of the private key for the certificate identified by certs_file.
+    :param local_certs_file: The filename of the certificate to present for authentication.
+    :param ca_certs_file: The filename of the CA certificates that form the certificate chain for the certificate
```
ghost: A minor point, but I don't think these certificates form the cert chain for the local certificate. Rather, I think they're used as the CA certificates store to verify the certificate presented by Spine when a connection is made to the inbound service.

This would seem to be backed up by the [documentation for `load_verify_locations`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_verify_locations), which says:
>Load a set of “certification authority” (CA) certificates used to validate other peers’ certificates [...]

ghost (dismissed):

mhs/inbound/main.py
ismail-s: I've taken a look on the command line at one of our Opentest certs. They do form a cert chain in terms of being signed by each other, with the root cert being self-signed.

I see what the Python docs say, that said the docs for [`load_cert_chain`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain) say:
>  The certfile string must be the path to a single file in PEM format containing the certificate as well as any number of CA certificates needed to establish the certificate’s authenticity.

So you could argue that we shouldn't just be passing the endpoint certificate there. I think the CA certs we're passing to `load_verify_locations` are being used to verify certs on both sides of the connection.

To try and understand this better, I've taken a quick dive through the implementation of `load_verify_locations`. This function essentially ends up calling [`SSL_CTX_load_verify_locations`](https://github.com/python/cpython/blob/69b3718b183a698202bd67488639bffd64a488bc/Modules/_ssl.c#L4053)  which is documented [here](https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_load_verify_locations.html). I'm more confused by reading these docs, they kind of sound like they back up my hypothesis, but they also sound like they don't .

@gareth-allan what do you think?

ghost: I think it's a difference in what these certificates are used for. The cert file passed to `load_cert_chain` is what's presented when a connection is made to this application, so it _could_ be a chain of certificates, if that's what's needed to allow the remote end of the connection to validate our certificate, but that's not necessary with the infrastructure we're using.

The CA certs passed to `load_verify_locations` are used to validate the certificate presented by the remote end of the connection. In our case, we use the sub CA and the root CA certs, so we can establish the certificate chain if a lone certificate is presented.

All that said, I don't think we can hope to explain the whole situation in a docstring, and I'm not keen to hold up this review over it. Could we just have our docstrings say something like "as passed to SSLContext.load_verify_locations()"?

ghost (approved):

PR #88, opened 2019-09-18
Feature/rt 98 scr population

Philip-Woods:
This PR contains the changes required for the SCR endpoint to populate a Gp summary upload message template via the SCR web service end point, using the supplied json dictionary. 

* Modifications have been made to the SCR library to remove some unnecessary  values within the template
* The SCR web service endpoint has been refactored to delegate responsibility to a message forwarder - which handles the population calls and will eventually delegate the calls to the MHS


SCRWebService/endpoints/summary_care_record.py
```diff
@@ -0,0 +1,78 @@
...
+import json
+from typing import Dict
+
+import tornado.web
+import tornado.ioloop
+import utilities.integration_adaptors_logger as log
+from message_handling import message_forwarder as mh
+from builder.pystache_message_builder import MessageGenerationError
+
+logger = log.IntegrationAdaptorsLogger('GP_SUM_UP')
```
Guardian-Development: is that the right logger name ? should be SUMMARY_CARE_RECORD ?

SCRWebService/endpoints/summary_care_record.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        self.forwarder = forwarder
+
+    def post(self):
+        """
+        Receives a json payload, parses the appropriate message details before processing the message contents
+        :return:
+        """
+
+        logger.info('001', 'Message received for gp summary upl')
```
Guardian-Development: upl ?

SCRWebService/endpoints/summary_care_record.py
```diff
@@ -0,0 +1,78 @@
...
+    def _extract_message_body(self):
+        """Extracts the message body from the request
+        :return: The message body
+        """
+        try:
+            return json.loads(self.request.body)
+        except json.decoder.JSONDecodeError as e:
+            logger.error('001', f'Failed to parse message body: {e}')
+            raise tornado.web.HTTPError(400, 'Failed to parse json body from request',
+                                        reason=f'Failed to parse json body from request: {str(e)}')
```
Guardian-Development: e isnt the request with regards to the error message: it is the exception.

SCRWebService/endpoints/tests/test_summary_care_record.py
```diff
@@ -0,0 +1,59 @@
...
+        self.forwarder = mock.MagicMock()
+
+        return Application([
+            (r'/gp_summary_upload', summary_care_record.SummaryCareRecord,
+             dict(forwarder=self.forwarder))
+        ])
+
+    def test_handler_returns_message_processing_response(self):
+        body = file_utilities.FileUtilities.get_file_dict(complete_data_path)
+        self.forwarder.forward_message_to_mhs.return_value = "Nice response message"
```
Guardian-Development, line 31: if we are going to mock things, we should assert the mock is invoked with the expected parameters (so we know the rest layer is deserialising the jSON correctly). This is also why i dont like mocking.

SCRWebService/endpoints/tests/test_summary_care_record.py
```diff
@@ -0,0 +1,59 @@
...
+        response = self.fetch(GP_SUMMARY_UPLOAD_URL, method='POST', headers={'interaction-id': '123'},
+                              body=json.dumps(body))
+
+        self.assertEqual(response.body.decode(), "Nice response message")
+
+    def test_empty_body_request_fails(self):
+        body = ''
```
Guardian-Development, line 39: id make the body empty json {}, so its still valid. This may be a seperate test though.

SCRWebService/message_handling/tests/test_message_forwarder.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_forwarding_message_attempts_to_populate_message_template(self):
+        template_mock = mock.MagicMock()
+        interactions_map = {'interaction': template_mock}
+
+        handler = mh.MessageForwarder(interactions_map)
+        input_json = {'check': 'one'}
+        handler.forward_message_to_mhs('interaction', input_json)
+        template_mock.populate_template.assert_called_with(input_json)
+
+    def test_exception_raised_during_population_is_raised(self):
```
Guardian-Development: test name doesnt make sense, why is the exception raise ? what is the behaviour desired?

Guardian-Development (changes requested):

SCRWebService/endpoints/summary_care_record.py
Philip-Woods: Upload* - Fixed it

SCRWebService/endpoints/tests/test_summary_care_record.py
Philip-Woods: I think it's correct as it is, since `{}` will parse as valid json it will be left to the template populator to decide if it's missing keys, so from the endpoint perspective it's correct input. With that said, I'll update the test name to reflect that it's a null body vs an empty json request

SCRWebService/message_handling/tests/test_message_forwarder.py
Philip-Woods: Updated it, think the intention is clearer now but might want double checking it makes sense

Guardian-Development (dismissed):

Guardian-Development (approved):

PR #89, opened 2019-09-19
RT-158 review comments

ismail-s:
Handle review comments on [JIRA ticket](https://gpitbjss.atlassian.net/browse/RT-158) that came out of review with @briandiggle.

Original PR: #78.


colin-bateson (approved):
Changes look fine. 

Just a note that these files will need to change no longer include http statuses which will be covered in a subsequent branch

PR #90, opened 2019-09-19
RT-202: adding initial DynamoDB table assertions

Guardian-Development:
Functionality:
- added the ability to assert the state of the mhs_store
- added an initial test that uses this assertion to check the state of a sync request
- added a way of making http requests against the MHS in a reusable OO way.

Notes:
- this test looks like its found a bug which i am happy to fix as part of this with guidance. 
- this is only the framework for doing dynamo assertions, more thought needs to be had in the different states we should assert.


Jenkinsfile
```diff
@@ -122,6 +122,11 @@ pipeline {
                                     returnStdout: true,
                                     script: "terraform output route_lb_target_group_arn"
                                 ).trim()
+                                env.MHS_DYNAMODB_TABLE_NAME = sh (
+                                    label: 'Obtaining the dynamodb table name used for the MHS state',
+                                    returnStdout: true,
+                                    scrpit: "terraform output mhs_state_table_name"
```
ghost:
```suggestion
                                    script: "terraform output mhs_state_table_name"
```

ghost (changes requested):

ghost (approved):

PR #92, opened 2019-09-19
RT-180: Redis Cache Implementation

ghost:
This PR adds a Redis-backed implementation of the CacheAdaptor.

Changes:
- Add `RedisCache` implementation of the `CacheAdaptor` interface.
- Update the `docker-compose.yml` to run a local Redis container and pass the necessary configuration values to use it to the route service
- Added Terraform configuration to deploy and use an ElastiCache Redis cluster as part of our exemplar blueprint
- Removed the original dictionary-backed cache implementation

Notes:
- I've also removed the attempted implementation of a pattern that would allow cache adaptor implementations to be loaded based on a config key. It didn't work without importing all of the implementations, so needs some more thought.


mhs/spineroutelookup/lookup/redis_cache.py
```diff
@@ -0,0 +1,91 @@
...
+                logger.info("0004", "No cache entry found for {key}.", {"key": key})
+                return None
+
+            value = json.loads(cached_json_value)
+
+            logger.info("0006", "Retrieved cache entry for {key}. {value}", {"key": key, "value": value})
+
+            return value
+        except redis.RedisError as re:
+            logger.error("0003", "An error occurred when attempting to load {key}. {exception}",
```
Guardian-Development: the numbers of these logger entries are out of order which makes it a bit hard to check there are no duplicates.

mhs/spineroutelookup/lookup/tests/test_redis_cache.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @async_test
+    async def test_should_retrieve_value_from_store_if_exists(self):
+        cache = redis_cache.RedisCache(REDIS_HOST, REDIS_PORT)
+        self.mock_redis.get.return_value = VALUE_DICTIONARY_JSON.encode()
+
+        value = await cache.retrieve_mhs_attributes_value(ODS_CODE, INTERACTION_ID)
+
+        self.assertEqual(value, VALUE_DICTIONARY)
+        self.mock_redis.get.assert_called_with(CACHE_KEY)
```
Guardian-Development, line 61: i think this is a bit of an explicit assertion, this may make the test more brittle than it needs to be. The fact we get the correct value back shows the behaviour is working correctly IMO.

Guardian-Development (dismissed):
approved as my 2 comments are nice to have, will leave it up to you if you fancy making the changes 👍

Guardian-Development (approved):

mhs/spineroutelookup/lookup/tests/test_redis_cache.py
ghost: As discussed offline, the goal here was to confirm that the correct argument was passed to the `mock_redis.get` call, rather than asserting that this specific method was called. Unfortunately, as far as I can tell this needs to be done using an assert statement, rather than when setting up the mock.

PR #95, opened 2019-09-20
Feature/rt 90 forward to mhs

Philip-Woods:


SCRWebService/endpoints/summary_care_record.py
```diff
@@ -73,6 +89,34 @@ def _extract_message_body(self):
         try:
             return json.loads(self.request.body)
         except json.decoder.JSONDecodeError as e:
-            logger.error('004', 'Exception raised whilst parsing message body: {exception}', {'exception': e})
+            logger.error('005', 'Exception raised whilst parsing message body: {exception}', {'exception': e})
```
Guardian-Development, line 92: these logger numbers should be in order (hard to check for duplicates etc)

SCRWebService/endpoints/summary_care_record.py
```diff
@@ -73,6 +89,34 @@ def _extract_message_body(self):
...
+        """
+        Attempts to extract the correlation-id from the message headers, if this fails a new one is generated,
+        either way the correlation id is set in the logger
+        :return: A UUID
+        """
+        correlation_id = self.request.headers.get('Correlation-Id', None)
+        if not correlation_id:
+            correlation_id = message_utilities.MessageUtilities.get_uuid()
+            log.correlation_id.set(correlation_id)
+            logger.info('006', "Failed to extract correlation-id from message, generated a new one")
```
Guardian-Development: reword message, should be something like "No correlation ID provided by caller, using {ID}"

SCRWebService/endpoints/tests/test_scr_adaptor_components.py
```diff
@@ -0,0 +1,148 @@
...
+                   body=json.dumps(body))
+
+        correlation_id_header = request_mock.call_args[1]['headers']['correlation-id']
+        message_id_header = request_mock.call_args[1]['headers']['message-id']
+        self.assertEqual(message_id_header, "messageId1212")
+        self.assertEqual(correlation_id_header, 'correlationId1212')
+
+    @mock.patch('utilities.message_utilities.MessageUtilities.get_uuid')
+    @mock.patch('comms.common_https.CommonHttps.make_request')
+    def test_when_no_correlation_id_is_provided_the_adaptor_generates_one_and_passes_it_to_the_mhs(self,
```
Guardian-Development, line 132: this correlation ID should be returned to the caller in the response headers i would think also ? (so the caller knows what ID is being used to track this request if they were having an issue with the system and needed to report it). This may be happening, but we should assert the response.

SCRWebService/message_handling/message_sender.py
```diff
@@ -0,0 +1,56 @@
...
+        :return:
+        """
+        logger.info('001', 'Preparing message headers to send to mhs address')
+        headers = self._build_headers(interaction_id, message_id, correlation_id)
+        response = await CommonHttps.make_request(url=self.mhs_address,
+                                                  headers=headers,
+                                                  body=message_body,
+                                                  method='POST')
+
+        logger.info('002', 'Response received from MHS')
```
Guardian-Development: this should probably log the response code (as retrieving a response without knowing it isnt successful is pretty useless)

SCRWebService/message_handling/message_sender.py
```diff
@@ -0,0 +1,56 @@
...
+        :return:
+        """
+        headers = {
+            'Interaction-Id': interaction_id,
+            'sync-async': 'true'
+        }
+        if message_id:
+            headers['message-id'] = message_id
+
+        if correlation_id:
```
Guardian-Development: there should always be a correlation id, no need for if

Guardian-Development (changes requested):
* will approve once changes are made, and we know whether the interactions with the MHS should be going over TLS.

SCRWebService/endpoints/summary_care_record.py
Philip-Woods: Once set as a context var the correlation ID is added to all log lines, so this line will print out the correlation id by default

SCRWebService/endpoints/tests/test_scr_adaptor_components.py
Philip-Woods: I believe the intention was to have correlation id as almost a purely internal thing unless they provide it, the MHS also doesn't return the correlation id  which makes me think this shouldn't be added

Guardian-Development (approved):

PR #100, opened 2019-09-23
RT-108 change MHS request body to JSON

ismail-s:
- Add marshmallow, apispec and spispec-webframeworks dependencies to MHS outbound. These are used for request validation and API documentation.
- Change request body to JSON and validate this request body. Currently only using the `payload` part of the request body, as part of work on the forward reliable workflow, we'll be using the attachments.
- dacb525 add missing [wraps](https://docs.python.org/3.7/library/functools.html#functools.wraps) decorators. This decorator propagates things like the docstring to the wrapper function.
- Add OpenAPI spec for MHS and generate this in the CI build. Currently this isn't being saved anywhere, or exported in a nice format like a html webpage. Given the current stage the project is in, I'm not sure how much time to spend on doing this.
- b00d497 Fix bug where error response had an incorrect status-line textual phrase being set


mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -26,6 +31,92 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
...
+
+
+              For sync interactions or async interactions that don't support
+              sync-async, this header must be set to false.
+          - name: from-asid
+            in: header
+            required: false
+            schema:
+              type: string
+            description: from-asid
```
ismail-s: @colin-bateson @Philip-Woods Do either of you know what this from-asid is (enough to write a description of it here)?

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -26,6 +31,92 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
...
+
+              When performing async requests where the response is put on the
+              inbound queue, the correlation ID will be put with the response on the
+              queue.
+
+
+              Note that this correlation ID gets sent to/from Spine.
+        responses:
+          200:
+            description: Successful response from Spine.
```
ismail-s, line 112: This could be documented more, but I'm not sure how much we know about what the response body will look like. I'm guessing that this will just be an XML response, and that's it. Am I right @colin-bateson?

mhs/outbound/outbound/request/synchronous/request_body_schema.py
```diff
@@ -0,0 +1,53 @@
+import dataclasses
+from typing import List
+
+import marshmallow.validate
+
+# These allowed content types for attachments are taken from the EIS part 2.5.4.2
+_ATTACHMENT_ALLOWED_CONTENT_TYPES = (
+    'text/plain', 'text/html', 'application/pdf', 'text/xml', 'application/xml', 'text/rtf', 'audio/basic',
+    'audio/mpeg', 'image/png', 'image/gif', 'image/jpeg', 'image/tiff', 'video/mpeg', 'application/msword'
+)
```
ismail-s: @colin-bateson In the EIS part 2.5.4.2 there are 2 tables, one has `application/octet-stream` and the other doesn't. Do we know whether this is a content-type that we should allow?

mhs/outbound/outbound/request/synchronous/request_body_schema.py
```diff
@@ -0,0 +1,53 @@
...
+                                        description='Whether the attachment payload is base64-encoded or not. This is '
+                                                    'only required for binary attachments eg images.',
+                                        truthy={True}, falsy={False})
+    content_type = marshmallow.fields.Str(required=True, description='Content type of the attachment',
+                                          validate=marshmallow.validate.OneOf(_ATTACHMENT_ALLOWED_CONTENT_TYPES))
+    payload = marshmallow.fields.Str(required=True,
+                                     description='The attachment, possibly base64-encoded as per is_base64.',
+                                     validate=marshmallow.validate.Length(min=1))
+    description = marshmallow.fields.Str(required=True, description='Description of the attachment',
+                                         validate=marshmallow.validate.Length(min=1))
```
ismail-s: @colin-bateson Should there be a max length for description? I can't see anything written down in the  EIS appendix A.1.1.4 and A.1.1.5 .

mhs/outbound/outbound/request/synchronous/request_body_schema.py
```diff
@@ -0,0 +1,53 @@
...
+
+@dataclasses.dataclass
+class RequestBody:
+    payload: str
+    attachments: List[Attachment]
+
+
+class RequestBodySchema(marshmallow.Schema):
+    payload = marshmallow.fields.Str(required=True, description='Payload to send to Spine',
+                                     validate=marshmallow.validate.Length(min=1))
```
ismail-s: @colin-bateson Should there be a max length for payload? Same question really as for https://github.com/nhsconnect/integration-adaptors/pull/100#discussion_r327096505

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -26,6 +31,93 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
...
+            required: true
+            schema:
+              type: string
+            description: ID of the interaction that you want to invoke. e.g. QUPC_IN160101UK05
+          - name: sync-async
+            in: header
+            required: true
+            schema:
+              type: string
+              enum: ["true", "false"]
```
ismail-s, line 52: So this looks a bit odd, but when I took a look through the [OpenAPI](https://swagger.io/specification) spec, I couldn't figure out how a boolean would get serialised for certain. So I've done this like this. I'm not against changing it back, if we know this for certain.

mhs/outbound/outbound/request/synchronous/handler.py
Philip-Woods: It's the ASID of the sending system, specifically used to generate the of headers the `sync` workflow, this should match the `from-asid` value used within the hl7 body of a sync message

mhs/outbound/outbound/request/synchronous/handler.py
colin-bateson: Yes it would be an XML response

mhs/outbound/outbound/request/synchronous/request_body_schema.py
colin-bateson: I'd say include it rather than exclude it

mhs/outbound/outbound/request/synchronous/request_body_schema.py
colin-bateson: Can you add a comment to the ticket to say that we are not implementing a max length, as this is not defined in the EIS?

mhs/outbound/outbound/request/synchronous/request_body_schema.py
colin-bateson: Can you add a comment to the ticket to say that we are not implementing a max length, as this is not defined in the EIS?

mhs/outbound/outbound/request/synchronous/request_body_schema.py
ismail-s: I looked again at the EIS and found some docs around max length for attachments but not description or payload (the HL7 payload). So I've put a comment onto [the story](https://gpitbjss.atlassian.net/browse/RT-107).

mhs/outbound/outbound/request/synchronous/request_body_schema.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@dataclasses.dataclass
+class RequestBody:
+    payload: str
+    attachments: List[Attachment]
+
+
+class RequestBodySchema(marshmallow.Schema):
+    payload = marshmallow.fields.Str(required=True, description='Payload to send to Spine',
```
ismail-s: I need to say here that this is HL7 payload

mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -48,15 +145,41 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
...
-        return body
+        try:
+            # Parse the body as JSON and validate it against RequestBodySchema
+            parsed_body: request_body_schema.RequestBody = request_body_schema.RequestBodySchema().loads(body)
+        except json.JSONDecodeError as e:
+            logger.error('0013', 'Invalid JSON request body')
+            raise tornado.web.HTTPError(400, 'Invalid JSON request body', reason='Invalid JSON request body') from e
+        except marshmallow.ValidationError as e:
+            # e.messages is a nested dict of all the validation errors
+            validation_errors = str(e.messages)
```
ismail-s, line 172: Note that I briefly looked into transforming `e.messages` into a string by iterating through it, but realised that it being nested (and the way it is nested) meant that this would be a bit complicated, and in any case, just running it through `str` gives us a formatted string of sorts.

mhs/outbound/outbound/request/synchronous/request_body_schema.py
```diff
@@ -0,0 +1,59 @@
+import dataclasses
```
ismail-s: I need to document this file more.

colin-bateson (dismissed):

colin-bateson (dismissed):

colin-bateson (dismissed):

colin-bateson (approved):

colin-bateson (approved):

PR #110, opened 2019-09-26
Update final work description write to use retries

Philip-Woods:


ismail-s (approved):
Went over this in-person with @Philip-Woods

PR #112, opened 2019-09-26
RT-180: Improve Cache Adaptor Error Handling

ghost:
This PR updates the `MHSAttributeLookup` component so that it handles exceptions raised by the `CacheAdaptor`, and the `RedisCache` implementation so that it re-raises Redis errors. This allows `MHSAttributeLookup` to differentiate between error cases and cache misses.

The LDAP base to search from when querying SDS has also been made configurable, as it is environment-specific. This involves defining a new environment variable to pass to the route service and updating the terraform/Jenkinsfile & docker-compose file to pass the correct value when we are using Opentest's SDS instance.

I've also updated the details of the Terraform [issue](https://github.com/terraform-providers/terraform-provider-aws/issues/8561) I was seeing when deploying an ElastiCache cluster. As the [latest comment](https://github.com/terraform-providers/terraform-provider-aws/issues/8561#issuecomment-535672966) identifies, it appears to be an issue with the length of the cluster/replication group name that occurs when encryption in transit is enabled. As such, I've shortened the suffix we use for our replication group, added some documentation on the acceptable maximum length of the environment ID variable and updated the 'known issues' section in the Terraform readme.


dshipp-bjss (approved):
Over the shoulder reviewed

PR #118, opened 2019-09-30
Test/component test happy path

Guardian-Development:


integration-tests/integration_tests/integration_tests/http/mhs_http_request_builder.py
```diff
@@ -61,11 +61,24 @@ def execute_post_expecting_success(self) -> Response:
...
 
         return response
+
+    def execute_post_expecting_error_response(self) -> Response:
+        """
+        Execute a POST request against the MHS using the configured body and headers within this class.
+        Asserts the response is 500.
+        :return: response from MHS
+        """
+        response = requests.post(self.mhs_host, headers=self.headers, data=self.body)
```
ghost: This is going to need to be:
```suggestion
        response = requests.post(self.mhs_host, headers=self.headers, data=self.body, verify=False)
```
Otherwise certificate validation is going to fail in our build environment. Maybe worth extracting the actual `requests.post` to a method, to avoid duplication?

ghost (changes requested):

ghost (approved):

PR #127, opened 2019-10-02
RT-255 Add unittests to check for duplicate log keys

ismail-s:
- Add unittests to check for duplicate log keys
  - The unittest is pretty hacked together, but making it more proper can be quite tricky (ie I could have gone and used Python's ast module, allowed different names for the logger variables), and this is more to get something that does the job without spending ages making it perfect.
  - I've added the unittest to common and just subclassed it in all other places to reduce duplication. Though this hasn't worked in various places, so I might need to create another techdebt to address this (or restructure things slightly)


colin-bateson (approved):
- Raise tech debt for common test usage across services.
- Raise tech debt to pull log levels from logger

ismail-s:
@colin-bateson I've created [RT-301](https://gpitbjss.atlassian.net/browse/RT-301).

PR #146, opened 2019-10-04
Use update_status_with_retries in more places

ismail-s:
Use `update_status_with_retries` in more places. The way I went about making these changes was I first reasoned about when there was potential for race conditions:

- If sync-async:
  - in MHS outbound, from when making the request to Spine to when the resynchroniser returns
  - in MHS inbound, for the whole time when handling the response that came from Spine
- If not sync-async:
  - in MHS outbound, from when making the request to Spine till the request has been fully handled
  - in MHS inbound, for the whole time when handling the response that came from Spine

Then I went through the code and made changes wherever we weren't handling potential race conditions.

I wanted to add some component tests around these scenarios, but I don't think that is going to be possible with our current setup, as it would require having more control of our DynamoDB instance and making it delay responding/respond in certain ways in coordination with Spine. This would not be a quick thing to add.


Philip-Woods (approved):

ismail-s:
This has been combined into #147, so closing this PR.

PR #152, opened 2020-02-20
NIAD-45: Set timeout and retries for LDAP connection

grzegorz-robak:


mhs/spineroutelookup/lookup/sds_connection_factory.py
```diff
@@ -5,12 +5,15 @@
 
 import definitions
 
+_MAX_RETRIES = 3
+_CONNECTION_TIMEOUT = 5
```
bartek-sarul: in such case it's worth adding a comment saying the unit of time span used

bartek-sarul (changes requested):

mhs/spineroutelookup/lookup/sds_connection_factory.py
grzegorz-robak: Good idea!

jaroslaw-wojciechowski (dismissed):

jaroslaw-wojciechowski (dismissed):

mhs/spineroutelookup/lookup/sds_connection_factory.py
```diff
@@ -5,12 +5,15 @@
 
 import definitions
 
+_MAX_RETRIES = 3
```
bartek-sarul: I was looking at code here and a common approach is to load values from environment variables but override with default one if env var is not provided
For example
https://github.com/nhsconnect/integration-adaptors/pull/153/files#diff-8a04b7e2f084c2f292ffdc46a46ccc61R88

bartek-sarul (changes requested):

mhs/spineroutelookup/lookup/sds_connection_factory.py
grzegorz-robak: Done. Added some additional logging also.

bartek-sarul (dismissed):

PR #153, opened 2020-02-24
NIAD-62: configurable server port

bartek-sarul:


mhs/inbound/nhs-inbound-env-example.yaml
```diff
@@ -1,4 +1,5 @@
 #Common env variables for inbound
+MHS_SERVER_PORT: "8081"
```
grzegorz-robak: I'd made server port env variable different for each component (like `MHS_INBOUND_SERVER_PORT`) so they won't clash with each other.

grzegorz-robak (dismissed):

bartek-sarul:
Most likely there is also some place that is invoking those services that need change to use new ports

petervdm:
> Most likely there is also some place that is invoking those services that need change to use new ports

Probably in the docker-compose.yml?

grzegorz-robak (dismissed):

bartek-sarul:
> Probably in the docker-compose.yml?

@petervdm containerised environment is not a problem as if you don't set the env var, it will use the default port 80 which is fine

what I'm looking for is any place where container A is trying to call container B via HTTP

I think I'ver found something

jaroslaw-wojciechowski:
please check port mappings in docker-compose.
These should be also updated:
https://github.com/nhsconnect/integration-adaptors/blob/18143106d5143f6844764304eddef14b04c95e87/docker-compose.yml#L8

for example healthchecks on original ports:
localhost:8080/healthcheck - router
localhost:80/healthcheck - outbound
localhost:8079/healthcheck - inbound

mhs/inbound/main.py
```diff
@@ -85,9 +85,10 @@ def start_inbound_server(local_certs_file: str, ca_certs_file: str, key_file: st
     healthcheck_application = tornado.web.Application([
         ("/healthcheck", healthcheck_handler.HealthcheckHandler)
     ])
-    healthcheck_application.listen(80)
+    server_port = int(config.get_config('INBOUND_SERVER_PORT', default='80'))
+    healthcheck_application.listen(server_port)
 
-    logger.info('011', 'Starting inbound server')
+    logger.info('011', 'Starting inbound server at port ' + str(server_port))
```
mattd-kainos: I think our logger is setup to accept template strings, for example `logger.info('0010', 'Returning response with {HttpStatus}', {'HttpStatus': status})`. You could also use an f-string here `f'Starting inbound server at port {server_port}'`. Using + for string concatenation in Python can be unpredictable.

bartek-sarul: good point

bartek-sarul:
@jaroslaw-wojciechowski dockers are working correctly without my change. This JIRA was only to fix port clashing when running all locally

mhs/inbound/main.py
```diff
@@ -85,10 +85,20 @@ def start_inbound_server(local_certs_file: str, ca_certs_file: str, key_file: st
...
+    server_port = int(config.get_config('INBOUND_SERVER_PORT', default='80'))
+    healthcheck_application.listen(server_port)
 
-    logger.info('011', 'Starting inbound server')
-    tornado.ioloop.IOLoop.current().start()
+    logger.info('011', 'Starting inbound server at port {server_port}', {'server_port': server_port})
+    tornado_io_loop = tornado.ioloop.IOLoop.current()
+    try:
+        tornado_io_loop.start()
+    except KeyboardInterrupt:
```
bartek-sarul, line 97: To silence the red log entry when you kill app in pycharm

bartek-sarul: I know it's a duplicate between all 3 servers but it might be hard to change that beacuse of the fcnkg logging lib

mhs/outbound/main.py
```diff
@@ -101,14 +101,25 @@ def start_tornado_server(data_dir: pathlib.Path, workflows: Dict[str, workflow.C
...
-        [(r"/", client_request_handler.SynchronousHandler,
-          dict(config_manager=config_manager, workflows=workflows)),
-         (r"/healthcheck", healthcheck_handler.HealthcheckHandler)])
+        [
+            (r"/", client_request_handler.SynchronousHandler,dict(config_manager=config_manager, workflows=workflows)),
+            (r"/healthcheck", healthcheck_handler.HealthcheckHandler)
+        ])
     supplier_server = tornado.httpserver.HTTPServer(supplier_application)
-    supplier_server.listen(80)
+    server_port = int(config.get_config('OUTBOUND_SERVER_PORT', default='80'))
```
bartek-sarul, line 109: No env var in cfg as the default port 80 is ok and doesn't break stuff in dockers

bartek-sarul:
@jaroslaw-wojciechowski  @mattd-kainos  please continue with code review

mhs/inbound/main.py
jaroslaw-wojciechowski: just a question, because it's out of the scope of this task - is it a graceful shutdown?
And should we even care about ongoing requests?

jaroslaw-wojciechowski (dismissed):

mattd-kainos (dismissed):

mhs/inbound/main.py
```diff
@@ -85,10 +85,20 @@ def start_inbound_server(local_certs_file: str, ca_certs_file: str, key_file: st
     healthcheck_application = tornado.web.Application([
         ("/healthcheck", healthcheck_handler.HealthcheckHandler)
     ])
-    healthcheck_application.listen(80)
+    server_port = int(config.get_config('INBOUND_SERVER_PORT', default='80'))
+    healthcheck_application.listen(server_port)
```
grzegorz-robak: This port is used just for healthcheck - all operations are done through 443 (see line 83). Should the other one be configurable too?

grzegorz-robak (changes requested):

mhs/inbound/main.py
mattd-kainos: Good catch! I think all of the services have two ports - healthcheck and API.

grzegorz-robak: AFAIK the rest are using just one port

mhs/inbound/main.py
bartek-sarul: no it's not
I tried doing the graceful shitdown using this https://github.com/tornadoweb/tornado/issues/1791
but looks like we use a different version of tornado so it's missing some pieces

mhs/inbound/main.py
bartek-sarul: I assume we shouldn't change that 443 even for local development as it will break spine unsolicited responses I guess

anyway the main reason was to change the port 80 to something else as all 3 were using it
443 looks to be used by inbound only

grzegorz-robak: So `INBOUND_SERVER_PORT` is a little misleading. Maybe `INBOUND_HEALTHCHECK_PORT`?

Might be good to change the variable name also

grzegorz-robak (approved):

PR #155, opened 2020-02-26
NIAD-48: use common retry component

bartek-sarul:


mhs/common/mhs_common/workflow/tests/test_synchronous.py
```diff
@@ -212,7 +212,7 @@ def setUp(self) -> None:
 
         wdo.set_outbound_status.assert_called_with(work_description.MessageStatus.OUTBOUND_MESSAGE_RESPONSE_RECEIVED)
         self.assertEqual(error, 500)
-        self.assertEqual(text, 'Error(s) received from Spine: HTTP 451: Unknown')
+        self.assertEqual(text[:40], 'Error(s) received from Spine: HTTP 451: ')
```
bartek-sarul, line 215: Because running this locally returned different error text but the status code is the same correct 451 one

grzegorz-robak (approved):

mhs/common/mhs_common/workflow/asynchronous_forward_reliable.py
```diff
@@ -91,25 +91,20 @@ def __init__(self, party_key: str = None, persistence_store: persistence_adaptor
...
-                                    "onto inbound queue", {"retry_delay": self.inbound_queue_retry_delay})
-                await asyncio.sleep(self.inbound_queue_retry_delay)
+        result = await retriable_action.RetriableAction(
+            lambda: self._put_message_onto_queue_with(message_id, correlation_id, payload, attachments=attachments),
+            self.inbound_queue_max_retries,
+            self.inbound_queue_retry_delay)\
+            .execute()
+
+        if not result.is_successful:
+            logger.error("0020",
```
mattd-kainos, line 101: At some point we should agree a standard on single vs double quotes, but I don't think we need to do that right now.

mattd-kainos (approved):
Nice that it was a straightforward change to use the Adaptor. :+1: for making tests more resilient.

PR #167, opened 2020-03-03
NIAD-42: Error responses now return Correlation-Id header

grzegorz-robak:


mhs/outbound/Pipfile.lock
```diff
@@ -18,38 +18,37 @@
     "default": {
         "aioboto3": {
             "hashes": [
-                "sha256:1a9850884fc49f860bcf523dff128eed904bb5440f325e56e0912f6b39e0e819",
-                "sha256:d93db5103ae6b98140c8f08e1164ec973198c990b0514946083a1b34fad90baf"
+                "sha256:327075031f053c6712e2fe875265a0acabf96604f68f3f09e1ce8d65456cc80e",
+                "sha256:3ca2b325b1f28906bd0b9311bcdd64b03e2bcc919fa97042be09cbb6e686c471"
             ],
-            "version": "==6.4.1"
+            "version": "==6.5.0"
```
jaroslaw-wojciechowski: Pipfile.lock changed but Pipfile not?

grzegorz-robak: I think that some libs don't have versions defined so it takes latest on update

jaroslaw-wojciechowski (dismissed):

mhs/outbound/Pipfile.lock
mattd-kainos: Should we roll these back and put in a separate ticket to update dependencies? If there is an issue with this merge it might not be obvious if its caused by code or dependency changes.

grzegorz-robak: ok, reverted Pipfile.lock

jaroslaw-wojciechowski (approved):

mattd-kainos (approved):

PR #168, opened 2020-03-05
NIAD-62 Updates local config examples with custom ports and sensible defaults

mattd-kainos:
NIAD-62 Updates local config examples with custom ports and sensible defaults


docker-compose.yml
```diff
@@ -50,7 +50,7 @@ services:
       - "8080:80"
     environment:
       - MHS_LOG_LEVEL=NOTSET
-      - MHS_SDS_URL=ldap://msg.opentest.hscic.gov.uk
+      - MHS_SDS_URL=ldap://192.168.128.11
```
mattd-kainos, line 53: I have to make this change for it to work. Does anyone else have DNS working on their local VPN?

mhs/spineroutelookup/nhs-spineroutelookup-env-example.yaml
```diff
@@ -1,5 +1,5 @@
 #Common env variables for outbound
-MHS_SPINE_ROUTE_LOOKUP_SERVER_PORT: "8083"
+MHS_SPINE_ROUTE_LOOKUP_SERVER_PORT: "8080"
```
mattd-kainos, line 2: route in docker-compose already runs on 8080

mhs/inbound/nhs-inbound-env-example.yaml
```diff
@@ -1,12 +1,12 @@
 #Common env variables for inbound
-MHS_INBOUND_HEALTHCHECK_SERVER_PORT: "8081"
+MHS_INBOUND_HEALTHCHECK_SERVER_PORT: "8082"
```
mattd-kainos, line 2: Wound up changing it to 8082 because I tested outbound (8081) first. Can be reverted if we need to, but the postman collection matched these ports now.

grzegorz-robak (approved):

docker-compose.yml
jaroslaw-wojciechowski: maybe adding one into etc/hots will help

jaroslaw-wojciechowski: like:
`192.168.128.11 msg.opentest.hscic.gov.uk`

mattd-kainos: Might was well just use the IP address here then? OpenTest certs can't be validated anyway.

jaroslaw-wojciechowski (approved):

PR #177, opened 2020-03-17
NIAD-69: Configuration for local component tests

grzegorz-robak:


integration-tests/all_component_test_env.yaml
```diff
@@ -0,0 +1,313 @@
+#common
```
bartek-sarul, line 1: shouldn't be in `.gitignore` ?

integration-tests/fake_spineroutelookup/fake_spineroutelookup/routing_response.py
```diff
@@ -1,10 +1,13 @@
...
+import os
 
 
 class RoutingResponse(object):
 
     def __init__(self):
-        self.mhs_endpoint = "https://fakespine"
+        fakespine_port = os.environ.get("FAKE_SPINE_PORT", default='')
+
+        self.mhs_endpoint = f"https://fakespine:{fakespine_port}" if fakespine_port else 'https://fakespine'
```
bartek-sarul: can be `f"https://fakespine:{fakespine_port or '443'}"`

mhs/outbound/fix_pycurl.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+pipenv run pip install --no-cache-dir --compile --ignore-installed --install-option="--with-openssl" --install-option="--openssl-dir=/usr/local/opt/openssl" pycurl
```
bartek-sarul, line 2: Can you give more info what it does?

bartek-sarul (changes requested):

integration-tests/all_component_test_env.yaml
grzegorz-robak: Those are env variables (certificates are generated) used by all modules for component tests to work

mhs/outbound/fix_pycurl.sh
grzegorz-robak: That wasn't supposed to be here :) This script fixes `ImportError: pycurl: libcurl link-time ssl backend (openssl) is different from compile-time ssl backend (none/other)` - it was easier for me to just run script instead of copying it from readme file. That can help others so we can leave it here or I'll remove it.

integration-tests/fake_spineroutelookup/fake_spineroutelookup/routing_response.py
grzegorz-robak: Seems to work - updated

bartek-sarul (approved):

PR #186, opened 2020-03-24
NIAD-98: set tracking ids when calling spine route lookup

bartek-sarul:


common/utilities/mdc.py
```diff
@@ -0,0 +1,21 @@
...
+    headers = {}
+    if correlation_id.get():
+        headers[HttpHeaders.CORRELATION_ID] = correlation_id.get()
+    if message_id.get():
+        headers[HttpHeaders.MESSAGE_ID] = message_id.get()
+    if interaction_id.get():
+        headers[HttpHeaders.INTERACTION_ID] = interaction_id.get()
+    if inbound_message_id.get():
+        headers[HttpHeaders.INBOUND_MESSAGE_ID] = inbound_message_id.get()
+    return headers or None
```
grzegorz-robak, line 21: I'm not sure how to deal with that in python: is `None` better than empty dict? I'd return just empty dict as it resolves to false and is iterable (can use `in`/`not in` on it)

common/utilities/mdc.py
```diff
@@ -0,0 +1,21 @@
...
+def build_tracking_headers():
+    headers = {}
+    if correlation_id.get():
+        headers[HttpHeaders.CORRELATION_ID] = correlation_id.get()
+    if message_id.get():
+        headers[HttpHeaders.MESSAGE_ID] = message_id.get()
+    if interaction_id.get():
+        headers[HttpHeaders.INTERACTION_ID] = interaction_id.get()
+    if inbound_message_id.get():
+        headers[HttpHeaders.INBOUND_MESSAGE_ID] = inbound_message_id.get()
```
grzegorz-robak, line 20: If you don't set keys in `headers`, trying to access `headers[HttpHeaders.CORRELATION_ID]` will rise `KeyError`. Maybe you should put all keys inside `header` but make their default value as `None` or empty string. That way calling `headers[HttpHeaders.CORRELATION_ID].get()` will equal to `None` or empty string

grzegorz-robak (changes requested):

common/utilities/mdc.py
grzegorz-robak: Do you have a test case when `headers` is `None`?

SCR/scr/tests/test_xmls/EmptyHtmlGpSummaryUpdate.xml
```diff
@@ -70,6 +70,7 @@
                             <seperatableInd value="false"/>
                             <presentationText classCode="OBS" moodCode="EVN">
                                 <value mediaType="text/plain">
+
```
bartek-sarul, line 73: This is needed

common/utilities/mdc.py
bartek-sarul: I used `None` because it's how it was used at https://github.com/nhsconnect/integration-adaptors/pull/186/files#diff-b1db7b05027c20d67989fe0d09d531d8L59

common/utilities/mdc.py
bartek-sarul: If I leave key with empty value, I assume it's exactly what will be added to HTTP headers - empty header. I think it's not something we want.
The headers builder is used only for HTTP requests.

common/utilities/mdc.py
grzegorz-robak: Ok we can leave it, I'm not that familiar with Python conventions - for me it looked exactly like passing around `null` in Java.

common/utilities/mdc.py
grzegorz-robak: ok

grzegorz-robak (dismissed):

jaroslaw-wojciechowski (dismissed):

PR #193, opened 2020-03-27
NIAD-24: added docker run script and updated local running readme

michael-kainos:


pageholder (dismissed):

docker-build-components-example.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+: '
+Make a copy of this file and rename to docker-build-components.sh
+as the renamed file is in .gitignore. This will avoid commiting the file
+with certificate information.
+'
+echo "Exporting environment variables"
+#Put your party key here (OpenTest):
+export MHS_SECRET_PARTY_KEY=""
```
mattd-kainos: do all of these variables need to be exported or none of them?

docker-build-components-example.sh
```diff
@@ -0,0 +1,25 @@
...
+
+#Put your Endpoint issuing subCA certificate and your Root CA certificate here (OpenTest)
+MHS_SECRET_CA_CERTS=""
+
+#Stops containers, builds images and starts containers
+echo "Stopping running containers"
+docker-compose -f docker-compose.yml -f docker-compose.lb.override.yml stop;
+
+echo "Build and starting containers"
+docker-compose -f docker-compose.yml -f docker-compose.lb.override.yml up -d --build
```
mattd-kainos: Personal preference really, but I tend to run "attached" or without the -d flag. Sometimes the logs in the console are useful.

docker-build-components-example.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+: '
+Make a copy of this file and rename to docker-build-components.sh
```
bartek-sarul: I don't like the idea of making this example shell file
Env vars can be set in bash/zsh startup script
This file would then have only docker commands and don't have to be in gitignore

docker-build-components-example.sh
michael-kainos: Should all be exported as they are used in the docker compose

docker-build-components-example.sh
michael-kainos: Makes sense, i only do this so a user does not try to carry on using the shell and accidently exits from their start up. Thoughts on this as a use case?

mhs/running-mhs-adaptor-locally.md
```diff
@@ -51,4 +51,31 @@ used to facilitate health checks in a load balanced environment.
...
+
+## Running MHS components and load balancer with docker
+
+To run the MHS services using docker containers follow the steps below:
+
+1. Make a file in the root call `export-env-vars-and-run-mhs-docker.sh` this name must match to be excluded from git
+2. Populate the file with the following information:
+  ```sh
+#!/bin/bash
+echo -e "\033[1;32mExporting environment variables\033[0m"
```
bartek-sarul: Hard to understand at first look
Please add variables as in https://github.com/nhsconnect/integration-adaptors/pull/193/files#diff-b42d9fb95f574b12b6c1a033fa91dddbR3

bartek-sarul (changes requested):

bartek-sarul (dismissed):

bartek-sarul (approved):

PR #199, opened 2020-03-31
NIAD-147: In-code LDAP mock that can be used instead of real LDAP con…

bartek-sarul:
…nection


mhs/spineroutelookup/lookup/sds_mock_connection_factory.py
```diff
@@ -0,0 +1,155 @@
...
+
+    session = boto3.Session(profile_name=aws_profile)
+    s3 = session.client('s3')
+
+    file_names = [_SERVER_INFO_FILE, _SERVER_SCHEMA_FILE, _SERVER_ENTRIES_FILE]
+    for file_name in file_names:
+        src_file_path = os.path.join(key_base, file_name)
+        dest_file_path = os.path.join(_MOCK_DATA_BASE_PATH, file_name)
+        logger.info("Downloading S3 file from '%s' to '%s'", src_file_path, dest_file_path)
+        s3.download_file(bucket_name, src_file_path, dest_file_path)
```
jaroslaw-wojciechowski, line 53: what will happen in case of s3 connection error ?

jaroslaw-wojciechowski: please ignore - should be handled by 'main'

jaroslaw-wojciechowski (dismissed):

PR #202, opened 2020-04-01
NIAD-64: fix failing forward reliable integration tests

bartek-sarul:


common/comms/common_https.py
```diff
@@ -61,4 +63,8 @@ class CommonHttps(object):
                         "proxy_port": http_proxy_port,
                         "code": response.code
                     })
+
+        if logger.isEnabledFor(logging.DEBUG):
```
bartek-sarul, line 67: Need this check because `decode()` fails on unit tests and it's easier to add this check there than modify tons of tests

mhs/inbound/inbound/request/handler.py
```diff
@@ -46,6 +47,8 @@ def initialize(self, workflows: Dict[str, workflow.CommonWorkflow],
     @time_request
     async def post(self):
         logger.info('Inbound POST received: {request}', fparams={'request': self.request})
+        if logger.isEnabledFor(logging.DEBUG):
```
bartek-sarul, line 50: Need this check because `decode()` fails on unit tests and it's easier to add this check there than modify tons of tests

Broom94 (approved):

common/comms/common_https.py
grzegorz-robak: I'd put a comment in code explaining that - what if I run mock with DEBUG level?

mattd-kainos: Would it be better to leave out logging of the body? I think the underlying http client will do this if enabled. There is also the warning at startup about logging clinical data.

mhs/inbound/inbound/request/handler.py
grzegorz-robak: You can consider: https://stackoverflow.com/questions/1835756/using-try-vs-if-in-python

grzegorz-robak (approved):

common/comms/common_https.py
bartek-sarul: The warning says:
```The current log level (NOTSET) is set below INFO level, it is known that libraries used by this application sometimes log out clinical patient data at DEBUG level. The log level provided MUST NOT be used in a production environment.```
I think it's ok to actually put this request+response in DEBUG

mattd-kainos: DEBUG not used for anything else, no reason to enable in production. Other libraries could potentially log out similar data if DEBUG were enabled also.

mattd-kainos (approved):

PR #204, opened 2020-04-02
NIAD-50: env var parser fix

bartek-sarul:


Broom94 (dismissed):

mhs/outbound/main.py
```diff
@@ -140,7 +140,7 @@ def main():
                                                   ca_certs=secrets.get_secret_config('CA_CERTS'))
     max_retries = int(config.get_config('OUTBOUND_TRANSMISSION_MAX_RETRIES', default="3"))
     retry_delay = int(config.get_config('OUTBOUND_TRANSMISSION_RETRY_DELAY', default="100"))
-    validate_cert = bool(config.get_config('OUTBOUND_VALIDATE_CERTIFICATE', default="True"))
+    validate_cert = config.get_config('OUTBOUND_VALIDATE_CERTIFICATE', default=str(True)).lower() == str(True).lower()
```
grzegorz-robak: If env var is set then `config.get_config()` will return String. I think that isn't the way it should be

grzegorz-robak (changes requested):

mhs/outbound/main.py
mattd-kainos: `default='true').lower() == 'true'` would be simpler. Is there a reason its written this way?

grzegorz-robak: ok, as I read it I saw `default=str(True).lower()`. Looks like it's easy to misread it. Please split that into two lines

grzegorz-robak (changes requested):

mhs/outbound/main.py
bartek-sarul: just to omit potential error in typing a string
anyway, I've added a str2bool function

grzegorz-robak (dismissed):

Jenkinsfile
```diff
@@ -152,6 +152,17 @@ pipeline {
                     }
                     post {
                         always {
+                            sh label: 'Docker status', script: 'docker ps -al'
+                            sh label: 'Dump container logs', script: '''
+                                docker logs ${BUILD_TAG_LOWER}_route_1 > logs/route.log
```
grzegorz-robak, line 158: I think that component tests doesn't use route, but fake spineroutelookup

grzegorz-robak (changes requested):

Jenkinsfile
bartek-sarul: it's just how the container is named

petervdm (approved):

grzegorz-robak (approved):

PR #214, opened 2020-04-10
NIAD-132: Allows fake-spine outbound SSL to be disabled when running …

mattd-kainos:
…behing a load balancer

* Adds FAKE_SPINE_OUTBOUND_SSL_ENABLED environment variable
* Removes FAKE_SPINE_HEALTHCHECK_PORT environment variable
* /healthcheck is always on the same port as the outbound handler: FAKE_SPINE_PORT


integration-tests/all_component_test_env.yaml
```diff
@@ -12,8 +12,7 @@ MHS_DYNAMODB_ENDPOINT_URL: http://localhost:8000
 INBOUND_SERVER_BASE_URL: https://inbound
 FAKE_SPINE_PORT: '8443'
 INBOUND_PROXY_PORT: '8888'
-FAKE_SPINE_HEALTHCHECK_PORT: '8091'
```
bartek-sarul, line 15: Not needed?

michael-kainos (approved):

PR #218, opened 2020-04-14
NIAD-145: Fixing handlers import after moving them from mhs_common to common

bartek-sarul:


nhais/main.py
```diff
@@ -1,6 +1,6 @@
 import tornado.httpclient
 import utilities.integration_adaptors_logger as log
-from requests import healthcheck_handler
+from mhs_common.request import healthcheck_handler
```
grzegorz-robak: This should be `mhs_common.handlers`, shouldn't it?

grzegorz-robak (changes requested):

nhais/main.py
bartek-sarul: no, it's in "common" now as it's shared with NHAIS

bartek-sarul: I agree, that there is still something wrong with this approach as we don't uniquely identify sub-dependency packages but let's merge this and then work on better solution

bartek-sarul: ahh damn, yes you are right

grzegorz-robak (approved):

mattd-kainos (approved):

Alicja-Kainos (approved):

PR #222, opened 2020-04-16
NIAD-180 changes needed to deploy fake-spine for v&p testing

mattd-kainos:


Jenkinsfile
```diff
@@ -116,7 +116,8 @@ pipeline {
         }
 
         stage('Test') {
-            parallel {
+            //parallel {
```
mattd-kainos, line 120: TBD if we're going to leave this disabled or find a permanent fix

.gitignore
```diff
@@ -143,12 +143,10 @@ mhs-reference-implementation/selenium_tests/data/local_asid
 /mhs/spineroutelookup/nhs-spineroutelookup-env.yaml
 export-env-vars-and-run-mhs-docker.sh
 /integration-tests/integration_tests/integration-test-env.yaml
+/integration-tests/integration_tests/integration-tests-env.yaml
```
Alicja-Kainos: Are there both files integration-test-env.yaml and integration-testa-env.yaml?

docker-compose.component.override.yml
```diff
@@ -16,6 +18,7 @@ services:
       - FAKE_SPINE_PRIVATE_KEY
       - FAKE_SPINE_CA_STORE
       - INBOUND_SERVER_BASE_URL=https://inbound
+      - MHS_LOG_LEVEL=DEBUG
```
Alicja-Kainos: should we use DEBUG level here? Isn't like that that some sensitive data would be visible in the logs? I remember that this should not be below INFO level

integration-tests/fake_spine/Pipfile
```diff
@@ -6,6 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
+integration-adaptors-common = {editable = true,path = "./../../common"}
```
Alicja-Kainos, line 9: Should it be modified here or you've just forgot to add -e when installing the package?

michael-kainos (dismissed):

integration-tests/fake_spine/Pipfile
mattd-kainos: It is my understanding that module dependencies should be included in the Pipfile

docker-compose.component.override.yml
mattd-kainos: Low risk since its only for testing, but changed to INFO for best practice 👍

.gitignore
mattd-kainos: removed the duplicate entry

Alicja-Kainos (approved):

michael-kainos (approved):

Jenkinsfile
```diff
@@ -289,7 +291,7 @@ pipeline {
                         }
                     }
                 }
-            }
+            } // parallel
```
grzegorz-robak, line 294: This comment should be before `}` as you commented `{`

grzegorz-robak (changes requested):

Jenkinsfile
mattd-kainos: @LukaszKwasniewski change to indicate where parallel would close?

LukaszKwasniewski: Yes, I've indicated were the parallel section ends.
@grzegorz-robak it is not commented out because the `parallel {` section at the top had to be replaced with `stages {` Jenkinsfiles does not allow two `stage {` sections one after another

grzegorz-robak (approved):

PR #223, opened 2020-04-17
NIAD-170: Create handler for post acceptance outbound fhir rest api

Broom94:


nhais/generate_jsonschema.py
```diff
@@ -0,0 +1,26 @@
...
+import os
+
+import utilities.integration_adaptors_logger as log
+
+logger = log.IntegrationAdaptorsLogger(__name__)
+
+def validate_schema(request):
+    request_body = json.loads(request)
+
+    ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
```
mattd-kainos: We'll take a performance hit loading this file for every request. Maybe wrap this in class the load the schema(s) in the __init__?  That might not be enough though because a new instance of Handler is created for every class.

nhais/generate_jsonschema.py
```diff
@@ -0,0 +1,26 @@
...
+    json_schema_patient = ROOT_DIR + "/json-schema-patient.json"
+
+    with open(json_schema_patient) as file:
+        data = json.load(file)
+
+    is_valid_schema = False
+
+    try:
+        validate(instance=request_body, schema=data)
```
mattd-kainos: For current or future optimisations: if we have a unit test asserting `json-schema-patient.json` is a valid JSONSchema document then we could skip the "validate the schema" step.

> validate() will first verify that the provided schema is itself valid, since not doing so can lead to less obvious error messages and fail in less obvious or consistent ways.

> If you know you have a valid schema already, especially if you intend to validate multiple instances with the same schema, you likely would prefer using the IValidator.validate method directly on a specific validator (e.g. Draft7Validator.validate).

https://python-jsonschema.readthedocs.io/en/latest/validate/#the-basics

nhais/generate_jsonschema.py
```diff
@@ -0,0 +1,26 @@
...
+        is_valid_schema = True
```
mattd-kainos: Could `return True` here and `return False` below instead of defining a variable

nhais/outbound/request/handler.py
```diff
@@ -26,6 +28,13 @@ def __init__(self, application: "Application", request: httputil.HTTPServerReque
...
-        edifact = self.fhir_to_edifact.convert(self.request.body.decode())
-        await self.mesh_wrapper.send(edifact)
-        self.set_status(202)
+        is_valid_schema = generate_jsonschema.validate_schema(self.request.body.decode())
+        if(is_valid_schema):
+            edifact = self.fhir_to_edifact.convert(self.request.body.decode())
+            unique_operation_id = message_utilities.get_uuid()
+            await self.mesh_wrapper.send(edifact)
+            self.set_status(202)
+            self.finish(f"<html><body>UniqueOperationID: {unique_operation_id}</body></html>")
```
mattd-kainos: The response needs to be a FHIR OperationOutcome with content type application/json

https://portal.developer.nhs.uk/docs/personal-demographics-dev/1/types/OperationOutcome

Something like: https://github.com/nhsconnect/integration-adaptors/blob/feature/NIAD-170-nhais-openapi-spec/nhais/specification/components/examples/responses/with-operation-id.json

Broom94: correct OperationOutcome now implemented

nhais/generate_jsonschema.py
Broom94: variables have been removed

nhais/generate_jsonschema.py
Broom94: I had a look into this, not sure I will be able to complete this in time for current estimation. Thoughts?

nhais/generate_jsonschema.py
Broom94: Currently moved to handler init, but yes I see your point. Not sure how else we can have this loaded before the handler

nhais/Pipfile
```diff
@@ -12,6 +12,7 @@ tornado = "*"
 integration-adaptors-common = {editable = true,path = "./../common"}
 pycurl = {version = "~=7.43",platform_system = "!= 'Windows'"}
 marshmallow = "~=3.2"
+jsonschema = "*"
```
bartek-sarul: put a concrete version here

nhais/generate_jsonschema.py
```diff
@@ -0,0 +1,18 @@
...
+
+    with open(json_schema_patient) as file:
+        data = json.load(file)
+
+    try:
+        validate(instance=request_body, schema=data)
+        return None
+    except Exception as e:
+        logger.error('Did not match json schema' + str(e))
+        return e
```
bartek-sarul: No, don't do that, just rethrow

nhais/outbound/request/handler.py
```diff
@@ -23,9 +27,52 @@ def __init__(self, application: "Application", request: httputil.HTTPServerReque
             password=config.get_config('OUTBOUND_QUEUE_PASSWORD', default=None))
         self.fhir_to_edifact = FhirToEdifact()
         self.mesh_wrapper = MeshOutboundWrapper(queue_adaptor)
+        root_dir = os.path.dirname(os.path.abspath(__file__))
+        self.json_schema_patient = root_dir + "/json-schema-patient.json"
```
bartek-sarul: better to use os.path

nhais/outbound/request/handler.py
```diff
@@ -23,9 +27,52 @@ def __init__(self, application: "Application", request: httputil.HTTPServerReque
...
+
+    def validate_uri_matches_payload_operation(self, operation, uri):
+        string_length_of_operation = uri.find(operation)
+        if (string_length_of_operation >= len(operation) - 1):
```
bartek-sarul: Why not just return the comparison result?

nhais/outbound/request/handler.py
```diff
@@ -23,9 +27,52 @@ def __init__(self, application: "Application", request: httputil.HTTPServerReque
...
+            return True
+        else:
+            return False
+
+    def set_unsuccesful_response(self, status, code, coding_code, details):
+        self.set_status(status)
+        data = {"issue":
```
bartek-sarul: I think you should use some FHIR lib to build FHIR JSON

bartek-sarul (changes requested):

nhais/outbound/request/handler.py
Broom94: Path is not callable object

nhais/outbound/request/handler.py
Broom94: It was for readability but I have amended this

nhais/outbound/request/handler.py
Broom94: FHIR client added

nhais/Pipfile
```diff
@@ -12,6 +12,8 @@ tornado = "*"
 integration-adaptors-common = {editable = true,path = "./../common"}
 pycurl = {version = "~=7.43",platform_system = "!= 'Windows'"}
 marshmallow = "~=3.2"
+jsonschema = "~=3.2"
```
mattd-kainos: do we still need jsonschema?

nhais/common/data/patient.json
```diff
@@ -0,0 +1,209 @@
+{
```
mattd-kainos: these json files should probably be in the tests/ folder that use them - they're test not application data

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,105 @@
...
+from fhir.resources.operationoutcome import OperationOutcomeIssue
+
+from tornado import httputil
+from outbound.schema.schema_validation_exception import SchemaValidationException
+
+from utilities import integration_adaptors_logger as log, timing
+from utilities import message_utilities
+
+from outbound.schema import validate_request
+import json
```
mattd-kainos: Run an "Optimize Imports" on the nahis/ folder. It will reorder these to be PEP 8 style

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,105 @@
...
+class AcceptanceAmendmentRequestHandler(tornado.web.RequestHandler):
+
+    def __init__(self, application, request: httputil.HTTPServerRequest, **kwargs: Any):
+        super().__init__(application, request, **kwargs)
+        self.fhir_to_edifact = FhirToEdifact()
+        self.mesh_wrapper = MeshOutboundWrapper()
+
+    def __set_unsuccesful_response(self, status, code, coding_code, details):
+        self.set_status(status)
+        # TODO: Consider building this up programmatically instead? Block until reliable library is found
```
mattd-kainos: Is this now done?

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,105 @@
...
+
+        pat2 = OperationOutcomeIssue()
+        pat2.severity = 'error'
+        pat2.code = code
+        pat2.details = details
+
+        pat = OperationOutcome()
+        pat.issue = [pat2]
+
+        data = {
```
mattd-kainos: Is this still needed?

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,105 @@
...
+
+        details = CodeableConcept()
+        details.coding = [coding]
+
+        pat2 = OperationOutcomeIssue()
+        pat2.severity = 'error'
+        pat2.code = code
+        pat2.details = details
+
+        pat = OperationOutcome()
```
mattd-kainos: What is a 'pat'? Use descriptive variable names here.

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        self.finish(pat.as_json())
+
+    @timing.time_request
+    async def post(self, patient_id):
+        try:
+            request_body = json.loads(self.request.body.decode())
+            patient = validate_request.validate_patient(request_body)
+            if (request_body['id'] == patient_id):
```
mattd-kainos: Use patient instead of request body for this check now

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,105 @@
...
+                edifact = self.fhir_to_edifact.convert(patient)
+                unique_operation_id = message_utilities.get_uuid()
+                await self.mesh_wrapper.send(edifact.as_json())
+                self.set_status(202)
+                self.set_header("OperationId", unique_operation_id)
+                await self.finish()
+            else:
+                self.__set_unsuccesful_response(400, "value", "ID_IN_URI_DOES_NOT_MATCH_PAYLOAD_ID",
+                                              f"URI id `{patient_id}` does not match PAYLOAD id `{request_body['id']}`")
+        except SchemaValidationException as e:
```
mattd-kainos: Is this still needed?

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,105 @@
...
+            if hasattr(e, 'message'):
+                details = f"{e.message}"
+                self.__set_unsuccesful_response(400, "value", "JSON_PAYLOAD_NOT_VALID_TO_SCHEMA", details)
+        except JSONDecodeError as e:
```
mattd-kainos, line 52: We should log the exceptions in all of these except blocks

nhais/outbound/request/tests/test_acceptance_amendment.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @patch.object(MeshOutboundWrapper, "send")
+    @patch.object(MeshOutboundWrapper, "__init__")
+    def test_happy_path(self, mock_init, mock_send):
+        mock_init.return_value = None
+        MagicMock.__await__ = lambda x: self.async_magic().__await__()
+
+        response = self.fetch(r'/fhir/Patient/9000000009', method="POST",
+                              body=VALID_REQUEST_BODY)
+        self.assertEqual(202, response.code)
```
mattd-kainos, line 59: Is the response an OperationOutcome? Is the correct OperationId in the response headers?

nhais/outbound/request/tests/test_acceptance_amendment.py
```diff
@@ -0,0 +1,132 @@
...
+    @patch.object(MeshOutboundWrapper, "send")
+    @patch.object(MeshOutboundWrapper, "__init__")
+    def test_invalid_payload_id_not_string(self, mock_init, mock_send):
+        mock_init.return_value = None
+        MagicMock.__await__ = lambda x: self.async_magic().__await__()
+
+        response = self.fetch('/fhir/Patient/9000000009', method="POST",
+                              body=INVALID_ID_REQUEST_BODY)
+
+        response_body = json.loads(response.body.decode())
```
mattd-kainos: Would it be better to create an fhir.resources OperationOutcome from the response and validate property values from that?

nhais/outbound/schema/tests/test_validate_request.py
```diff
@@ -0,0 +1,41 @@
...
+with open(missing_json_patient_id) as file:
+    MISSING_ID_REQUEST_BODY = json.load(file)
+
+class TestValidateRequest(unittest.TestCase):
+
+    def test_valid_schema(self):
+        try:
+            response = validate_patient(VALID_REQUEST_BODY)
+        except SchemaValidationException:
+            self.fail("myFunc() raised ExceptionType unexpectedly!")
```
mattd-kainos: This is not needed - the test will fail if an exception is raised

nhais/outbound/schema/validate_request.py
```diff
@@ -0,0 +1,25 @@
...
+from outbound.schema.schema_validation_exception import SchemaValidationException
+
+_logger = log.IntegrationAdaptorsLogger(__name__)
+
+
+def validate_patient(request_body):
+    try:
+        patient = FHIRElementFactory.instantiate('Patient', request_body)
+        __validate_patient_id_exists_in_payload(request_body)
+    except ValidationError as e:
```
mattd-kainos: Does FHIRElementFactory raise ValidationError? I suspect this is old JSONSchema code

nhais/outbound/schema/validate_request.py
```diff
@@ -0,0 +1,25 @@
...
+        raise SchemaValidationException(message=e.message, path=e.path)
+    except FHIRValidationError as e:
+        raise FHIRValidationError(e.args[0])
```
mattd-kainos: Need to raise a SchemaValidationException in the correct message and path here. It might take a bit more work to get the complete path. You'll need to traverse the error tree to its leaf, append to the path at each node, and take the message from the leaf node.

nhais/outbound/schema/validate_request.py
```diff
@@ -0,0 +1,25 @@
...
+    return patient
+
+
+def __validate_patient_id_exists_in_payload(request_body):
```
mattd-kainos, line 40: Unused code?

nhais/common/data/patient.json
mattd-kainos: Are they actually duplicates of the files already in tests/?

nhais/outbound/request/acceptance_amendment.py
mattd-kainos: Yes it is needed!

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,105 @@
...
+            self.__set_unsuccesful_response(400, "value", "PAYLOAD_IS_NOT_JSON_FORMAT",
+                                          "Payload is missing, Payload required.")
+        except TypeError as e:
+            self.__set_unsuccesful_response(400, "value", "ID_IN_PAYLOAD_IS_MISSING",
+                                            'Payload is missing id, id is required.')
+        except FHIRValidationError as e:
```
mattd-kainos: validate_patient should raise SchemaValidationException instead

nhais/outbound/schema/schema_validation_exception.py
```diff
@@ -0,0 +1,6 @@
+
+class SchemaValidationException(Exception):
```
mattd-kainos: Since we're not using JSONSchema anymore it might make sense to rename to 'RequestValidationException' or similar?

nhais/common/data/patient.json
Broom94: I originally moved these to common since other tests needed the same json files for tests. Was going to delete these duplicates in requests -> tests -> data

mattd-kainos (changes requested):
Looking good and nearly there. Just needs some tidy of JSONSchema leftovers.

nhais/outbound/request/acceptance_amendment.py
Broom94: This is used to validate patient id exists in payload since this is not a required field in the fhir schema but is needed for our implementation. I will refactor this to make it clear

nhais/outbound/schema/validate_request.py
Broom94: This is used to check that the id exists, there is an if statement in the acceptance amendment that then checks if id in payload match uri

nhais/outbound/schema/schema_validation_exception.py
Broom94: Renamed

nhais/outbound/request/acceptance_amendment.py
Broom94: Now fixed

nhais/outbound/schema/validate_request.py
Broom94: Removed

nhais/outbound/request/tests/test_acceptance_amendment.py
Broom94: All tests now create operation outcome object

nhais/outbound/request/tests/test_acceptance_amendment.py
Broom94: It is not an operationoutcome, I'm not sure how we can validate if the response has the correct operationid since its generated for each message.

nhais/outbound/schema/tests/test_validate_request.py
Broom94: Removed

nhais/outbound/request/acceptance_amendment.py
Broom94: logged in each exception now within this class, do we need to log exceptions that are raised to here?

nhais/outbound/request/acceptance_amendment.py
Broom94: Change made

nhais/Pipfile
Broom94: removed

nhais/outbound/request/tests/test_acceptance_amendment.py
Broom94: I believe it was agreed before this was the expected outcome until we decided what the correct response should be. That is that operationid is in the header and no operationoutcome

nhais/mesh/mesh_outbound.py
```diff
@@ -1,17 +1,22 @@
...
-        self.queue_retry_delay = float(config.get_config('RETRY_DELAY', default='0.1'))
+    def __init__(self):
+        self.queue_adaptor = proton_queue_adaptor.ProtonQueueAdaptor(
+            urls=config.get_config('OUTBOUND_QUEUE_BROKERS').split(','),
+            queue=config.get_config('OUTBOUND_QUEUE_NAME'),
+            username=config.get_config('OUTBOUND_QUEUE_USERNAME', default=None),
+            password=config.get_config('OUTBOUND_QUEUE_PASSWORD', default=None),
+            max_retries=int(config.get_config('OUTBOUND_QUEUE_MAX_RETRIES', default='3')),
+            retry_delay=int(config.get_config('OUTBOUND_QUEUE_RETRY_DELAY', default='100')) / 1000)
+        self.queue_max_retries = int(config.get_config('OUTBOUND_MAX_RETRIES', default='3'))
```
bartek-sarul: Why those 2 on lines 18 and 19 are here? Line 16 and 17 already sets those values

bartek-sarul (changes requested):

nhais/outbound/request/acceptance_amendment.py
mattd-kainos: > do we need to log exceptions that are raised to here?

I don't understand they are caught not raised here?

mattd-kainos: Use logger.exception instead of logger.error so we get the stack traces in the logs

nhais/outbound/request/tests/test_acceptance_amendment.py
mattd-kainos: You're right, there is another ticket for this 👍

nhais/mesh/mesh_outbound.py
```diff
@@ -1,25 +1,27 @@
...
+            queue=config.get_config('OUTBOUND_QUEUE_NAME'),
+            username=config.get_config('OUTBOUND_QUEUE_USERNAME', default=None),
+            password=config.get_config('OUTBOUND_QUEUE_PASSWORD', default=None),
+            max_retries=int(config.get_config('OUTBOUND_QUEUE_MAX_RETRIES', default='3')),
+            retry_delay=int(config.get_config('OUTBOUND_QUEUE_RETRY_DELAY', default='100')) / 1000)
         self.transmission = None
 
     async def _publish_message_to_inbound_queue(self, message):
-
         result = await retriable_action.RetriableAction(
```
bartek-sarul: No. What I meant above is that the "put message onto queue" already has the retry functionality built in

nhais/nhais-env-example.yaml
```diff
@@ -1,11 +1,12 @@
 #Common env variables
-NHAIS_SERVER_PORT: "80"
+NHAIS_OUTBOUND_SERVER_PORT: "80"
 NHAIS_LOG_LEVEL: NOTSET
-NHAIS_MAX_RETRIES: "10"
-NHAIS_RETRY_DELAY: "0.1"
+NHAIS_OUTBOUND_MAX_RETRIES: "10"
+NHAIS_OUTBOUND_RETRY_DELAY: "0.1"
```
bartek-sarul: Delay is in miliseconds

nhais/nhais-env-example.yaml
```diff
@@ -1,11 +1,12 @@
...
 
 #User-specific env variables
-NHAIS_OUTBOUND_QUEUE_HOST: "localhost:5672/outbound"
+NHAIS_OUTBOUND_QUEUE_BROKERS: amqp://localhost:5672
+NHAIS_OUTBOUND_QUEUE_NAME: 'outbound'
```
bartek-sarul: We'll probably be using the same broker for all queues. On prod this is not a problem as api user will define it's own but for testing purpose I suggest prefixing this with "nhais"

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __init__(self, application, request: httputil.HTTPServerRequest, **kwargs: Any):
+        super().__init__(application, request, **kwargs)
+        self.fhir_to_edifact = FhirToEdifact()
+        self.mesh_wrapper = MeshOutboundWrapper()
+
+    def __set_unsuccesful_response(self, status, code, coding_code, details):
+        self.set_status(status)
+
+        coding = Coding()
```
bartek-sarul: Might be more readable if you extract builing OperationOutcome resource into a separate method

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,86 @@
...
+        operation_outcome.issue = [operation_outcome_issue]
+
+        self.finish(operation_outcome.as_json())
+
+    @timing.time_request
+    async def post(self, patient_id):
+        try:
+            request_body = json.loads(self.request.body.decode())
+            patient = validate_request.validate_patient(request_body)
+            if patient.id == patient_id:
```
bartek-sarul, line 36: Check if it's possible to do this kind of validation on some "filter" level. Maybe a validation decorator? Then we could add it on all HTTP handler methods

nhais/outbound/request/acceptance_amendment.py
```diff
@@ -0,0 +1,86 @@
...
+                                                f"URI id `{patient_id}` does not match PAYLOAD id `{patient.id}`")
+                logger.error('Error, id doesnt match')
+        except RequestValidationException as e:
+            details = f"{e.path} {e.message}"
+            self.__set_unsuccesful_response(400, "value", "JSON_PAYLOAD_NOT_VALID_TO_SCHEMA", details)
+            logger.error(f'Error: {details}')
+        except JSONDecodeError as e:
+            self.__set_unsuccesful_response(400, "value", "PAYLOAD_IS_NOT_VALID_JSON_FORMAT",
+                                            "Payload is either missing, empty or not valid to json, this is required.")
+            logger.error('Error: Payload is missing, empty or not a valid json')
```
bartek-sarul: Might be more safe to log before `self.__set_unsussesful...`

bartek-sarul (changes requested):

nhais/outbound/request/acceptance_amendment.py
mattd-kainos: The location of the id in the payload will be different for other requests. Let's leave this as is for now. If there is a common way to do it we can include that in NIAD-177 and NIAD-178

nhais/nhais-env-example.yaml
mattd-kainos: name the outbound queue 'nhais_mesh_outbound'. Reason being for inbound there will be two queues: mesh and supplier

mattd-kainos (dismissed):

nhais/main.py
```diff
@@ -1,8 +1,14 @@
 import tornado.httpclient
+import tornado.httpserver
+import tornado.ioloop
+import tornado.web
+
 import utilities.integration_adaptors_logger as log
 from handlers import healthcheck_handler
+from outbound.request import acceptance_amendment
```
bartek-sarul: It's a minor thing but why import differently than 2 other on line 9 and 10?

nhais/mesh/mesh_outbound.py
```diff
@@ -1,39 +1,34 @@
...
-
-        if not result.is_successful:
-            logger.error("Exceeded the maximum number of retries, {max_retries} retries, when putting "
-                         "message onto inbound queue",
-                         fparams={"max_retries": self.queue_max_retries})
+    async def _publish_message_to_outbound_queue(self, message):
+        try:
+            await self._put_message_onto_queue_with(message)
+        except Exception as e:
+            raise e
```
bartek-sarul: What's the point?

bartek-sarul (changes requested):

bartek-sarul (approved):

mattd-kainos (approved):

PR #226, opened 2020-04-20
NIAD-186: updated documentation with MHS_INBOUND_USE_SSL variable and Rest-Clie…

Alicja-Kainos:
…nt request collection


mhs/inbound/main.py
```diff
@@ -87,7 +87,7 @@ def start_inbound_server(local_certs_file: str, ca_certs_file: str, key_file: st
                                                              config_manager=config_manager))])
 
     ssl_ctx = build_ssl_context(local_certs_file, ca_certs_file, key_file) \
-        if str2bool(config.get_config('INBOUND_USE_SSL', default=str(True))) \
```
grzegorz-robak: That shouldn't be changed - MHS prefix is dropped after `config` is set up

mhs/inbound/mhs-inbound-env-example.yaml
```diff
@@ -7,6 +7,7 @@ MHS_SYNC_ASYNC_STATE_TABLE_NAME: sync_async_state
 MHS_DYNAMODB_ENDPOINT_URL: http://localhost:8000
 AWS_ACCESS_KEY_ID: test
 AWS_SECRET_ACCESS_KEY: test
+MHS_INBOUND_USE_SSL: true
```
grzegorz-robak: I'd set it to `false` - as that's default for OpenTest

mhs/README.md
```diff
@@ -56,15 +56,22 @@ Please refer to the [API Documentation](outbound/openapi-docs.html) for further
...
+ - synchronous request to outbound, with retries while sending request to nhs application other than spine
+ - synchronous request to outbound
+ - solicited request to inbound
+ - unsolicited request to inbound.
+ 
+There are also health, reliability and routing checks.
+ 
+Before sending these requests, you will need to create a setting.json file as described in [README](../rest-client/README.md).
+ 
+Before sending requests directly to the inbound service, set 'MHS_INBOUND_USE_SSL: false' in .yaml inbound configuration file.
```
grzegorz-robak: Would be good to write why it has to set to `false`

grzegorz-robak (changes requested):

mhs/inbound/mhs-inbound-env-example.yaml
mattd-kainos: Wouldn't inbound SSL be enabled by default for OpenTest?

mattd-kainos (dismissed):

grzegorz-robak (dismissed):

mhs/inbound/mhs-inbound-env-example.yaml
Alicja-Kainos: I asked on the team channel and guys say it should be disabled, @mattd-kainos

Alicja-Kainos: @bartek-sarul confirmed it is enabled, I will  remove to previous version then

grzegorz-robak (dismissed):

mhs/inbound/mhs-inbound-env-example.yaml
Alicja-Kainos: MHS_INBOUND_USE_SSL is only for testing purpose to enable request being sent directly to inbound without ssl

bartek-sarul (approved):

Broom94 (approved):

PR #230, opened 2020-04-22
NIAD-75: Spine route lookup can connect to LDAP in a lazy way

grzegorz-robak:


mhs/spineroutelookup/lookup/sds_connection_factory.py
```diff
@@ -41,7 +42,56 @@ def build_sds_connection_tls(ldap_address: str, private_key: str, local_cert: st
...
+    logger.info('Configuring LDAP connection using TLS')
+
+    return _configure_ldap_connection(server)
+
+
+def _configure_ldap_connection(server) -> ldap3.Connection:
+    lazy_ldap = str2bool(config.get_config("LAZY_LDAP", default=str(True)))
+    if lazy_ldap:
+        connection = ldap3.Connection(server,
+                                      lazy=True,
```
jaroslaw-wojciechowski, line 54: do we need lazy_ldap var?
cant we use:
lazy=config.get_config("LAZY_LDAP", default=str(True)) ?

grzegorz-robak: I can inline it into:
````python
def _configure_ldap_connection(server) -> ldap3.Connection:
    if str2bool(config.get_config("LAZY_LDAP", default=str(True))):
````
That would be better?

jaroslaw-wojciechowski (approved):

Broom94 (approved):

PR #239, opened 2020-04-29
NIAD-139: outbound sequence number manager generate

Alicja-Kainos:


nhais/sequence/interchange_id.py
```diff
@@ -0,0 +1,15 @@
...
+class InterchangeIdGenerator:
+    """A component that provides sequential interchange id.
+    Definition:
+    SIS (Send Interchange Sequence) - sequence number for the entire EDIFACT interchange """
+
+    def __init__(self):
+        self.table_name = 'transaction_id_counter'
+
+    async def generate_interchange_id(self, sender, recipient) -> int:
+        key = "SIS-" + str(sender) + "-" + str(recipient)
```
mattd-kainos: better to use an fstring than + concatenation
f"SIS-{sender}-{recipient}"

nhais/sequence/message_id.py
```diff
@@ -0,0 +1,15 @@
...
+class MessageIdGenerator:
+    """A component that provides sequential message id
+    Definition:
+    SMS (Send Message Sequence) - sequence number applied for each message within an interchange"""
+
+    def __init__(self):
+        self.table_name = 'transaction_id_counter'
+
+    async def generate_message_id(self, sender, recipient) -> int:
+        key = "SMS-" + str(sender) + "-" + str(recipient)
```
mattd-kainos: as above

nhais/sequence/interchange_id.py
Alicja-Kainos: Thank you, good idea :)

nhais/sequence/tests/component_test_ids_generators.py
```diff
@@ -0,0 +1,110 @@
...
+class ComponentTestIds(unittest.TestCase):
+
+    def setUp(self):
+        config.setup_config("MHS")
+        self.table_name = 'test_ids_generators_table'
+        self.key = 'transaction_id'
+        self.endpoint = config.get_config('DYNAMODB_ENDPOINT_URL', None)
+        self.region_name = 'eu-west-2'
+
+        '''Create instances of three generators and replace the default table_name
```
mattd-kainos: Undecided about this comment style. It seems that they are used that way, but they're actually string literals not assigned to a variable rather than comments. PEP 8 prefers # block comments it seems https://www.python.org/dev/peps/pep-0008/#block-comments

Alicja-Kainos: changed into block comments

mattd-kainos: One more string comment here

nhais/sequence/interchange_id.py
```diff
@@ -0,0 +1,15 @@
+from sequence.sequence_factory import get_sequence_generator
+
+
+class InterchangeIdGenerator:
```
grzegorz-robak, line 4: (Idea) This class has the same implementation as `MessageIdGenerator` but with different key value - maybe we should create generic generator and make key as input for it? I mean that we could have i.e. something like this:
````
interchange_key = InterchangeKey(sender, recipient)
interchange_id = IdGenerator().next(key)
````
That way generator code is extracted from key that's changing and we could easily reuse generators in different adapters

File: nhais/sequence/tests/component_test_ids_generators.py (hunk 3 above)
Comment (grzegorz-robak) [inline nhais/sequence/tests/component_test_ids_generators.py]: What is the use case for overriding table_name? Shouldn't that stay immutable?

File: nhais/sequence/tests/component_test_ids_generators.py [hunk 5]
```diff
@@ -0,0 +1,110 @@
...
+
+
+class ComponentTestIds(unittest.TestCase):
+
+    def setUp(self):
+        config.setup_config("MHS")
+        self.table_name = 'test_ids_generators_table'
+        self.key = 'transaction_id'
+        self.endpoint = config.get_config('DYNAMODB_ENDPOINT_URL', None)
+        self.region_name = 'eu-west-2'
```
grzegorz-robak, line 20: Are you required to supply region in tests?

nhais/sequence/tests/component_test_ids_generators.py
Alicja-Kainos: done

nhais/sequence/interchange_id.py
Alicja-Kainos: @mattd-kainos , what do you think about it?

nhais/sequence/tests/component_test_ids_generators.py
Alicja-Kainos: botocore.exceptions.NoRegionError: You must specify a region. <- exception if no region specified

grzegorz-robak: ok 👍

nhais/sequence/tests/component_test_ids_generators.py
Alicja-Kainos: In the test we use separate table that should be created for each test and deleted after the test run, so I don't wont to use the same table that is use in runtime , for I need to recreate table for each test run.

nhais/sequence/interchange_id.py
grzegorz-robak: It might be over-engineering after all... If this works and has just few lines than I probably stretched that too much

mattd-kainos: Could do it that way, but I don't think it affects reusability because the reusable bit SequenceGenerator is already its own interface / implementation.

Another option would be to have a single class with methods to generate each type of identifier. The reasoning would be in case we need to add transactions in the future.

mattd-kainos: A single class would also let us define the table_name in one place? The name of the dynamo table probably needs to be changed as well since it is now used for all sequence numbers.

mattd-kainos (changes requested):
At least the table name needs to be updated. Could also consider changing class structure of the generators, but the code is simple enough that a future refactoring would be easy.

nhais/sequence/interchange_id.py
grzegorz-robak: I think it is designed that way to store all ids in one table. Do we want each id to have it's own table?

Alicja-Kainos: all ids use the same table

Alicja-Kainos:
> At least the table name needs to be updated. Could also consider changing class structure of the generators, but the code is simple enough that a future refactoring would be easy.

@mattd-kainos  I've changed the table name into 'generated_id_counter'

Broom94 (dismissed):

jaroslaw-wojciechowski (dismissed):

grzegorz-robak (dismissed):

jaroslaw-wojciechowski (approved):

Broom94 (approved):

grzegorz-robak (approved):

mattd-kainos (approved):

PR #243, opened 2020-05-04
NIAD-155 outbound sequence number manager record send

Alicja-Kainos:
Any suggestion how the saving would be done better according to best practices when it comes to python and dynamodb are welcome :)


nhais/outbound/state/work_description.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        if persistence_store is None:
+            raise ValueError('Expected persistence store')
+
+        self.persistence_store = persistence_store
+        self._deserialize_data(store_data)
+
+    async def publish(self):
+        """
+        Attempts to publish the local state of the work description to the state store, checks versions to avoid
```
grzegorz-robak: At what point it checks version?

nhais/outbound/state/work_description.py
```diff
@@ -0,0 +1,136 @@
...
+def create_new_work_description(persistence_store: pa.PersistenceAdaptor,
+                                segments
+                                ) -> WorkDescription:
+    """
+    Builds a new local work description instance given the details of the message, these details are held locally
+    until a `publish` is executed
+    """
+
+    if persistence_store is None:
+        raise ValueError('Expected persistence store to not be None')
```
grzegorz-robak: The same check is done in `WorkDescription` constructor

grzegorz-robak (dismissed):

nhais/outbound/state/work_description.py
Alicja-Kainos: right, it should be removed from one place :)

nhais/outbound/state/work_description.py
Alicja-Kainos: That is the case: I did not want to use the version checking like in MHS (get, than update). Anyway, the scope of the task is to save the state, not doing update, if I got it right :)

nhais/outbound/state/work_description.py
Alicja-Kainos: done

nhais/outbound/state/work_description.py
Alicja-Kainos: @mattd-kainos, @bartek-sarul , what do you think about it, any ideas? should we store the version in the table or do the versioning other way?

grzegorz-robak: Sure, my point wasn't to make versioning, but if it doesn't do that then remove versioning from description :)

Alicja-Kainos: removed XD

grzegorz-robak (approved):

mattd-kainos (changes requested):
Will need to be merged to integration-adaptor-nhais instead

Alicja-Kainos:
changes moved to separate nhais repo;
nhais module from integration-adaptor repo is removed int this pull request (https://github.com/nhsconnect/integration-adaptors/pull/244#pullrequestreview-406372033)

PR #246, opened 2020-05-07
NIAD-243: perf improvements

bartek-sarul:


common/persistence/dynamo_persistence_adaptor.py
```diff
@@ -26,41 +28,87 @@ class RecordRetrievalError(RuntimeError):
...
         """Add an item to a specified table, using a provided key.
 
-        :param key: The key used to identify the item.
         :param data: The item to store in persistence.
-        :return: The previous version of the item which has been replaced. (None if no previous item)
         """
-        logger.info('Adding data for {key}', fparams={'key': key})
+
+        if 'key' not in data:
+            raise ValueError("Key missing")
```
grzegorz-robak: That message is missing some details also. Seeing that in logs it could mean anything. Please add for example table name

common/persistence/dynamo_persistence_adaptor.py
```diff
@@ -26,41 +28,87 @@ class RecordRetrievalError(RuntimeError):
...
+
+        logger.info('Adding data for {key}', fparams={'key': data['key']})
```
grzegorz-robak: Should we log full data at INFO level?

grzegorz-robak: I'd add table name here also (check all log messages for that)

grzegorz-robak (dismissed):

common/persistence/dynamo_persistence_adaptor.py
bartek-sarul: Key is just the identifier. I would be careful logging anything else in info level as it might consist some medical data

grzegorz-robak (approved):

PR #252, opened 2020-05-14
NIAD-249: rename sync-async header into wait-for-response

Alicja-Kainos:
renames sync-async header into wait-for-response; added MHS-Outbound.yaml provided by Matt Lewis


mhs/outbound/outbound/request/synchronous/handler.py
```diff
@@ -179,12 +179,13 @@ def _parse_body(self):
...
+    def _extract_wait_for_response_header(self):
+        wait_for_response_header = self.request.headers.get(HttpHeaders.WAIT_FOR_RESPONSE, None)
+        if not wait_for_response_header:
+            logger.error('Failed to parse wait-for-response header from message')
+            raise tornado.web.HTTPError(400, 'wait-for-response header missing',
+                                        reason='wait-for-response header missing')
+        if wait_for_response_header.lower() == 'true':
             return True
         else:
             return False
```
grzegorz-robak: (Optional) This implementation assumes anything different than `True/true` as `False` (so i.e. `AAA` is also false). But all falsy values (like `0`) will show up as header is missing. Do we keep it that way?

integration-tests/integration_tests/integration_tests/end_to_end_tests/int_asynchronous_express_messaging_pattern_tests.py
```diff
@@ -64,17 +64,18 @@ def test_should_return_successful_response_from_spine_to_message_queue(self):
...
         # Act
-        responses = send_messages_concurrently(messages, interaction_id='QUPC_IN160101UK05', sync_async=True)
+        responses = send_messages_concurrently(messages, interaction_id='QUPC_IN160101UK05', wait_for_response=True)
 
         # Assert
         all_sync_async_states = MHS_SYNC_ASYNC_TABLE_DYNAMO_WRAPPER.get_all_records_in_table()
         if has_errors(responses):
-            # TODO: NIAD-72 remove prints and has_errors once sync-async feature is stable
+            # TODO: NIAD-72 remove prints and has_errors once wait-for-response feature is stable
+            # TODO: 2020-05-14 should it be removed now? please comment while review
```
grzegorz-robak: I think it should be removed as we don't have pending bugs for it

Jenkinsfile
```diff
@@ -71,13 +71,13 @@ pipeline {
...
                             }
                         }
-                        stage('Check documentation') {
-                            steps {
-                                dir('mhs/outbound') {
-                                    sh label: 'Check API docs can be generated', script: 'pipenv run generate-openapi-docs > /dev/null'
-                                }
-                            }
-                        }
+//                         stage('Check documentation') {
```
grzegorz-robak: why is that commented out?

integration-tests/integration_tests/integration_tests/end_to_end_tests/int_asynchronous_express_messaging_pattern_tests.py
Alicja-Kainos: I'll delete it :)

Jenkinsfile
Alicja-Kainos: Matt Lewis asked me to disable generating of the open api. Shouldn't it be commented out, @grzegorz-robak?

mhs/outbound/outbound/request/synchronous/handler.py
Alicja-Kainos: @petervdm, should the case of any string other than true or false be taken as false or should the exception by thrown?

Jenkinsfile
grzegorz-robak: I don't know why he asked you for it. If it isn't needed anymore then we should delete it. Or if it should be commented out for some time then reason for this should be placed as comment also.

Alicja-Kainos: Matt Lewis confirmed, that it should be deleted. I understood it wrong before, I thought it is jus temporary.
So I am removing it

MattLewis80: I asked because we had a lot of queries in regards to the API spec from our test partner.  @petervdm and myself discussed this and agreed that the current auto generator should stop being used for MHS and all future adaptors

mhs/outbound/outbound/request/synchronous/handler.py
petervdm: Well spotted @grzegorz-robak  The test partner reported this yesterday and I was going to raise another ticket for it, but you might as well do it in here @Alicja-Kainos

Alicja-Kainos: @petervdm , I'll do it here :) But I need to know if invalid string should be treated like an exception or like false. I think this should be an exception, shouldn't be?

petervdm: Yes an exception

petervdm (approved):

jaroslaw-wojciechowski (approved):

PR #253, opened 2020-05-15
NIAD-260: Using git common dependency

bartek-sarul:


SCR/Pipfile
```diff
@@ -6,7 +6,7 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-integration-adaptors-common = {editable = true,path = "./../common"}
+integration-adaptors-common = {editable = true,git = "https://git@github.com/nhsconnect/integration-adaptor-common.git",tag = '0.1.6'}
```
jaroslaw-wojciechowski, line 9: tag = '0.1.6' for every project, maybe it should be a global value.

bartek-sarul: this will be fixed in next PR where there is a single Pipfile for the whole MHS

jaroslaw-wojciechowski (approved):

PR #254, opened 2020-05-15
Set the deregistration delay to remove the old container from LB quicker

LukaszKwasniewski:


bartek-sarul:
moved to:
https://github.com/nhsconnect/integration-adaptor-mhs/pull/6

PR #261, opened 2020-06-04
adding tfvars file for the vp environment

jaff47:


terraform/aws/components/base/vpc_peering_dlt.tf
```diff
@@ -0,0 +1,76 @@
...
+}
+
+# Add a route to the DLT VPC in the NHAIS VPC route table
+resource "aws_route" "base_to_dlt_route" {
+  route_table_id = aws_route_table.private.id
+  destination_cidr_block = data.aws_vpc.dlt_vpc.cidr_block
+  vpc_peering_connection_id = aws_vpc_peering_connection.dlt_peering.id
+  depends_on = [aws_vpc_peering_connection.dlt_peering]
+}
+resource "aws_security_group_rule" "nhais_inbound_security_group_amazon_dlt_ingress_rule" {
```
LukaszKwasniewski: This rule is not needed, and does not meet the intended use for core SG.
Core SG is there to allow all resources in Env to certain things like VPC Endpoints, NAT, and in this case the DLT.
Ingress from DLT to env should be done on the SG for Load Balancer

terraform/aws/components/base/vpc_peering_dlt.tf
```diff
@@ -0,0 +1,76 @@
...
+  to_port = 80
+  protocol = "tcp"
+  # Not making any assumptions here about the internal structure of the DLT VPC.
+  # This can be changed and made more specific to lock this down more.
+  cidr_blocks = [
+    data.aws_vpc.dlt_vpc.cidr_block]
+  description = "Allow inbound requests to nhais tasks"
+}
+
+resource "aws_security_group_rule" "nhais_inbound_security_group_amazon_dlt_egress_rule" {
```
LukaszKwasniewski: Does anything in the ENV make connection to DLT?
I guess all the connections are initiated at the DLT

terraform/aws/modules/module_ecs_service/sg_rules_dlt.tf
```diff
@@ -0,0 +1,33 @@
+# vpc peering to the network where DLT is hosted
+
+data "aws_vpc" "dlt_vpc" {
+  id = var.dlt_vpc_id
+}
+
+# DLT traffic - From the Distributed load tester to the load balancer
+
+resource "aws_security_group_rule" "dlt_ingress_rule_to_loadbalancer" {
```
LukaszKwasniewski, line 10: Not exactly the way I intended it to be done but if it works it can stay, but:
1, Make it optional with count parameter, not all environments require DLT connection
2. The ports should not be hardcoded, the port should be the same as the listener port in LB, use the same variable as in LB listener

terraform/aws/components/base/vpc_peering_dlt.tf
jaff47: Cool, let me remove test and get back to you.

terraform/aws/modules/module_ecs_service/sg_rules_dlt.tf
jaff47: What do you think?

terraform/aws/components/base/vpc_peering_dlt.tf
jaff47: I don't think so, let me pull this and test. Will get back to you shortly

terraform/aws/modules/module_ecs_service/sg_rules_dlt.tf
LukaszKwasniewski: The port is ok
The count needs to be added

terraform/aws/components/base/vpc_peering_dlt.tf
jaff47: Removed now

jaff47: Done,

terraform/aws/components/base/vpc_peering_dlt.tf
jaff47: Removed now

terraform/aws/modules/module_ecs_service/sg_rules_dlt.tf
jaff47: Done

terraform/aws/modules/module_ecs_service/variables.tf
```diff
@@ -180,6 +180,12 @@ variable "enable_load_balancing" {
   default = true
 }
 
+variable "enable_dlt" {
+  type = bool
+  description = "Should the containers in the service be attached to dlt"
+  default = true
```
mattd-kainos: false by default, enable specifically for vnp

terraform/aws/modules/module_ecs_service/variables.tf
```diff
@@ -253,3 +259,8 @@ variable "lb_allowed_cidrs" {
   description = "List of CIDRs that will be allowed to access the Load Balancer"
   default = []
 }
+variable "dlt_vpc_id" {
+  type = string
+  description = "VPC where Distributed Load testing environment is hosted"
+  default = "vpc-03f843c08b01876d5"
```
LukaszKwasniewski: remove the default from here and supply via component and module declaration

mattd-kainos (changes requested):
Just the enable_dlt false by default change

terraform/aws/modules/module_ecs_service/variables.tf
jaff47: removed now

mattd-kainos (approved):

LukaszKwasniewski (approved):

PR #285, opened 2020-08-12
NIAD-421: MESH client certs validation

jaroslaw-wojciechowski:


terraform/aws/components/nhais/variables.tf
```diff
@@ -151,6 +151,13 @@ variable "nhais_mesh_host" {
   default = ""
 }
 
+variable "NHAIS_MESH_CERT_VALIDATION" {
+  type = string
+  description = "false will disable certificate validation for SSL connection"
+  #default = "false"
+  default = "false"
```
jaroslaw-wojciechowski: has to be changed to 'true' after uploading sub ca cert.
For now, it's 'false' to fix deploy.

terraform/aws/components/nhais/data_secrets.tf
```diff
@@ -34,6 +34,10 @@ data "aws_secretsmanager_secret" "nhais_mesh_endpoint_cert" {
   name = "nhais_${var.environment}_mesh_endpoint_cert"
 }
 
+data "aws_secretsmanager_secret" "nhais_mesh_sub_ca" {
+  name = "nhais_${var.environment}_mesh_sub_ca"
```
jaroslaw-wojciechowski, line 38: right now dummy values in secret manager

jaroslaw-wojciechowski:
tf test passed: http://ec2-3-9-114-173.eu-west-2.compute.amazonaws.com/job/Terraform/625/

terraform/aws/components/nhais/variables.tf
mattd-kainos: NHAIS_MESH_CERT_VALIDATION needs to go into locals_env_variables.tf - this is where non-sensitive environment variables are defined

This file contains variable that are internal to terraform. You should also create an "nhais_mesh_cert_validation" variable here. The environment variable will get its "valueFrom" this variable.

Then in terraform/aws/etc you can set a default value for each environment.

build1: true (this *should* use OpenTest but we don't actually use build1 for anything)
vp: false (this uses fake-mesh(
ptl: true (this uses PTL/INT MESH API - it is important that correct CA certs are configured in the secret since we're testing this currently)

terraform/aws/components/nhais/data_secrets.tf
mattd-kainos: ptl: we need to add the real CA cert for INT
build1: use the CA cert for OpenTest
vp: can be anything / blank, we'll use opentest and disable cert validation

mattd-kainos (changes requested):
Needs a bit more to set NHAIS_MESH_CERT_VALIDATION as an environment variable and define defaults

jaff47 (approved):

mattd-kainos (approved):

PR #289, opened 2020-08-26
Niad 324 aws document db tls

jaff47:


terraform/aws/components/base/variables.tf
```diff
@@ -88,7 +88,7 @@ variable "docdb_master_password" {
 
 variable "docdb_tls" {
   type = string
-  default = "disabled"
+  default = "enabled"
```
LukaszKwasniewski, line 91: make this enabled / disabled on per environment basis - move it to etc/

LukaszKwasniewski: I guess better option would be to set the default to disabled (like it already is) and enable it via variable set in etc

terraform/aws/components/nhais/variables.tf
```diff
@@ -78,7 +78,19 @@ variable "mq_broker_name" {
 variable "nhais_mongo_options" {
   type = string
   description = "Options for Mongo"
-  default = "replicaSet=rs0&readPreference=secondaryPreferred&retryWrites=false"
+  default = "replicaSet=rs0&readPreference=secondaryPreferred&retryWrites=false&tls=true"
```
LukaszKwasniewski: As TF may not allow default for variable to be based on other variable we may need to remove the default here and set it in global without the tls and with tls on envs where it is enabled

terraform/aws/components/nhais/variables.tf
```diff
@@ -78,7 +78,19 @@ variable "mq_broker_name" {
...
+}
+
+variable "nhais_mongo_documentdb_tls_enabled" {
+  type = string
+  description = "Enable TLS on AWS Document DB"
+  default = "true"
```
LukaszKwasniewski: same as with base, I guess default is false, the actual value set in etc/ files.

grzegorz-robak: default is false, we set true only for exemplar

jaff47: It should be only enabled on V&P now, once testing is completed will apply to PTL (exemplar)

LukaszKwasniewski (approved):

PR #300, opened 2020-12-15
Feature/niad 806

LukaszKwasniewski:


terraform/aws/components/nhais/locals_env_variables.tf
```diff
@@ -1,5 +1,9 @@
 locals {
   environment_variables = concat(var.nhais_environment_variables,[
+    {
+      name = "NHAIS_OUTBOUND_HTTP_PROXY"
```
mattd-kainos: We're unlikely to update NHAIS to support an HTTP proxy for MESH API connections at this point, but let's leave this in as a template for GP2GP and other adaptors.

Could we change to the name to `NHAIS_MESH_API_HTTP_PROXY`? Inbound/outbound have a slightly different meaning for NHAIS.

mattd-kainos: Edit: changed to NHAIS_MESH_API_HTTP_PROXY - since we'd only apply the proxy to HTTP requests for the MESH API rather than making this a global configuration.

mattd-kainos (approved):

PR #313, opened 2021-01-11
Feature/niad 804 2

LukaszKwasniewski:


terraform/aws/components/gp2gp/outputs.tf
```diff
@@ -1,25 +1,26 @@
-output "gp2gp_url" {
-  value = aws_route53_record.gp2gp_lb_record.name
-  description = "The DNS name of the Route53 record pointing to the gp2gp service's load balancer."
-}
+# output "gp2gp_url" {
```
mattd-kainos: Should we just delete these elements instead of leaving them commented out?

mattd-kainos (approved):

PR #349, opened 2021-03-12
Add VPC Endpoint for secrets manager, required by fargate 1.4

LukaszKwasniewski:


terraform/aws/components/gp2gp/data_secrets.tf
```diff
@@ -13,3 +13,7 @@ data "aws_secretsmanager_secret" "docdb_master_username" {
 data "aws_secretsmanager_secret" "docdb_master_password" {
   name = "docdb-master-password"
 }
+
+data "aws_secretsmanager_secret" "gp2gp_ssl_trust_store_password" {
+  name = "nhais_${var.environment}_ssl_trust_store_password"
```
RIashfaqahmed, line 18: will secret variable be nhais or gp2gp prefixed?

LukaszKwasniewski: the variable is prefixed, the actual secret was first set for nhais and I've just reused it. I see no point in creating another secret with different name and the same value.

RIashfaqahmed (approved):

PR #352, opened 2021-03-19
NIAD-1016: Add lab-results adapter in VP environment

RIashfaqahmed:


terraform/aws/etc/eu-west-2_vp.tfvars
```diff
@@ -57,6 +57,20 @@ gp2gp_create_mhs_mock = true
...
 gp2gp_ssl_trust_store_url = "s3://nhsd-aws-truststore/rds-truststore.jks"
 
+# Settings for "lab-results" component
+lab-results_service_minimal_count = 1
+lab-results_service_desired_count = 1
+lab-results_service_maximal_count = 1
+lab-results_service_target_request_count = 1200
+lab-results_service_container_port = 8080
+lab-results_service_launch_type = "FARGATE"
+lab-results_log_level = "DEBUG"
```
michael-kainos: Can we set this to INFO as this will use up lots of resources, we should only be setting this to debug when required

terraform/aws/etc/eu-west-2_vp.tfvars
```diff
@@ -57,6 +57,20 @@ gp2gp_create_mhs_mock = true
...
+lab-results_mesh_host = "https://msg.opentest.hscic.gov.uk/messageexchange/"
```
michael-kainos: i think this needs to be a fake mesh container host, can you check this with @bartek-sarul

michael-kainos (changes requested):

terraform/aws/etc/eu-west-2_vp.tfvars
bartek-sarul: Yes, VP should be using fake-mesh container

terraform/aws/etc/eu-west-2_vp.tfvars
RIashfaqahmed: Log level changed to INFO

terraform/aws/etc/eu-west-2_vp.tfvars
RIashfaqahmed: "lab-results_mesh_host" is set to use https://mesh.vp.nhsredteam.internal.nhs.uk:8829/messageexchange/

michael-kainos (approved):

PR #358, opened 2021-03-26
NIAD-1154:GPC-Consumer ECS Service Setup

RIashfaqahmed:
Configuration files required to setup GPC-Consumer ECS service with GP2GP component

### DIFF
--- a/terraform/aws/components/gp2gp/gpc-consumer-locals.tf
+++ b/terraform/aws/components/gp2gp/gpc-consumer-locals.tf
@@ -1,0 +1,23 @@
+locals {
+  gpc-consumer_default_tags = {
+    Project = var.project
+    Environment = var.environment
+    Component = "gpc-consumer" #var.component
+  }
+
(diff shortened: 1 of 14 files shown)


terraform/aws/components/gp2gp/gpc-consumer-locals.tf
```diff
@@ -0,0 +1,22 @@
+locals {
+  gpc-consumer_default_tags = {
+    Project = var.project
+    Environment = var.environment
+    Component = "gpc-consumer" #var.component
```
LukaszKwasniewski: component in tags should be the same as the var.componet and the component applied

RIashfaqahmed: gpc-consumer default Component tage changed to the component applied

LukaszKwasniewski (dismissed):

LukaszKwasniewski (approved):

PR #395, opened 2021-09-15
NIAD-1635: MHS documentation update (#394)

bartek-sarul:


bartek-sarul:
Already approved in https://github.com/nhsconnect/integration-adaptors/pull/394

PR #460, opened 2023-09-22
NIAD-2832: daisy chain gp2gp adaptors

benhession:
## What
Update the AWS terraform scripts to daisy chain the GP2GP and PS Adaptor by default and support the newest versions of the GP2GP, GP Connect Consumer and Patient Switching adaptors. Change the PTL scripts to use a GP Connect provider mock by default. 

## Why
There is a development requirement to test the adaptors in the INT environment. Upgrading the terraform scripts to support the newest versions of the adaptors allows fixes to be tested pre-release. Enabling daisy chaining for the deployment allows us run the sending (GP2GP Adaptor) and receiving (PS Adaptor) simultaneously. Finally, using a mock GP Connect provider allows us to use the sending adaptor to test documents over INT and ensure they are consumed by the incumbent systems.

### DIFF
--- a/terraform/aws/components/gp2gp/gpc-consumer-locals_env_variables.tf
+++ b/terraform/aws/components/gp2gp/gpc-consumer-locals_env_variables.tf
@@ -12,29 +12,30 @@ locals {
       name  = "GPC_CONSUMER_LOGGING_LEVEL"
       value = var.gpc-consumer_log_level
     },
-    {
-      name  = "GPC_CONSUMER_URL"
-      value = "http://${module.gpc-consumer_ecs_service.loadbalancer_dns_name}:${var.gpc-consumer_service_container_port}"
-    },
-    {
-      name  = "GPC_CONSUMER_OVERRIDE_GPC_PROVIDER_URL"
-      value = var.gp2gp_create_gpcapi_mock ? "http://${module.gpcapi_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}" : var.gpc-consumer_override_gpc_provider_url
-    },
-    {
-      name  = "GPC_CONSUMER_GPC_GET_URL"
-      value = var.gp2gp_create_sdsapi_mock ? "http://${module.sdsapi_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}" : var.gpc-consumer_override_gpc_provider_url
-    },
+#    {
+#      name  = "GPC_CONSUMER_URL"
+#      value = "http://${module.gpc-consumer_ecs_service.loadbalancer_dns_name}:${var.gpc-consumer_service_container_port}"
+#    },
+#    {
+#      name  = "GPC_CONSUMER_OVERRIDE_GPC_PROVIDER_URL"
+#      value = var.gp2gp_create_gpcapi_mock ? "http://${module.gpcapi_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}" : var.gpc-consumer_override_gpc_provider_url
+#    },
+#    {
+#      name  = "GPC_CONSUMER_GPC_GET_URL"
+#      value = var.gp2gp_create_sdsapi_mock ? "http://${module.sdsapi_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}" : var.gpc-consumer_override_gpc_provider_url
+#    },
     {
       name  = "GPC_CONSUMER_SDS_URL"
-      value = var.gp2gp_create_sdsapi_mock ? "http://${module.sdsapi_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}" : var.gpc-consumer_sds_url
+#      value = var.gp2gp_create_sdsapi_mock ? "http://${module.sdsapi_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}" : var.gpc-consumer_sds_url
+      value = var.gp2gp_create_wiremock ? "http://${module.gp2gp_wiremock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_wiremock_container_port}/spine-directory/" : var.gpc-consumer_sds_url
     },
     {
-      name  = "GPC_CONSUMER_SSP_FQDN"
-      value = var.gpc-consumer_ssp_fqdn
+      name  = "GPC_CONSUMER_SSP_URL"
+      value = var.gpc-consumer_ssp_url
     },
-    {
-      name  = "GPC_ENABLE_SDS"
-      value = var.gpc_enable_sds
-    }
+#    {
+#      name  = "GPC_ENABLE_SDS"
+#      value = var.gpc_enable_sds
+#    }
   ])
 }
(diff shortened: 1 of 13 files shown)


terraform/aws/components/gp2gp/gpc-consumer-locals_env_variables.tf
```diff
@@ -12,29 +12,13 @@ locals {
...
-      name  = "GPC_CONSUMER_OVERRIDE_GPC_PROVIDER_URL"
-      value = var.gp2gp_create_gpcapi_mock ? "http://${module.gpcapi_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}" : var.gpc-consumer_override_gpc_provider_url
-    },
-    {
-      name  = "GPC_CONSUMER_GPC_GET_URL"
-      value = var.gp2gp_create_sdsapi_mock ? "http://${module.sdsapi_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}" : var.gpc-consumer_override_gpc_provider_url
-    },
     {
       name  = "GPC_CONSUMER_SDS_URL"
-      value = var.gp2gp_create_sdsapi_mock ? "http://${module.sdsapi_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}" : var.gpc-consumer_sds_url
```
adrianclay, line 29: Sooooo, it looks like we wouldn't be using the SDS Mock API at all anymore? Did you have any thoughts on whether to remove entirely?

Also, because I'm not clear from the PR description, what was the thinking behind using wiremock vs the SDS Mock API? (FYI I've not looked into the latter)

terraform/aws/components/gp2gp/locals_env_variables.tf
```diff
@@ -34,7 +34,7 @@ locals {
     },
     {
       name  = "GP2GP_MHS_INBOUND_QUEUE"
-      value = var.mhs_inbound_queue_name
+      value = var.daisy_chaining_active ? var.daisy_chaining_queue : var.mhs_inbound_queue_name
```
adrianclay: I think I'd actually prefer something like 
```suggestion
      value = var.gp2gp_inbound_queue_name
```

Simpler, and then if you want to use the daisy chaining queue name, you can specify it within the relevant `tfvars` file. 🤔

terraform/aws/components/gp2gp/locals_env_variables.tf
```diff
@@ -56,17 +56,13 @@ locals {
       name = "GP2GP_STORAGE_CONTAINER_NAME"
       value = aws_s3_bucket.gp2gp_extract_cache_bucket.id
     },
-    {
-      name = "GP2GP_GPC_HOST"
-      value = var.gp2gp_gpc_host
-    },
```
adrianclay, line 62: Nice.

terraform/aws/components/gp2gp/locals_env_variables.tf
```diff
@@ -56,17 +56,13 @@ locals {
...
     {
       name = "GP2GP_MHS_OUTBOUND_URL"
       value = var.gp2gp_create_mhs_mock ? "http://${module.mock_mhs_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_mhs_port}/mock-mhs-endpoint" : "http://mhs-outbound.${trimsuffix(data.terraform_remote_state.base.outputs.r53_zone_name,".")}/"
     },
     {
       name = "GP2GP_GPC_GET_URL"
-      value = var.gp2gp_create_gpcc_mock ? "http://${module.gpcc_mock_ecs_service[0].loadbalancer_dns_name}:${var.gp2gp_mock_port}/GP0001/STU3/1/gpconnect" : "http://${module.gpc-consumer_ecs_service.loadbalancer_dns_name}:${var.gpc-consumer_service_container_port}/B82617/STU3/1/gpconnect"
+      value = var.gp2gp_create_wiremock ? "http://${module.gpc-consumer_ecs_service.loadbalancer_dns_name}:${var.gpc-consumer_service_container_port}/@ODS_CODE@/STU3/1/gpconnect" : var.gp2gp_gpc_get_url
```
adrianclay, line 65: I have similar questions about the GPCC mock as I had for the SDS mock above.

terraform/aws/components/pss/data_secrets.tf
```diff
@@ -33,3 +33,7 @@ data "aws_secretsmanager_secret" "postgres_gpc_facade_password" {
 data "aws_secretsmanager_secret_version" "postgres_psdbowner_username" {
   secret_id = data.aws_secretsmanager_secret.postgres_psdbowner_username.id
 }
+
+data "aws_secretsmanager_secret" "gpc-consumer_sds_apikey" {
+  name = var.secret_name_sds_apikey
```
adrianclay, line 38: What was the thinking behind making this a variable? Looks like the rest of this file uses hardcoded values.

adrianclay: Oh, I see we use this variable already within the GP2GP component, makes sense.

terraform/aws/components/pss/s3.tf
```diff
@@ -0,0 +1,56 @@
+resource "aws_s3_bucket" "pss_attachment_bucket" {
+  bucket = "${local.resource_prefix}-attachment-storage-bucket"
+  tags = merge(local.default_tags, {
+      Name = "${local.resource_prefix}-attachment-storage-bucket"
+    })
+}
+
+resource "aws_s3_bucket_lifecycle_configuration" "pss_attachment_bucket_lifecycle_config" {
```
adrianclay, line 8: Nice use of a lifecycle.

terraform/aws/components/pss/variables.tf
```diff
@@ -172,3 +172,32 @@ variable "pss_gp2gp_translator_container_port" {
...
+variable "pss_attachment_bucket_retention_period" {
+  type = number
+  description = "Number of days objects will be retained in pss_attachment_storage_bucket"
+  default = 7
+}
+
+variable "pss_storage_type" {
+  type = string
+  description = "The storage type used for attachments"
+  default = "LocalMock"
```
adrianclay: Minor: Seems a little strange to make this configurable given we're implementing AWS specific infra.

terraform/aws/components/gp2gp/gpc-consumer-locals_env_variables.tf
benhession: The SDS Mock API configuration relies on there being two mocks, the SDS Mock and the GPC Mock.  The "wiremock" mocks both these services.  The environment variables used to configure the consumer adaptor to use  the "two mock" configuration has been removed from the GPC Consumer adaptor.

I think removing the SDS Mock and the GPC Mock from the scripts makes a lot of sense. 

Something else to note is the changes already made will break other environments defined in these scripts, namely, `build1`. `build 2`, `build3`, `kdev` and `vp`. As these also expect older versions of the adaptor. I'm not sure what the value in maintaining these environments is? Maybe they should also be removed?

terraform/aws/components/pss/variables.tf
benhession: Agreed, I'll hardcode it to S3

adrianclay (approved):

PR #470, opened 2024-07-16
NIAD-3004

martin-nhs:
- Added terraform environment variable for redactions support.


terraform/aws/components/gp2gp/locals_env_variables.tf
```diff
@@ -75,6 +75,10 @@ locals {
     {
       name = "GP2GP_GPC_OVERRIDE_FROM_ASID"
       value = var.gp2gp_gpc_override_from_asid
+    },
+    {
+      name = "GP2GP_REDACTIONS_ENABLED"
+      value = var.gp2gp_redactions_enabled
```
adrianclay, line 81: I'm no Terraform expert, but I'm guessing it magically converts the "bool" type to a string?

martin-nhs: "Terraform automatically converts number and bool values to strings when needed. It also converts strings to numbers or bools, as long as the string contains a valid representation of a number or bool value."

https://developer.hashicorp.com/terraform/language/expressions/types#type-conversion

In this case, the environment variable will be parsed to either `true` or `false`, and Spring configuration will be able to parse this into a boolean value.

adrianclay (approved):

